Issue #14, opened 2021-03-29
Replace internal usage of Redis-py with SmartRedis clients


Spartee:
# Description
Currently we depend on both Redis-py and Redis-py-cluster. Given that we have our own clients, we should be able to perform the orchestrator status check with SILC instead of depending on these libraries.

# Justification
Users will have to download and install less dependencies

# Implementation Strategy

Currently ``Orchestrator.check_cluster_status is implemented with Redis-py-cluster as follows:

```python
    def check_cluster_status(self):
        """Check that a cluster is up and running

        :raises SmartSimError: If cluster status cannot be verified
        """
        # TODO use silc for this, then we don't have to create host dictionary
        host_list = []
        for host in self.hosts:
            for port in self.ports:
                host_dict = dict()
                host_dict["host"] = host
                host_dict["port"] = port
                host_list.append(host_dict)

        trials = 10
        logger.debug("Beginning database cluster status check...")
        while trials > 0:
            # wait for cluster to spin up
            time.sleep(2)
            try:
                redis_tester = RedisCluster(startup_nodes=host_list)
                redis_tester.set("__test__", "__test__")
                redis_tester.delete("__test__")
                logger.debug("Cluster status verified")
                return
            except ClusterDownError:
                logger.debug("Cluster still spinning up...")
                time.sleep(5)
                trials -= 1
        if trials == 0:
            raise SmartSimError("Cluster setup could not be verified")

```

With SILC python we could

```python
    def check_cluster_status(self):
        """Check that a cluster is up and running

        :raises SmartSimError: If cluster status cannot be verified
        """
        # pick a single host and port
       client = silc.Client(address=host+port, cluster=False)

        trials = 10
        logger.debug("Beginning database cluster status check...")
        while trials > 0:
            # wait for cluster to spin up
            time.sleep(2)
            try:
                client.put_tensor(np.array(1,2,3,4))
                logger.debug("Cluster status verified")
                return
            except RedisReplyError:
                logger.debug("Cluster still spinning up...")
                time.sleep(5)
                trials -= 1
        if trials == 0:
            raise SmartSimError("Cluster setup could not be verified")

```

Spartee:
Acceptance Criteria
 - Remove dependence on Redis-py and Redis-py-cluster by replacing with SmartRedis
 - Implement the function at the bottom with SmartRedis
 - Implement tests to confirm functionality of ``check_cluster_status``

PR #50, opened 2021-05-26
Ray Cluster features

al-rigazzi:
This draft PR adds Ray cluster setup and deployment to SmartSim. 

Depending on what we end up deciding, the `smartsim.ray` directory might become `smartsim.exp.ray` in the real PR.

The `ray.raystarter` script file allows us to bypass some IP inconsistencies we encountered when starting the head node in the standard way. We need to be on the node already to correctly get the head node IP.

I also rolled the batch preamble into this PR, it is implemented in the BatchSettings base class.


requirements.txt
```diff
@@ -1,8 +1,8 @@
 psutil>=5.7.2
 coloredlogs==10.0
 pandas>=1.1.3
-redis==3.0.1
-redis-py-cluster==2.0.0
+redis==3.5.3
```
Spartee, line 4: Will these be needed for Ray? We are planning on taking them out with #14 

Also, the smartsim ray functionaility should have it's own dependency section)

```cfg
[options.extras_require]
dev =
    black>=20.8b1
    torchvision==0.8.2
    torch==1.7.1
    isort>=5.6.4
    pylint>=2.6.0
    matplotlib>=3.3.3
    pytest>=6.0.0
    pytest-cov>=2.10.1

doc=
    sphinx==3.1.1
    breathe==4.27.0
    sphinx_rtd_theme>=0.5.0
    sphinx-fortran==1.1.1
    nbsphinx>=0.8.2

ray=
 .....
.......
```

The ray section will just be the needed requirements to run Ray workloads.

users will install with ``pip install smartsim[ray]`` or ``pip install smartsim[dev,ray]`` which will also install ray extra dependencies.

smart
```diff
@@ -61,7 +61,7 @@ def clean(install_path, _all=False):
         rai_path = lib_path.joinpath("redisai.so")
         if rai_path.is_file():
             rai_path.unlink()
-            print("Succesfully removed existing RedisAI installation")
+            print("Successfully removed existing RedisAI installation")
```
Spartee, line 64: thanks.

smartsim/control/controller.py
```diff
@@ -352,6 +360,85 @@ def _launch_orchestrator(self, orchestrator):
...
+                    # catch if it fails or launcher doesn't support it
+                except LauncherError:
+                    logger.debug("WLM Ray worker node aquisition failed")
+                except SSUnsupportedError:
+                    logger.debug("WLM Ray worker node acquisition unsupported")
+        else:
+            head_step = self._create_job_step(ray_cluster.head_model)
+            self._launch_step(head_step, ray_cluster.head_model)
+            ray_cluster._get_ray_head_node_address()
+            if isinstance(self._launcher, SlurmLauncher):
```
Spartee, line 404: Why do we specify slurm here? because slurm is the only one with node acquisition implemented? I would think we should leave this conditional out and let the ``WLMLauncher`` class throw a ``SSUnsupportedError``. 

That way when we do find mechanisms to parse node information from PBS and/or cobalt, adding them requires less refactoring.

smartsim/control/controller.py
```diff
@@ -378,22 +465,28 @@ def _launch_step(self, job_step, entity):
...
         :rtype: Step
         """
         batch_step = self._launcher.create_step(
-            entity_list.name, entity_list.path, entity_list.batch_settings
+            entity.name, entity.path, entity.batch_settings
         )
-        for entity in entity_list.entities:
-            # tells step creation not to look for an allocation
+        if isinstance(entity, EntityList):
+            for e in entity.entities:
```
Spartee, line 480: no need to specify ``entity.entities`` can just use the ``__iter__`` method so

```python
for _entity in entity_list

```

Spartee, line 480: No single letter variable names

smartsim/control/controller.py
```diff
@@ -378,22 +465,28 @@ def _launch_step(self, job_step, entity):
...
         :return: job step instance
         :rtype: Step
         """
         batch_step = self._launcher.create_step(
-            entity_list.name, entity_list.path, entity_list.batch_settings
+            entity.name, entity.path, entity.batch_settings
         )
-        for entity in entity_list.entities:
-            # tells step creation not to look for an allocation
+        if isinstance(entity, EntityList):
```
Spartee, line 479: Why is the condition here needed? Is it ever the case that an object is launched as a batch besides an ``EntityList ``?

smartsim/experiment.py
```diff
@@ -110,6 +112,12 @@ def stop(self, *args):
             for entity in args:
                 if isinstance(entity, SmartSimEntity):
                     self._control.stop_entity(entity)
+                elif isinstance(entity, RayCluster):
+                    if entity._launcher == 'local':
```
Spartee, line 116: I don't like that this function is deviating from the two main base classes.

 - Since the stop method is ray specific, possibly we make it a method in the ``RayCluster`` class? 
 - Are we sure ``ray`` will always be in the path?

Spartee: Is there something specific ``ray stop`` is doing that we cannot emulate with traditional process killing?

smartsim/experiment.py
```diff
@@ -380,7 +388,7 @@ def sprint(p):
             print(p, flush=True)
 
         sprint("\n")
-        models, ensembles, orchestrator = separate_entities(args)
+        models, ensembles, orchestrator, ray_clusters = separate_entities(args)
```
Spartee, line 391: I can see this getting possibly unruly if we add more object types, should ``separate_entities`` return a usable data structure instead? Something easy to add to later on? 

Something like

```python

class Deployment:

    def __init__(self, args):

   def get_entities():
        #seperates and returns entities
 
   def get_db():
         # ...

   and so on 

```

Advantage would be to have a data structure to be carried through the controller until jobs are created. The same data structure could be reused in other places/functions that require entity separation due to type agnostic APIs.

smartsim/ray/raycluster.py
```diff
@@ -0,0 +1,378 @@
...
+    def _initialize_entities(self):
+        self.head_model = RayHead(name="head", params=None,
+                                  path=os.path.join(self.path, 'head'),
+                                  ray_password = self._ray_password,
+                                  ray_port=self._ray_port,
+                                  launcher=self._launcher,
+                                  run_args=self._run_args,
+                                  ray_num_cpus=self._ray_num_cpus,
+                                  alloc=self._alloc,
+                                  batch=self.batch,
```
Spartee, line 99: Can a user specify a queue or a partition for batch workloads? I see ``run_args`` but no method for parameter or ``batch_args``

smartsim/ray/raycluster.py
```diff
@@ -0,0 +1,378 @@
...
+    :param ray_port: Port at which the head node will be running
+    :type ray_port: int
+    :param ray_num_cpus: Number of CPUs used by Ray
+    :type ray_num_cpus: int
+    :param run_args: Arguments to pass to launcher to specify details such as partition, time, and so on
+    :type run_args: dict[str,str]
+    :param launcher: Name of launcher to use for starting the cluster
+    :type launcher: str
+    :param alloc: ID of allocation to run on, only used if launcher is Slurm and allocation is
+                  obtained with ``ray.slurm.get_allocation``
```
Spartee, line 183: ``ray.slurm``?

Spartee: ``smartsim.slurm``?

smartsim/ray/raycluster.py
```diff
@@ -0,0 +1,378 @@
...
+    :type time: str
+    """
+    def __init__(self, 
+                 name, 
+                 params, 
+                 path, 
+                 ray_password, 
+                 ray_port=6780, 
+                 run_args={}, 
+                 ray_num_cpus=12,
```
Spartee, line 197: Im going to argue the default should be lower than this. esp for local launching. maybe 4?

smartsim/settings/alpsSettings.py
```diff
@@ -42,10 +42,19 @@ def __init__(self, exe, exe_args=None, run_args=None, env_vars=None):
         :type run_args: dict[str, str], optional
         :param env_vars: environment vars to launch job with, defaults to None
         :type env_vars: dict[str, str], optional
+        :param block_in_batch: whether the execution of this call should block when executed
+                               in a batch.
+        :type block_in_batch: bool
+        :param expand_exe: whether the executable path should be expanded. It is recommended
```
Spartee, line 48: Would the alternative be that it's in your ``PATH``?

smartsim/settings/alpsSettings.py
```diff
@@ -103,6 +112,21 @@ def set_hostlist(self, host_list):
         if not all([isinstance(host, str) for host in host_list]):
             raise TypeError("host_list argument must be list of strings")
         self.run_args["nodelist"] = ",".join(host_list)
+        
+    def set_excludelist(self, host_list):
```
Spartee, line 116: ``set_excluded_hosts``?

Spartee (changes requested):
A couple things to think through. We need to be smart about how we add to the base data structures of the library. to many branching conditionals leads to untestable/fragile code. 

We should talk about the ``Deployment`` data structure I mentioned in the review. Having ``separate_entities`` return a data structure with helpful methods could reduce some of the overhead in adding new inheritors of ``EntityList``. Which I imagine we will want to add more soon.

Some quick notes:
 - need to clean up build/install with dependencies
 - need more testing in both the ``on_wlm`` and ``full_wlm`` categories.
 - ``RayCluster`` object should have some helper methods for things like
     - ``get_head_address`` - returns pretty formatted address
     - ``add_batch_arg`` - for adding arguments to batch e.g. ``queue``
     - ``get_dashboard_address`` - return dashboard address

 - Documentation

Haven't looked through the tutorial yet, but I will get to that next.

requirements.txt
al-rigazzi: I had to update the version numbers because they were conflicting with what Ray required. If we take them out, Ray will install what it needs and we will not have to care about versions anymore, which is good.

smartsim/control/controller.py
al-rigazzi: OK, I like this.

smartsim/control/controller.py
al-rigazzi: That's the case for Ray, when we launch as batch. Two batch files are created, one for head and one for workers, but both head and workers inherit from `Model`, as each one is only one executable.

smartsim/experiment.py
al-rigazzi: Addressing both comments. This is a mechanism I had to implement for `local` launcher only. When launching locally, you have to have ray in the path, because it is needed to start the cluster (it is imported in the `raystarter.py` script. 
Yes can make this a method of `RayCluster`, I think it will be much cleaner.
Emulating `ray stop` seems to be complicated, we would need to track which processes are spawned by `ray start` and also make sure there are no other issues (like checkpointing and so on I might not be aware of).

smartsim/ray/raycluster.py
al-rigazzi: Mh, right. I might have mixed up the two.

smartsim/settings/alpsSettings.py
al-rigazzi: I think I got it from Slurm, but I like yours more.

smartsim/settings/alpsSettings.py
al-rigazzi: Actually the problem with this is that if it is an executable you have in another conda (or virtual) env, you cannot expand it here.

requirements.txt
al-rigazzi: Adapted the requirements to those currently in `develop`, created `ray` section in `setup.cfg` (with only Ray, right now).

smartsim/experiment.py
al-rigazzi: In the end, it seems like launching the head with `--block` is enough to make it kill all other processes. As RayCluster is an EntityList, we will rely on the standard mechanism here.

smartsim/ray/raycluster.py
al-rigazzi: Fixed

al-rigazzi:
Docs are located in another branch

Spartee (commented):
Much much cleaner. Great work. a few comments.

setup.cfg
```diff
@@ -65,7 +65,10 @@ doc=
     sphinx_rtd_theme>=0.5.0
     sphinx-fortran==1.1.1
     nbsphinx>=0.8.2
+    myst-parser>=0.15.1
```
Spartee, line 68: I don't think this is needed anymore

smartsim/exp/ray/raycluster.py
```diff
@@ -0,0 +1,493 @@
...
+    :type dictionary: dict
+    :param key_list: the list of keys to delete from the dictionary.
+    :type key: any
+    """
+    for key in key_list:
+        if key in dictionary:
+            del dictionary[key]
+
+
+### TODO
```
Spartee, line 53: This can be taken out?

smartsim/exp/ray/raycluster.py
```diff
@@ -0,0 +1,493 @@
...
+
+    attempts = 0
+    head_ip = None
+    while head_ip is None:
+        _time.sleep(5)
+        with open(head_log) as fp:
+            line = fp.readline()
+            while line:
+                plain_line = re.sub("\033\\[([0-9]+)(;[0-9]+)*m", "", line)
+                if "Local node IP:" in plain_line:
```
Spartee, line 288: Does this parse from the raystarter?

smartsim/exp/ray/raycluster.py
```diff
@@ -0,0 +1,493 @@
...
+class RayWorker(SmartSimEntity):
+    def __init__(
+        self,
+        name,
+        path,
+        ray_password,
+        ray_port,
+        run_args=None,
+        ray_args=None,
+        interface="eth0",
```
Spartee, line 403: Wondering if the default interface should be ``ipogif0`` which is the high speed network on XC.

``eth0`` would be cloud and local....

smartsim/exp/ray/raystarter.py
```diff
@@ -0,0 +1,84 @@
...
+]
+
+if args.ray_args:
+    cliargs += args.ray_args
+
+if args.redis_password:
+    cliargs += [f"--redis-password={args.redis_password}"]
+
+# On some systems, ssh to compute nodes (and port forwarding) is not allowed.
+# If that's the case, the user should bind the dashboard to 0.0.0.0,
```
Spartee, line 72: this comment seems to be about the host address but the following lines are about the port number. Im I missing something?

smartsim/experiment.py
```diff
@@ -463,6 +466,36 @@ def sprint(p):
...
+            for rc in ray_clusters:
+                name = colorize(rc.name, color="green", bold=True)
+                num_models = colorize("# of nodes: " + str(len(rc)), color="green")
+                if rc.batch:
+                    batch_settings = colorize(
+                        "Ray batch Settings: \n" + str(rc.batch_settings),
+                        color="green",
+                    )
+                head_run_settings = colorize(
+                    "Head run Settings: \n" + str(rc.entities[0].run_settings),
```
Spartee, line 480: ``Ray head settings``

smartsim/experiment.py
```diff
@@ -463,6 +466,36 @@ def sprint(p):
...
+                    color="green",
+                )
+                run_settings = head_run_settings
+                if len(rc) > 0:
+                    worker_run_settings = colorize(
+                        "\nWorker run Settings: \n" + str(rc.entities[1].run_settings),
```
Spartee, line 486: ``Ray worker settings``. I was confused about this when running.

tests/with_ray/full_wlm/test_ray_pbs_batch.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+# retrieved from pytest fixtures
+if pytest.test_launcher not in pytest.wlm_options:
+    pytestmark = pytest.mark.skip(reason="Not testing WLM integrations")
+
+
+environ["OMP_NUM_THREADS"] = "1"
+try:
+    import ray
```
Spartee, line 22: The ``sys.module`` approach doesn't work in some cases for me.

see https://github.com/Spartee/SmartSim/blob/c00d8b301ea9dfb99dda4faba1dd83c3212ee739/tests/backends/test_onnx.py#L9

tests/test_manifest.py
```diff
@@ -6,8 +7,15 @@
 from smartsim.control import Manifest
 from smartsim.database import Orchestrator
 from smartsim.error import SmartSimError
+from smartsim.exp.ray import RayCluster
 from smartsim.settings import RunSettings
 
+try:
+    import ray
+except ImportError:
```
Spartee, line 15: ``sys.modules`` doesn't work sometimes for me.

see how I did it here -> https://github.com/Spartee/SmartSim/blob/c00d8b301ea9dfb99dda4faba1dd83c3212ee739/tests/backends/test_onnx.py#L9

tests/with_ray/full_wlm/test_ray_slurm_batch.py
```diff
@@ -0,0 +1,71 @@
...
+
+environ["OMP_NUM_THREADS"] = "1"
+try:
+    import ray
+except ImportError:
+    pass
+
+
+pytestmark = pytest.mark.skipif(
+    ("ray" not in sys.modules),
```
Spartee, line 27: See other comments about ``sys.modules``

tests/with_ray/on_wlm/test_ray_pbs.py
```diff
@@ -0,0 +1,74 @@
...
+
+environ["OMP_NUM_THREADS"] = "1"
+try:
+    import ray
+except ImportError:
+    pass
+
+
+pytestmark = pytest.mark.skipif(
+    ("ray" not in sys.modules),
```
Spartee, line 28: see other comments about sys.modules

tests/with_ray/on_wlm/test_ray_slurm.py
```diff
@@ -0,0 +1,121 @@
...
+
+environ["OMP_NUM_THREADS"] = "1"
+try:
+    import ray
+except ImportError:
+    pass
+
+
+pytestmark = pytest.mark.skipif(
+    ("ray" not in sys.modules),
```
Spartee, line 28: see other comments about sys.modules

tests/with_ray/test_ray.py
```diff
@@ -0,0 +1,99 @@
...
+environ["OMP_NUM_THREADS"] = "1"
+try:
+    import ray
+    import ray.util
+except ImportError:
+    pass
+
+pytestmark = pytest.mark.skip(reason="Local launch is currently disabled for Ray")
+
+# pytestmark = pytest.mark.skipif(
```
Spartee, line 24: Assuming we are leaving this in as a future?

tutorials/05_starting_ray/05_starting_ray_builtin.ipynb
```diff
@@ -0,0 +1,227 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
```
Spartee, line 5: Should we include a simple example of port forwarding here? I knew how to do it, but others might not.

Spartee (changes requested):
Looks great. Some small things primarily in testing and defaults.

smartsim/exp/ray/raycluster.py
al-rigazzi: Yeah, I left it till the last moment to make sure we did not forget any suggestion.

tests/with_ray/full_wlm/test_ray_slurm_batch.py
al-rigazzi: OK, updated on all Ray tests.

tests/with_ray/full_wlm/test_ray_pbs_batch.py
al-rigazzi: OK, updated on all Ray tests.

tests/with_ray/on_wlm/test_ray_pbs.py
al-rigazzi: OK, updated on all Ray tests.

tests/with_ray/on_wlm/test_ray_slurm.py
al-rigazzi: OK, updated on all Ray tests.

smartsim/exp/ray/raycluster.py
al-rigazzi: No, this parses from the head node log.

smartsim/exp/ray/raycluster.py
al-rigazzi: Good idea.

tutorials/05_starting_ray/05_starting_ray_builtin.ipynb
al-rigazzi: Should I write it as text? Because I usually do this with `ssh`.

PR #51, opened 2021-05-28
Fix to documentation

EricGustin:
## Description
There were two locations where the `run_settings` parameter for `Experiment.create_model` and  `Model.__init__` was documented as being of type dictionary when it should have been documented as having the type RunSettings. This was due to the documentation not being entirely updated when the concept of RunSettings was introduced in the 0.3.0 release.


mellis13 (approved):
Looks good -- thanks!

PR #52, opened 2021-06-03
WLMLauncher

Spartee:
## Description

This PR introduces the ``WLMLauncher`` parent class from which ``SlurmLauncher``, ``PbsLauncher`` and ``CobaltLauncher`` now inherit from. 

This change was motivated by a big fix in the PBS launcher where a managed and unmanaged job launched at the same time could cause a race-condition. Status updates are now returned with job name so order of arrival doesn't matter. While this is a relatively small change, it should address any problems users where encountering running the scaling test suite on PBS or Cobalt systems.

## TODO
- [ ] Test on Theta and Theta-GPU


al-rigazzi (approved):
LGTM

PR #55, opened 2021-06-16
Remove usage of redis-py-cluster

ctandon11:
Remove usage of redis-py-cluster

The orchestrator utilized a third-party library called redis-py-cluster to check the cluster status after initialization

this has been replaced with SmartRedis such that the check_cluster_status function now utilizes the SmartRedis client

### DIFF
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -139,8 +139,12 @@ class Orchestrator(EntityList):
         :raises SmartSimError: If cluster status cannot be verified
         """
         # pick a single host and port
-        host = self.hosts[0]
-        port = self.ports[0]
+        try:
+            host = self.hosts[0]
+            port = self.ports[0]
+        except:
+            raise SmartSimError("Hosts and ports are not populated")
+
         address = ":".join((host, str(port)))
         client = Client(address=address, cluster=True)
 
@@ -151,6 +155,7 @@ class Orchestrator(EntityList):
             time.sleep(2)
             try:
                 client.put_tensor("cluster_test", np.array([1, 2, 3, 4]))
+                receive_tensor = client.get_tensor("cluster_test")
                 logger.debug("Cluster status verified")
                 return
             except RedisReplyError:
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -31,8 +31,9 @@ import sys
 import time
 from os import getcwd
 
-from rediscluster import RedisCluster
-from rediscluster.exceptions import ClusterDownError
+import numpy as np
+from smartredis import Client
+from smartredis.error import RedisReplyError
 
 from ..config import CONFIG
 from ..entity import DBNode, EntityList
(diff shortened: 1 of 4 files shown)


smartsim/database/orchestrator.py
```diff
@@ -132,32 +133,27 @@ def create_cluster(self):  # cov-wlm
...
-        # TODO use smartredis for this, then we don't have to create host dictionary
-        host_list = []
-        for host in self.hosts:
-            for port in self.ports:
-                host_dict = dict()
-                host_dict["host"] = host
-                host_dict["port"] = port
-                host_list.append(host_dict)
+        # pick a single host and port
+        host = self.hosts[0]
```
Spartee, line 142: What should we do if a user calls this method but the hosts and ports are not populated? like when the orchestrator is not launched. I tend to think we should have a try-except block here and throw a ``SmartSimError`` if hosts and ports are not populated.

this would also catch if the node aquisition strategy failed and ``check_cluster_status`` is still called.

smartsim/database/orchestrator.py
```diff
@@ -132,32 +133,27 @@ def create_cluster(self):  # cov-wlm
...
         trials = 10
         logger.debug("Beginning database cluster status check...")
         while trials > 0:
             # wait for cluster to spin up
             time.sleep(2)
             try:
-                redis_tester = RedisCluster(startup_nodes=host_list)
-                redis_tester.set("__test__", "__test__")
-                redis_tester.delete("__test__")
+                client.put_tensor("cluster_test", np.array([1, 2, 3, 4]))
```
Spartee, line 153: I think we will want to use ``delete_tensor`` once that is put in to SmartRedis in https://github.com/CrayLabs/SmartRedis/pull/66

for now though, we should ensure that the tensor is actually present in the database and can be retrieved so lets add a ``client.get_tensor("cluster_test")`` here

Spartee (changes requested):
Some comments. So the reason this PR failed the github tests is because SmartRedis was not listed in the dependencies for SmartSim. there are a few places to specify this.

first, in ``setup.cfg``, under ``install_requires`` you will want to take out the ``redis-py`` and ``redis-py-cluster`` dependencies and replace them with ``smartredis==0.1.1``. This will download SmartRedis whenever Smartsim is installed.

There are also a ``requirements.txt`` and ``requirements-dev.txt`` file in the repo that need to have the same change as the ``setup.cfg`` file.

Otherwise, congrats on your first PR!! :)

smartsim/database/orchestrator.py
```diff
@@ -132,32 +133,32 @@ def create_cluster(self):  # cov-wlm
         self.check_cluster_status()
         logger.info(f"Database cluster created with {str(len(self.hosts))} shards")
 
-    def check_cluster_status(self):  # cov-wlm
+    def check_cluster_status(self):
```
al-rigazzi, line 136: Just a question, why did we remove ` # cov-wlm`?
The reason why we had it was that this is a function which can only be tested when running on WLM, did this change (i.e. can we call this function when running locally)?

al-rigazzi (commented):
Looks good! I only have one question which I added as a comment.

smartsim/database/orchestrator.py
Spartee: @ctandon11 ?

ctandon11: @al-rigazzi @Spartee you are right, this function still cannot be run locally. I can add `# cov-wlm` back in.

Spartee (approved):
LGTM! 👍🏻

PR #58, opened 2021-06-30
Edit Makefile and test fix

Spartee:
Small edits to tests introduced in PR #53 that left some testing files around.

Also introduces a new testing flag in the makefile for users who want to run the test suite on interactive allocations without the ``full_wlm`` tests which assume full control over the host machine.


EricGustin (approved):
looks good!

Issue #54, opened 2021-06-10
Function in Orchestrator that returns database address


EricGustin:
# Description
Currently the `Orchestrator`'s hosts are a hidden property. The function to be added should return the hosts with their ports in a formatted manner.

# Justification
Implementing this function will allow for better interaction with the `Orchestrator`. By allowing the user to retrieve addresses in a formatted way, this will create an easier way to send data to the database.

# Implementation Strategy
The function should be something like:

```python
class Orchestrator(EntityList):

... 

    @property
    def address(self):
          if not self._hosts:
               raise SmartSimError()
          addresses = []
          for host, port in itertools.product(self._hosts, self._ports):
             addresses.append(":".join((host, port)))
      return addresses
```

PR #60, opened 2021-07-02
Orchestrator Function that returns DB address

ctandon11:
Addresses ticket #54. 

The Orchestrator's `hosts` and `ports` were
hidden properties.

`_get_address()` now allows the user to retrieve
database addresses in a formatted way.


smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,19 @@ def check_cluster_status(self):  # cov-wlm
         if trials == 0:
             raise SmartSimError("Cluster setup could not be verified")
 
+    def _get_address(self):
```
Spartee, line 168: Lets make this a public method since we want users to find this. In Python, the underscore is typically used to denote methods that should only be used internal to the class.

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,19 @@ def check_cluster_status(self):  # cov-wlm
...
+        """Return database addresses
+
+        :return: addresses
+        :rtype: list[str]
+        """
+        if not self._hosts:
+            raise SmartSimError()
```
Spartee, line 175: I like this conditional, we should probably have an error message though so the user isn't left without knowing what happened. maybe something like, "Could not find database address".

Spartee (changes requested):

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,42 @@ def check_cluster_status(self):  # cov-wlm
...
+        if len(self._hosts) > 1:
+            self.check_cluster_status()
+        else:
+            self.is_active()
+        addresses = []
+        for host, port in itertools.product(self._hosts, self.ports):
+            addresses.append(":".join((host, port)))
+        return addresses
+
+    def is_active(self):
```
Spartee, line 185: I think this function should return True or False instead of throwing an error.

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,42 @@ def check_cluster_status(self):  # cov-wlm
...
+        """Check if database is running
+
+        :raise SmartSimError: if database cannot be verified
+        """
+        try:
+            host = self._hosts[0]
+            port = self.ports[0]
+        except SmartSimError as e:
+            raise SmartSimError("Database is not active") from e
```
Spartee, line 194: nice job on the re-raise here!

Spartee (changes requested):

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,43 @@ def check_cluster_status(self):  # cov-wlm
...
+    def is_active(self):
+        """Check if database is running
+
+        :returns: True if database is active, False otherwise
+        :rtype: bool
+        """
+        try:
+            host = self._hosts[0]
+            port = self.ports[0]
+        except SmartSimError as e:
```
Spartee, line 194: should this not be ``return False``?

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,43 @@ def check_cluster_status(self):  # cov-wlm
...
+        return addresses
+
+    def is_active(self):
+        """Check if database is running
+
+        :returns: True if database is active, False otherwise
+        :rtype: bool
+        """
+        try:
+            host = self._hosts[0]
```
Spartee, line 192: what happens if this is called on an ``Orchestrator`` object that hasn't been launched yet?

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,43 @@ def check_cluster_status(self):  # cov-wlm
         if trials == 0:
             raise SmartSimError("Cluster setup could not be verified")
 
+    def get_address(self):
+        """Return database addresses
+
+        :return: addresses
+        :rtype: list[str]
```
Spartee, line 172: we should add a ``:raises SmartSimError: <fill in description for error>`` here

Spartee (changes requested):
We should add some test cases for these! this will help us work through edge cases.

smartsim/database/orchestrator.py
ctandon11: wouldn't `if not self._hosts: raise SmartSimError(...)` in `get_address()` account for that before `is_active()` is even called?

Spartee: Since this method is user facing, the user should also be able to call this directly. 

Ex. 

'''Python
db = Orchestrator()
db.is_active()
'''

ctandon11: gotcha. so if I add the same check in `is_active()` that should account for this, right?

smartsim/database/orchestrator.py
```diff
@@ -165,6 +165,44 @@ def check_cluster_status(self):  # cov-wlm
...
+            return False
+        host = self._hosts[0]
+        port = self.ports[0]
+        address = ":".join((host, str(port)))
+        client = Client(address=address, cluster=False)
+        try:
+            client.put_tensor("db_test", np.array([1, 2, 3, 4]))
+            receive_tensor = client.get_tensor("db_test")
+            return True
+        except:
```
Spartee, line 203: So we don't want to accept all exception types here. 

well want to import the ``RedisReplyError`` from SmartRedis like this 

```python
from smartredis.error import RedisReplyError, RedisConnectionError

....
        address = ":".join((host, str(port)))
        client = Client(address=address, cluster=False)
        try:
            client.put_tensor("db_test", np.array([1, 2, 3, 4]))
            receive_tensor = client.get_tensor("db_test")
            return True
        except (RedisReplyError, RedisConnectionError):
            return False
```

Spartee (changes requested):
Small change! otherwise looks good! after this, this will be able to go in!

PR #61, opened 2021-07-03
Manifest and preamble

### LINKED ISSUE #37: Batch Script preamble
# Description
Add the ability for users to input lines into the batch script generated by the ``Step._write_script``. Effectively, this is so users can do things like ``module load xyz`` and ``conda activate env``. 

# Justification
Users needing to load specific environments or modules prior to execution. 

# Implementation Strategy
Add the following to ``BatchSettings`` base class.
 - ``_preamble`` - list of str, created by calling ``add_preamble``
 - ``add_preamble(self, line)`` - adds string parameter line to ``self._preamble``

Then we need to add it to the ``_write_script`` method of Step. There are two ways I think this could be done. 

Create a subclass of Step called ``BatchStep`` and have all the batch step implementations inherit from ``BatchStep``.
 - ``QsubBatchStep``
 - ``SbatchStep``
 - ``CobaltBatchStep``
then implement a ``add_preamble`` method to the ``BatchStep`` class so it gets inherited from each. Should probably do some santization of user input. This would probably be the best long term

OR

implement it in each batch ``Step`` class individually. 

Additionally this might need to consider which shell the user is using.

al-rigazzi:
This PR addresses the proliferation of different entity types and corresponding behaviors.

* Implemented `Manifest` object which acts as a container of deployables with checks and filters.
* Added `preamble` to `BatchSettings` to allow users to add non-WLM lines to batch scripts

Closes #37


smartsim/control/controller.py
```diff
@@ -273,6 +268,8 @@ def _launch(self, entities, entity_lists, orchestrator):
 
         # create all steps prior to launch
         steps = []
+        entity_lists = manifest.ensembles
+        entities = manifest.models
         if entity_lists:
```
Spartee, line 273: I think since the manifest returns empty lists we can just use ``manifest.ensembles`` and ``manifest.models`` directly in the conditional.

smartsim/control/manifest.py
```diff
@@ -0,0 +1,102 @@
...
+        self._check_types(self._deployables)
+        self._check_names(self._deployables)
+
+    @property
+    def db(self):
+        _db = None
+        for deployable in self._deployables:
+            if isinstance(deployable, Orchestrator):
+                if _db:
+                    raise SmartSimError("Manifest was given two orchestrators")
```
Spartee, line 58: Should the user know about the manifest object? To me it seems like an internal representation. Would this error be confusing?

smartsim/control/manifest.py
```diff
@@ -0,0 +1,102 @@
...
+        return _ensembles
+
+    def _check_names(self, deployables):
+        used = []
+        for deployable in deployables:
+            name = getattr(deployable, "name", None)
+            if not name:
+                raise AttributeError(f"Entity has no name. Please set name attribute.")
+            if name in used:
+                raise SmartSimError("User provided two entities with the same name")
```
Spartee, line 91: When I refactor the jobManager with the different classes for jobs, we should put in a check where the manifest is passed to the JM to see if any new entities are defined by the same name

smartsim/generation/generator.py
```diff
@@ -80,12 +80,13 @@ def generate_experiment(self, *args):
...
         e.g. ``THERMO=;90;``
+
         """
-        entities, entity_lists, orchestrator = separate_entities(args)
+        generator_manifest = Manifest(*args)
         self._gen_exp_dir()
-        self._gen_orc_dir(orchestrator)
-        self._gen_entity_list_dir(entity_lists)
-        self._gen_entity_dirs(entities)
+        self._gen_orc_dir(generator_manifest.db)
```
Spartee, line 87: Should we just pass the ``Manifest`` object itself?

Spartee (commented):
Looks much cleaner. Lets make sure Coverage is still over 90%. Tests passed for me.

smartsim/control/manifest.py
al-rigazzi: OK, maybe I can change the message to "Attempted to create two orchestrators"?

smartsim/control/manifest.py
al-rigazzi: OK, it makes sense to me.

smartsim/generation/generator.py
al-rigazzi: OK, I like the idea of unifying the interface.

smartsim/control/controller.py
al-rigazzi: Yes, we can do this. The only reason I did not was that currently, every time we call `manifest.ensembles` or `manifest.models`, the whole list of deployable is searched and filtered (i.e. the separate lists are not maintained, always recomputed). Therefore, doing the check and then iterating over the lists means computing them twice. Anyhow, given that I don't think the user will create lists of millions of deployables, this shouldn't impact performances.

al-rigazzi: NVM, we can totally skip the conditional, as we are iterating over the resulting list. If the list is empty, the iteration will simply not be run.

smartsim/generation/generator.py
al-rigazzi: @Spartee I'm actually going to push back on this. `_gen_entity_dirs` is used in a slightly different way (because it also has an optional parameter) and passing a `Manifest` to it would generate duplicates.

Spartee: Should we edit the generator code then to make the interface cleaner? My guess is we could make some of the code more re-usable? Maybe not in this ticket but file a future one? Im going to leave that up to you.

smartsim/settings/alpsSettings.py
```diff
@@ -102,7 +102,7 @@ def set_hostlist(self, host_list):
             raise TypeError("host_list argument must be a list of strings")
         if not all([isinstance(host, str) for host in host_list]):
             raise TypeError("host_list argument must be list of strings")
-        self.run_args["nodelist"] = ",".join(host_list)
+        self.run_args["node-list"] = ",".join(host_list)
```
Spartee, line 105: thanks

Spartee (approved):
Minor comments, LGTM otherwise. good to go assuming coverage is good.

I did not re-run the tests this review since these were minor changes, fyi.

al-rigazzi:
@Spartee thanks for the review. I'll add a flag for lsf-related code, to keep the coverage rate of the rest of the library untouched.

al-rigazzi:
@Spartee ok, I'll check coverage, but it should be unaffected. Thanks for the review.

PR #67, opened 2021-07-16
Fix orchestrator bugs introduced in 98341ca

Spartee:
There were two problems with the ``get_address`` and ``is_active`` implmentations

1. itertools was not imported
2. ports are integers and hence cause the string formation to fail.

New tests were added that increase ``Orchestrator`` coverage to 99%. 

This was my fault for not catching this in review.


al-rigazzi:
Just as a curiosity, which part of `98341ca` introduced these bugs?

Spartee:
@al-rigazzi see the PR description.

PR #68, opened 2021-07-16
Add a NCAR casper system example 

jedwards4b:
Newer versions of PBS include a feature create_resv_from_job, this feature allows a user to:

- Launch a job that creates a reservation for multiple nodes - possibly a combination of cpu and gpu nodes. 
- Launch within that reservation the SmartSim database and one or more client jobs.


tutorials/casper/Makefile
```diff
@@ -0,0 +1,17 @@
+REDIS_HOME = $(HOME)/sandboxes/SmartRedis
```
Spartee, line 1: ``SMARTREDIS_HOME`` might disentangle the two references, SmartRedis and Redis(the database)

tutorials/casper/README.md
```diff
@@ -0,0 +1,57 @@
+
+# Casper 
+
+```bash
+module purge
+module load gnu/9.1.0 ncarcompilers openmpi netcdf ncarenv cmake
+```
```
Spartee, line 7: might want to put a note in that this does NOT work with gcc 10 (for future reference)

tutorials/casper/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+I also needed a newer version of gmake, it's in /glade/work/jedwards/make-4.3/bin/make
```
Spartee, line 9: ``conda install make`` will also provide a new version of ``make`` for the user that will satisfy this requirement

tutorials/casper/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+I am using a python environment created with:
+```
+ncar_pylib -c 20201220 /glade/work/$USER/casper_npl_clone
+```
+
+``pip install smartsim``
+
+``smart --device gpu``
```
Spartee, line 18: please note that this also could be ``smart --device cpu`` in the case where the user would like to run the database (specifically ML) on cpu nodes

tutorials/casper/utils.py
```diff
@@ -0,0 +1,85 @@
+import subprocess, os, io
+
+def _convert_to_fd(filearg, from_dir, mode="a"):
+    filearg = _get_path(filearg, from_dir)
+
+    return open(filearg, mode)
+
+_hack=object()
```
Spartee, line 8: Will you explain the usage here?

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+        if int(jobs_left[0]) + int(jobs_left[1]) == 1:
+            jobs_done = True
+        else:
+            time.sleep(60)
+
+
+
+
+
+# create the experiment and specify PBS because cheyenne is a PBS system
```
Spartee, line 94: casper?

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+
+import socket, subprocess
+import numpy as np
+from utils import run_cmd
+from smartsim import Experiment, constants
+from smartsim.database import PBSOrchestrator
+
+"""
+Launch a distributed, in memory database cluster and use the
+SmartRedis python client to send and recieve some numpy arrays.
```
Spartee, line 26: I don't think it does this part anymore

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+#PBS  -V
+#PBS  -l walltime=$walltime
+#PBS  -A $account
+##PBS  -q regular
+#PBS  -V
+#PBS  -l select=$db_nodes:ncpus=1:ompthreads=1:mpiprocs=1$ngpus
+
+import os, sys, time
+
+# The python environment is not passed properly to submitted jobs on casper
```
Spartee, line 14: does this problem go away with a more typical ``conda`` based approach?

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+    exp.start(db, block=True)
+
+    # get the status of the database
+    statuses = exp.get_status(db)
+    print(f"Status of all database nodes: {statuses}")
+
+    return db
+
+def monitor_client_jobs(rsvname):
+    jobs_done=False
```
Spartee, line 80: Might want to put a first sleep here instead in the case where jobs may not have arrived yet given the NCAR scheduler delay?

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+    print(f"Starting Orchestrator on hosts: {db_hosts}")
+    # batch = False to launch on existing allocation
+    db = PBSOrchestrator(port=port, db_nodes=len(db_hosts), batch=False,
+                          run_command="mpirun", hosts=db_hosts)
+
+    # generate directories for output files
+    # pass in objects to make dirs for
+    exp.generate(db, overwrite=True)
+
+    # start the database on interactive allocation
```
Spartee, line 70: "inside the batch allocation"

tutorials/casper/launch_database_cluster.template
```diff
@@ -0,0 +1,109 @@
...
+db = launch_cluster_orc(exp, db_hosts, db_port)
+
+rsvname = os.environ["RSVNAME"]
+# stay alive until client jobs have completed
+monitor_client_jobs(rsvname)
+
+# shutdown the database because we don't need it anymore
+exp.stop(db)
+# delete the job reservation
+run_cmd("pbs_rdel {}".format(rsvname))
```
Spartee, line 109: @al-rigazzi @mellis13 we used to have launcher functions for PBS reservations but since many systems didn't have it we took it out. do you think we should reintroduce them?

tutorials/casper/README.md
```diff
@@ -0,0 +1,57 @@
...
+  --member-nodes MEMBER_NODES
+                        Number of nodes per ensemble member, default=1
+  --account ACCOUNT     Account ID
+  --db-port DB_PORT     db port, default=6780
+```
+It creates pbs jobs from each of the 3 templates
+1. resv_job.template
+2. launch_database_cluster.template
+3. launch_client.template
+
```
Spartee, line 51: Possibly a note here that this is not the expected way for users to launch models/infra through SmartSim and that this is a great example of how to launch both the ML infra(database) and simulation separately.

tutorials/casper/README.md
```diff
@@ -0,0 +1,57 @@
...
+``pip install smartredis``
+
+First you need to build the smartredis_put_get_3D.F90 fortran example:
+```
+make
+```
+
+launch.py is the primary launch script 
+```
+usage: launch.py [-h] [--db-nodes DB_NODES] [--ngpus-per-node NGPUS_PER_NODE]
```
Spartee, line 29: I think we may want to consider some SmartSim functions similar to this where the database is just launched with no job management and interactivity. It just launches itself and then it's all up to the user. 

@al-rigazzi what do you think about this?

Spartee (commented):
Get stuff @jedwards4b. As discussed I think we will want to move this code to a seperate repository of SmartSim/NCAR examples and then add a README to the tutorials section that links to that repo and the others that will be created soon.

Minor comments and questions. Also some questions for @mellis13 and @al-rigazzi

tutorials/casper/README.md
jedwards4b: It does work with gnu 10 and gnu 11.  I've removed the version number.

tutorials/casper/Makefile
jedwards4b: I've refactored this to use the installed SmartRedis module.

tutorials/casper/README.md
jedwards4b: My understanding is that --device gpu also builds a cpu version, this seems to be confirmed by testing.

tutorials/casper/utils.py
jedwards4b: cleaned up

tutorials/casper/README.md
Spartee: SmartRedis, possibly, SmartSim backends will not.

see https://github.com/RedisAI/RedisAI/issues/777 

Confirmed by RedisAI devs as well.

PR #72, opened 2021-08-01
Add condition to redis IP sleep

al-rigazzi:
When we parse RedisIP info, we always sleep 5 seconds, even if we found the host. This means that if we have N nodes, we wait for N*5 seconds, but most likely, all nodes will have written their info when we parse it for one. This quick fix removes the sleep operation when we find host (but leaves it otherwise).


Spartee (approved):
LGTM works well.

PR #73, opened 2021-08-06
Redis starter script for Orchestrator

Spartee:
Recently it was discovered that redis will often change the IP address of a shard unless it is bound to a specific IP address.

To address this, each Orchestrator has been adapted to use the ``redis_starter.py`` script that binds the redis instance to a specific address. Essentially this is a Python script that takes in a command and does a lookup on the network interface specified by the user in order to bind the server to a specific address.

Also in this PR is a change back to redis-py and redis-py-cluster for the orchestrator functions ``is_active`` and ``check_cluster_status``. This was after a bug was discovered in clang/pybind/cray PrgEnv. As there isn't much we can do about the bug at the moment we now require both dependencies in the library.

TODO
 - [x] Adapt LSF
 - [ ] Test on Theta
 - [ ] Test on Swan
 - [x] Test on Osprey (have to specific "ib0" for orchestrator tests)
 - [x] Test on Cheyenne/Casper

All tests pass (gpu and cpu) on Horizon and MacOS.


al-rigazzi:
@Spartee question: how does this work with RedisIP? I see on Summit that the IP returned by RedisIP is not the one we are binding to, do we need to update it?

al-rigazzi:
Would it be possible to add an interface to be used for tests as we do for launchers and accounts? So that a user could do something along the lines of `export SMARTSIM_TEST_INTERFACE=ib0` and run tests?

PR #74, opened 2021-08-09
Lsf redis starter

al-rigazzi:
This PR modifies the LSF Orchestrator so that it uses the new `redis_starter.py` script to start the DB nodes.

It also includes a new LSF option for `jsrun`, that is `--individual_stdio`, which allows MPMD scripts to write output (and error) to separate files.


smartsim/database/lsfOrchestrator.py
```diff
@@ -48,7 +47,7 @@ def __init__(
         project=None,
         time=None,
         db_per_host=1,
-        host_map=None,
+        interface="ib0",
```
Spartee, line 50: add interface to docstring

smartsim/launcher/step/lsfStep.py
```diff
@@ -123,6 +123,22 @@ def __init__(self, name, cwd, run_settings):
         if not self.run_settings.in_batch:
             self._set_alloc()
 
+
+    def get_output_files(self):
+        """Return two paths to error and output files based on cwd"""
+        output = self.get_step_file(ending=".out")
+        error = self.get_step_file(ending=".err")
+
+        if self.run_settings.individual_suffix:
```
Spartee, line 132: lets do some explaining here in comments. be verbose because this doesn't make any sense

Spartee (commented):
Looks great, a couple comments.

I will test this on summit.

smartsim/launcher/step/lsfStep.py
al-rigazzi: OK

al-rigazzi:
This was merged in another branch.

PR #76, opened 2021-08-11
Add new FindTensorFlow.cmake

al-rigazzi:
This PR patches a problem RedisAI has to find pre-installed TF. 

We store a modified version of RAI's `FindTensorFlow.cmake` file, where we add TF 2.4 (which covers all patch releases) to the accepted TF versions. Moreover, we change the library name (the shared object file has changed). 

Notice this was solely tested on Summit (and it's our current solution on that system).


Spartee (approved):
LGTM

PR #79, opened 2021-08-13
Fixes for summit build

al-rigazzi:
Summit users found that the installation instructions were incomplete. Moreover, some files were not copied with `pip install .` and this made building the TF backend impossible.

This PR updates both the docs and the usage of our custom module file for CMake.


Spartee (approved):
LGTM 👍🏻

doc/installation.rst
```diff
@@ -427,11 +427,14 @@ that SmartSim needs already built into it. Users can follow these instructions
...
-
   # setup Python and build environment
   module load open-ce
-  conda create -p /ccs/home/<USERNAME>/.conda/envs/smartsim --clone open-ce-1.2.0-py38-0
+  conda create -p /ccs/home/$USER/.conda/envs/smartsim --clone open-ce-1.2.0-py38-0
   conda activate smartsim
+  # fix broken cmake module
+  pip uninstall cmake
+  pip install cmake
+  conda install git-lfs make -y
```
Spartee, line 437: I think we should add ``git-lfs install`` here.

Spartee (approved):
LGTM 👍🏻

doc/installation.rst
al-rigazzi: I've added it to the instructions.

PR #81, opened 2021-09-07
Managed JsrunSetting 

al-rigazzi:
This PR introduces managed Jsrun steps. It uses jslist to do this. This mechanism works well as long as SmartSim is the only program running jobs inside an allocation, otherwise, race conditions could break the mechanism.
This change (along with others included in this PR) is due to new behavior of LSF on Summit.
Currently, it is also not possible to run more than one application on the same host (through ERF/MPMD), but this does not seem to be a SmartSim problem (simple tests without SmartSim fail too).


al-rigazzi:
I'm closing this PR for the moment, because something is wrong with the diff

Issue #84, opened 2021-09-09
Use base RunSettings in non-local launchers


Spartee:
# Description
Some application binaries call parallel launcher commands themselves within their codebase. SmartSim currently enforces each ``RunSettings`` child class to use one launcher like ``srun`` or ``aprun``.

# Justification
there are three use cases that will benefit from this. 
1. Applications that internalize the launcher commands as stated above.
2. Ensembles that wish to run a serial application either in between or in conjunction with parallel code.
3. Users that have a custom launch command SmartSim doesn't support (e.g. ``mpiexec``)

# Implementation Strategy
All the launchers have a create step function that looks something like

(from ``smartsim/smartsim/launcher/slurm/slurmLauncher.py``)
```python
    def create_step(self, name, cwd, step_settings):
        """Create a Slurm job step
        :param name: name of the entity to be launched
        :type name: str
        :param cwd: path to launch dir
        :type cwd: str
        :param step_settings: batch or run settings for entity
        :type step_settings: BatchSettings | RunSettings
        :raises SSUnsupportedError: if batch or run settings type isnt supported
        :raises LauncherError: if step creation fails
        :return: step instance
        :rtype: Step
        """
        try:
            if isinstance(step_settings, SrunSettings):
                step = SrunStep(name, cwd, step_settings)
                return step
            if isinstance(step_settings, SbatchSettings):
                step = SbatchStep(name, cwd, step_settings)
                return step
            if isinstance(step_settings, MpirunSettings):
                step = MpirunStep(name, cwd, step_settings)
                return step
            raise SSUnsupportedError("RunSettings type not supported by Slurm")
        except SSConfigError as e:
            raise LauncherError("Step creation failed: " + str(e)) from None
```

Adding ``RunSettings`` to this function will enable all three justifications above. All of these commands will run as `unmanaged` workflows (e.g. not reliant on workload manager for status updates). 

The only caveat is that if the parallel launch command used to run the workload spawns a process that is non-local or disapears after a remote process has been started, then the status updates will not work and report as completed. Not sure how we can get around this for custom processes other than to provide a debug warning.

PR #90, opened 2021-09-24
Add base RunSettings support for slurm, pbs, and cobalt

EricGustin:
This PR addresses issue #84. The SmartSim `SlurmLauncher`, `PBSLauncher`, and `CobaltLauncher` now support base `RunSettings`. 

Tested on slurm and pbs.


al-rigazzi (commented):
I added a comment, but in general it looks good.

PR #103, opened 2021-10-22
Ray on Cobalt

al-rigazzi:
This PR adds Cobalt support to the Ray launcher. It also includes some minor tweaks to tests (to comply to new Ray interface).

The Cobalt tests only use aprun and were run on Theta.


smartsim/exp/ray/raycluster.py
```diff
@@ -442,22 +522,35 @@ def _build_ray_exe_args(self, ray_password, ray_args, ray_port, interface):
 
         return " ".join(ray_starter_args)
 
-    def _build_run_settings(self, launcher, alloc, run_args, ray_exe_args):
+    def _build_run_settings(self, launcher, alloc, run_args, ray_exe_args, run_command):
```
Spartee, line 525: I think we could safely migrate this to use the ``settings.create_run_settings`` method. and the also the ``create_batch_settings`` method when that goes in.

tests/with_ray/on_wlm/test_ray_pbs.py
```diff
@@ -55,7 +55,7 @@ def test_ray_launch_and_shutdown(fileutils, wlmutils, caplog):
 
     exp.generate(cluster)
     exp.start(cluster, block=False, summary=False)
-    ctx = ray.client(cluster.get_head_address() + ":10001").connect()
+    ctx = ray.client("ray://" + cluster.get_head_address() + ":10001").connect()
```
Spartee, line 58: should we just be including this ``ray://`` bit in the ``cluster.get_head_address()`` function? port also?

Any reason not to?

Spartee (commented):
Some comments, lmk what you think.

tests/with_ray/on_wlm/test_ray_pbs.py
al-rigazzi: I think it is better to keep it like this, because if it is local, the port is different (and you don't add `ray://`).

al-rigazzi:
Closing this in favor of new `create_settings` approach.

PR #104, opened 2021-10-26
Fix Slurm handling of comma-separated env vars

al-rigazzi:
This is a minimal-modification PR. Whenever an env var is found to contain a comma, we set it locally to the `srun` process and add the name of then env var (not the value!) to the list of exported env vars. Its value is then correctly passed by `srun` to the launched application. This does not change anything in how other env variables are exported/formed/formatted and the result is still human-readable.

Basically, instead of trying to run the unsuccessful command:
```bash
srun --export SSKEYIN=name_0,name_1,OTHER_VAR=value python script.py
```
we run
```bash
SSKEYIN=name_0,name_1 srun --export SSKEYIN,OTHER_VAR=value python script.py
```


Spartee (approved):
LGTM, glad to have this fix in.

Issue #40, opened 2021-04-19
Tagged parameters on the same line of file within generator files


al-rigazzi:
# Description
If there are two tagged parameters on the same line (e.g. `ray start --head --port=;port; --redis-password=;password;`), the two parameters get collapsed to a single one (in the same example, we may end up with `ray start --head --port=6379`). It seems that the substitution is implicitly assuming that there is only a single parameter per line (which makes sense in standard config files, but not for other use cases of generation/configuration).

# How to reproduce
Put in a `template.sh` file:
```bash
echo "My two parameters are ;port; and ;password;, OK?"
```
and then in SmartSim run:
```python
from smartsim import Experiment
from smartsim.settings import RunSettings

exp = Experiment("param_test")
model_params = {"port": 6379, "password": "unbreakable_password"}
model_settings = RunSettings("bash", "template.sh")
parameterized_model = exp.create_model("param-model", path="./param_test",
                                   run_settings=model_settings, params=model_params)
parameterized_model.attach_generator_files(to_configure=['./template.sh'])
exp.generate(parameterized_model, overwrite=True)
exp.start(parameterized_model)
```

The result will be
```bash
My two parameters are 6379, OK?
```

# Expected behavior
Every parameter should be correctly substituted and the result should be:
```bash
My two parameters are 6379 and unbreakable_password, OK?
```

# System
 - OS: RHEL
 - Commit of library: c2b848b
 - Workload manager: local, but I guess any

PR #107, opened 2021-11-01
Improve parameter handling with and without generation

### LINKED ISSUE #16: Enable usage of ensemble/model params without generation
# Description
Include methods for using the ensemble generated parameter sets as executable arguments for each underlying model instead of relying on generation to write them into files.

# Justification
@neuralvis ran into this as he wanted to parameterize launches with the permutation strategies that generate params for each underlying model object. He wanted the params to be turned into executable arguments for each model.

# Implementation Strategy
``Ensemble`` and ``Model`` should both have a method ``add_params_to_exe_args`` that takes the params in each Model object and adds them to the ``RunSettings.exe_args`` used by each model.

al-rigazzi:
This PR enhances the current way parameters and models are generated in two ways:

- add the possibility of specifying `arg_params` in `Ensemble`s: `arg_params` are treated like `params` but set as `exe_args` in the created `Model`s -- thus they are set even without the generation step.
- allow multiple tagged parameters on same line

This PR resolves #16 #40

### DIFF
--- a/smartsim/entity/model.py
+++ b/smartsim/entity/model.py
@@ -24,27 +24,34 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from smartsim.error.errors import SSConfigError
 from ..error import EntityExistsError
-from ..utils.helpers import init_default
+from ..utils.helpers import cat_arg_and_value, init_default
 from .entity import SmartSimEntity
 from .files import EntityFiles
 
 
 class Model(SmartSimEntity):
-    def __init__(self, name, params, path, run_settings):
+    def __init__(self, name, params, path, run_settings, params_as_args=None):
         """Initialize a model entity within Smartsim
 
         :param name: name of the model
         :type name: str
-        :param params: model parameters for writing into configuration files.
+        :param params: model parameters for writing into configuration files or
+                       to be passed as command line arguments to executable.
         :type params: dict
         :param path: path to output, error, and configuration files
         :type path: str
         :param run_settings: launcher settings specified in the experiment
         :type run_settings: RunSettings
+        :param params_as_args: list of parameters which have to be
+                               interpreted as command line arguments to
+                               be added to run_settings
+        :type params_as_args: list[str]
         """
         super().__init__(name, path, run_settings)
         self.params = params
+        self.params_as_args = params_as_args
         self.incoming_entities = []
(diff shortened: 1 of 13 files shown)


Spartee:
A few questions about this implementation

1. Should ``params`` and ``arg_params`` be separate? Why?
2. If they should be separate, shouldn't ``Model`` also have arg_params?
3. What happens if a conflicting argument is placed in both ``arg_params`` and in ``params``?

My main concern is that this isn't entirely straightforward given the naming convention. 

Alternative approaches to consider:
- a flag, ``params_as_args`` that specifies if the ``params`` argument should be used as command line args
- a method, ``Ensemble.params_to_args()`` that calls ``Model.params_to_args`` on each model in the ensemble.

thoughts on those?

Either way I think ``Models`` should have this too.

al-rigazzi:
Thanks for the review, @Spartee!

I agree, naming is confusing. I think it is fine to just call all of them `params`, and then have a method that uses info given in `params_as_args` to generate the corresponding command line arguments. 

I'll adapt the code in the following way:
- all parameters are specified as `params`
- a list `params_as_args` has to be passed to specify whether a `param` has to be turned into an argument, this is then also set in all `Model`s in the `Ensemble`
- in `Ensemble._initialize_entitities()`, the function `params_to_args()` is called, and this is called on each `Model` in the `Ensemble`.

The only minor drawback is that `params_as_args` in `Model`, is not particularly useful when not running within an ensemble (it somehow just duplicates what can be done with `exe_args`, since a single model is not parameterized).

smartsim/entity/model.py
```diff
@@ -111,6 +118,18 @@ def attach_generator_files(self, to_copy=None, to_symlink=None, to_configure=Non
...
         self.files = EntityFiles(to_configure, to_copy, to_symlink)
 
+    def params_to_args(self):
+        """Convert parameters to command line arguments and update run settings.
+        """
+        for param in self.params_as_args:
+            if not param in self.params:
+                raise SSConfigError(f"Tried to convert {param} to command line argument " +
+                                    f"for Model {self.name}, but its value was not found in model params")
+            if self.run_settings is None:
```
Spartee, line 128: will this case ever happen?

Spartee (approved):
one comment but otherwise LGTM

smartsim/entity/model.py
al-rigazzi: @Spartee it seems like we don't check whether `run_settings` is `None`, thus a very bad user might initialize a `Model` with `run_settings=None`. I don't know if there is any reason why we are allowing this, otherwise we might want to guard against it in another part of the code?

Spartee: yea I think step creation would just fail, but when we go to the server arch I think thisll be more important.

PR #108, opened 2021-11-01
Add base run settings to tests for LSF

al-rigazzi:
This PR actually just adds the tests for LSF base settings, as they were already supported in the LSF launcher. Therefore, the status is now:
- in the LSF launcher, `RunSettings` objects are supported
- in `conftest.py`, `get_base_run_settings` can generate `RunSettings` for LSF.

Let me know if there was something else to be added!


EricGustin (approved):
Given that `create_step` for `lsflauncher` was previously moved up to the base `launcher`, these tests look good to me.

Issue #91, opened 2021-09-29
Experiment.create_run_settings


Spartee:
# Description
When writing a SmartSim driver script, ``RunSettings`` are often created to define how an entity is to be executed. These are largely convenience wrappers around parallel run commands.  

### Currently Supported
 - ``MpirunSettings`` for ``mpirun``
 - ``AprunSettings`` for ``aprun``
 - ``JsrunSettings`` for ``jsrun``
 - ``SrunSettings`` for ``srun``
 - ``RunSettings`` for custom and serial launches

# Justification
Users that wish to make their SmartSim scripts more portable to various systems that use different workload managers will benefit from these because it will perform a similar switch case that we have seen implemented in a few cases.

# Implementation Strategy
 
A single function should be created with abstract arguments that apply generally to each of the various run command types.

```python
class Experiment:
   ...

   def create_run_settings(self, exe, exe_args, run_args, env_vars, **kwargs):
     
```
This function will use the fact that the Experiment knows the launcher type and will:
 - determine which ``RunSettings`` child class to use based on the launcher type
 - initialize a new instance of that class
 - return it.

## to think about
- functions like ``RunSettings.set_cpus_per_task`` will need to be unified accross all classes if not already.

PR #110, opened 2021-11-09
Create run settings

Spartee:
Implementation of the ``Experiment.create_run_settings`` function which is the first step towards better portability of driver scripts between system types.

See issue #91 for more details.


smartsim/experiment.py
```diff
@@ -31,15 +31,13 @@
...
-from smartsim.control.manifest import Manifest
-
+from .settings import settings
 from .control import Controller, Manifest
 from .entity import Ensemble, Model
 from .error import SmartSimError
 from .generation import Generator
 from .utils import get_logger
-from .utils.helpers import colorize, init_default
+from .utils.helpers import colorize, init_default, is_valid_cmd
```
al-rigazzi, line 40: Is `is_valid_cmd` used? I don't see it in the code

al-rigazzi (approved):
I added just a small comment, otherwise looks good. We can unify LSF too, or do that in another PR, let me know what you think is better.

PR #112, opened 2021-11-21
Experiment.create_batch_settings

Spartee:
# Description

create_batch_settings is a portability improvement for users writing driver scripts.

With this function, a user no longer needs to specify a batch settings class, the function will initialize a given class based on the launcher

# TODO
- [x] Refactor the batch (``full_wlm``) tests


al-rigazzi (approved):
LGTM! We should create a follow-up ticket for LSF

PR #113, opened 2021-11-22
Add missing functions to LSFSettings API

al-rigazzi:
One function `set_hostlist` needs to `pass`, as this is only available (and already present) at batch level. The other two missing functions map the concept of node to that of RS, as it is the closest entity in LSF.

### DIFF
--- a/smartsim/settings/base.py
+++ b/smartsim/settings/base.py
@@ -201,10 +201,15 @@ class RunSettings:
 
 
 class BatchSettings:
-    def __init__(self, batch_cmd, batch_args=None):
+    def __init__(self, batch_cmd, batch_args=None, **kwargs):
         self._batch_cmd = batch_cmd
         self.batch_args = init_default({}, batch_args, dict)
         self._preamble = []
+        print("BatchSettings", kwargs)
+        self.set_nodes(kwargs.get("nodes", None))
+        self.set_walltime(kwargs.get("time", None))
+        self.set_queue(kwargs.get("queue", None))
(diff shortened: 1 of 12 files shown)


Spartee (approved):
Looks good to me. This should unify the interfaces nicely.

smartsim/settings/slurmSettings.py
```diff
@@ -224,13 +224,13 @@ def __init__(self, nodes=None, time="", account=None, batch_args=None, **kwargs)
...
         :type batch_args: dict[str, str], optional
         """
-        super().__init__("sbatch", batch_args=batch_args)
-        if nodes:
-            self.set_nodes(nodes)
-        if time:
-            self.set_walltime(time)
-        if account:
-            self.set_account(account)
+        print("SbatchSettings", kwargs)
```
Spartee, line 227: another one

smartsim/settings/base.py
```diff
@@ -201,10 +201,15 @@ def __str__(self):
 
 
 class BatchSettings:
-    def __init__(self, batch_cmd, batch_args=None):
+    def __init__(self, batch_cmd, batch_args=None, **kwargs):
         self._batch_cmd = batch_cmd
         self.batch_args = init_default({}, batch_args, dict)
         self._preamble = []
+        print("BatchSettings", kwargs)
```
Spartee, line 208: Assuming you didn't mean to leave these in here? :)

Spartee (commented):
Couple print statements left in the code

PR #116, opened 2021-12-03
Remove Pandas from experiment summary

MattToast:
- Refactor `Experiment.summary` method by removing usage of a pandas dataframe and instead utilizing tabulate to format the summary table
- Remove pandas as requirement for SmartSim
- Add tabulate as as requirement for SmartSim


Spartee (changes requested):
hey Matt! great stuff on the PR! 

A couple comments

theres a test in ``on_wlm/test_simple_entity_launch.py`` that tests the summary function and will need to be changed. In addition, we should add another test in the `test_experiment.py`` file to test another simple run.

MattToast:
Changes:
- Refactored `on_wlm/test_simple_entity_launch.py` to work with the summary `str` return type
- Added a test to `test_experiment.py` to test summary method without WLM integrations

smartsim/experiment.py
```diff
@@ -467,35 +467,35 @@ def summary(self):
         The summary will show each instance that has been
         launched and completed in this ``Experiment``
 
-        :return: pandas Dataframe of ``Experiment`` history
-        :rtype: pd.DataFrame
+        :return: tabulate string of ``Experiment`` history
```
Spartee, line 470: Lets allow ``format`` as a parameter here and pass to the tabulate function. Also should add in https://github.com/astanin/python-tabulate#table-format to the docstring so that people know what options are available.

smartsim/experiment.py
```diff
@@ -467,35 +467,35 @@ def summary(self):
...
+                        job.entity.name,
+                        job.entity.type,
+                        job.history.jids[run],
+                        run,
+                        job.history.job_times[run],
+                        job.history.statuses[run],
+                        job.history.returns[run],
+                    ]
+                )
+        return tabulate(values, headers, showindex=True, tablefmt="plain")
```
Spartee, line 498: lets set the default to ``github``. so the added parameter will be a keyword arg like ``format=github`` and then that input parameter will be passed here.

Spartee (changes requested):
Small changes! Otherwise looking great!

MattToast:
Changes:
- Added `format` as a parameter to `Experiment.summary` and passed value to call to `tabulate`
- Updated docstring with link to formats

Spartee (approved):
LGTM, lets chat before you merge!

PR #118, opened 2021-12-16
RayCluster with create_settings

al-rigazzi:
In this PR, we change the way Ray is started (internally): instead of creating `SrunSettings`, `AprunSettings`, and `MpirunSettings` objects, we just create a `BaseRunSettings` object with the correct `run_command`. The same holds for batch settings.

Also, notably, this automatically adds support for `Cobalt` as a launcher.


smartsim/experiment.py
```diff
@@ -207,11 +207,11 @@ def get_status(self, *args):
             statuses = []
             for entity in manifest.models:
                 statuses.append(self._control.get_entity_status(entity))
-            for entity_list in manifest.ensembles:
+            for entity_list in manifest.all_entity_lists:
                 statuses.extend(self._control.get_entity_list_status(entity_list))
-            orchestrator = manifest.db
-            if orchestrator:
-                statuses.extend(self._control.get_entity_list_status(orchestrator))
+            # orchestrator = manifest.db
```
Spartee, line 212: meant to be commented out?

smartsim/exp/ray/raycluster.py
```diff
@@ -129,12 +146,23 @@ def __init__(
             interface=interface,
             alloc=alloc,
             num_nodes=num_nodes,
+            run_command=run_command if run_command else "auto",
+            host_list=host_list,
             **kwargs,
         )
         if batch:
-            self._build_batch_settings(num_nodes, time, batch_args, launcher)
+            self.batch_settings = create_batch_settings(
```
Spartee, line 154: Assuming that the previous ``SSUnsupportedError`` catches the cases where ``create_run_settings`` may return an incompatible run command?

smartsim/exp/ray/raycluster.py
```diff
@@ -67,6 +75,10 @@ class RayCluster(EntityList):
     :type batch: bool
     :param time: The walltime the cluster will be running for
     :type time: str
+    :param run_command: specify launch binary. This is currently used only if launcher is Cobalt or PBS.
```
Spartee, line 78: Lets shorten these a bit

Spartee (changes requested):
A few comments. nothing major. Looks good.

smartsim/exp/ray/raycluster.py
al-rigazzi: I cannot think of a case in which this happens and it is not caught. I might be proven wrong, but all test cases passed.

smartsim/exp/ray/raycluster.py
al-rigazzi: I tried to reduce it to the minimum.

smartsim/launcher/lsf/lsfLauncher.py
```diff
@@ -26,6 +26,8 @@
 
 import time
 
+from smartsim.settings.settings import RunSettings
```
Spartee, line 29: use relative improt here

Spartee (approved):
One minor comment. otherwise LGTM!

PR #120, opened 2021-12-17
Automatic launcher detection

al-rigazzi:
This PR adds the automatic launcher detection to `Experiment` and to `create_run_settings` and `create_batch_settings`.

Two detect the launcher, we check that all commands specified in the corresponding `<launcher>Commands.py` file are available.
To distinguish between Cobalt and PBS, we call `qsub --version` and parse its output.


smartsim/experiment.py
```diff
@@ -65,7 +65,9 @@ def __init__(self, name, exp_path=None, launcher="local"):
         :param exp_path: path to location of ``Experiment`` directory if generated
         :type exp_path: str, optional
         :param launcher: type of launcher being used, options are "slurm", "pbs",
-                         "cobalt", "lsf", or "local". Defaults to "local"
+                         "cobalt", "lsf", or "local", if set to "auto",
```
Spartee, line 68: Period instead of comma

smartsim/experiment.py
```diff
@@ -38,7 +38,7 @@
 from .generation import Generator
 from .settings import settings
 from .utils import get_logger
-from .utils.helpers import colorize, init_default
+from .utils.helpers import colorize, detect_launcher, init_default
```
Spartee, line 41: Is a user ever going to want to use this function? If so, we may considering creating a new module thats more user facing as the helpers are going to be moved to the ``_core`` directory.

I think it's possible a user may want to inspect the value returned by ``detect_launcher`` for specific use cases. I could go either way on this one. If you don't think so, just leave it.

If you do think we should have a module, names?

``expUtils``?
``wlmUtils``?
``wlm``? 

I could see the last one being good if we end up moving the slurm (allocation) api to the ``wlm`` module.

Spartee (changes requested):
Few comments. Most important being the module placement. Lets chat about that whenever you have a chance.

Spartee (approved):
Please see one comment and make sure that test case will pass. otherwise LGTM.

tests/test_experiment.py
```diff
@@ -123,3 +123,8 @@ def test_summary(fileutils):
     assert m.type == row["Entity-Type"]
     assert 0 == int(row["RunID"])
     assert 0 == int(row["Returncode"])
+
+
+def test_launcher_detection(wlmutils):
+    exp = Experiment("test-launcher-detection", launcher="auto")
```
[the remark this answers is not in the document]
al-rigazzi: I moved this test to `on_wlm`. The only case we cannot test, this way, is if the detected launcher is `local`. I think it is quite a corner case (and the default value anyhow) and testing for it is not too relevant, but I'm open to suggestions.

PR #130, opened 2022-01-24
Add Build Matrix for CI

EricGustin:
This PR adds various dimensions to the CI build matrix for SmartSim. The build matrix now uses MacOS & Ubuntu, GNU8, RedisAI 1.2.3 & 1.2.5, and Python3.7-3.9. The build matrix excludes building with RedisAI 1.2.5 when on MacOS.

Effort and research was conducted on whether the CI can be ran with the Intel compiler. There is a follow up ticket on these findings. 

In addition to the CI build matrix, there were modifications to `_core/builder.py` by @Spartee that removed the usage of `shell=True` and replaced it with `shell=False` as the default behavior for running commands.


.github/workflows/build_wheels.yml
```diff
@@ -1,4 +1,4 @@
-name: run-tests-local
+name: build-wheels
```
Spartee, line 1: Seems like this was a mistake?

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+          sudo apt-get install -y wget
+
+      - name: Install GNU make for MacOS
+        if: "contains( matrix.os, 'macos' )"
+        run: brew install make || true
+
+      - name: Install SmartSim on MacOS
```
Spartee, line 49: we can probably collapse the macos and ubuntu stages here into a single run. This decreases the number of docker layers created by the CI and shoul dmake things faster.

.github/workflows/build_wheels.yml
Spartee: We should still be testing the wheel builds.

Spartee: Another thought here, 

we should add
```bash
export SMARTSIM_SUFFIX=develop
```
prior to the wheel build so the wheels are named with the dirty version.

Spartee (changes requested):

.github/workflows/build_wheels.yml
EricGustin: We are still testing the wheel builds. Wheel builds occur in `build_wheels.yml` and the extended build matrix which will eventually contain code coverage reporting is in `run_tests.yml`.

I added 
```yml
env:
    SMARTSIM_SUFFIX: "develop"
```
to the wheel builds in my most recent push.

Spartee (approved):
LGTM. Glad we discussed this.

PR #131, opened 2022-01-25
Refactor Orchestrator to launch in single command

al-rigazzi:
This PR adds a new `single_cmd` option to all Orchestrators. The single-cmd launch is based on the underlying MPMD mechanism (depending on the launcher).

Also, in this PR, the new `Experiment.create_orchestrator` function was introduced.


al-rigazzi:
Closing, as we will open another one with only one Orchestrator class.

PR #132, opened 2022-01-27
Remove numpy

EricGustin:
# Description
This PR removes the numpy dependency from SmartSim since SmartRedis already has numpy as a dependency. `step.py` needed to be altered such that it does not use numpy to generate a unique string. numpy was also removed from `requirements.txt`, `requirements-dev.txt`, and `setup.py`.


Spartee (approved):
LGTM

PR #133, opened 2022-01-27
Docker Documentation Builds

Spartee:
## Description

This PR adds the ability to build the documentation inside a docker build so that users don't have to worry about the specific build steps. The build exports the ``docs`` folder into the current clone so that the user doesn't have to do it themselves. build with ``make docks``. The manual documentation build is still available with ``make docs``.

Also added is a way for us to have develop documentation. Each PR that is merged, a new workflow will build the documentation in docker and push it to the doc branch which is hosted on github pages. this will only occur if the PR is reviewed, accepted, and merged.

There is also a new ``make tutorials``  that builds the tutorials into a developer docker with the current clone of SmartSim. This will eventually have a prod docker build too whch downloads the pinned version of smartsim for the latest release. 

## minor things
- fixed pip installed torch retrieval on ubuntu (@EricGustin this could be the tf error I was talking about)
- updated the doc requirements versions (sphinx)
- added docker composition levels
- start of minor documentation updates for 0.4.0


doc/conf.py
```diff
@@ -17,11 +17,17 @@
 # -- Project information -----------------------------------------------------
 
 project = 'SmartSim'
-copyright = '2021, Hewlett Packard Enterprise'
-author = 'HPE AI & Advanced Productivity'
+copyright = '2022, Hewlett Packard Enterprise'
```
al-rigazzi, line 20: I think we could write `2021-2022, Hewlett Packard Enterprise`, to indicate the year the process went public.

doc/index.rst
```diff
@@ -14,8 +14,6 @@
    :caption: Tutorials
 
    tutorials/01_getting_started/01_getting_started
```
al-rigazzi, line 16: I made the same comment on @MattToast's recent PR: I think we should move away from numbered tutorials, just keep a clean list. Otherwise mixing numbered and unnumbered ones is a bit confusing.

doc/requirements-doc.txt
```diff
@@ -1,6 +1,7 @@
 sphinx==3.1.1
 breathe==4.27.0
-sphinx_rtd_theme>=0.5.0
+#sphinx_rtd_theme>=0.5.0
```
al-rigazzi, line 3: Is there a reason to keep this?

docker/dev/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# - add openmpi
+# - add tutorials from other sections
+# - add permissions to the craylabs user
+
+RUN useradd --system --create-home --shell /bin/bash -g root -G sudo craylabs
+
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y build-essential \
+    git gcc make git-lfs wget libopenmpi-dev openmpi-bin \
+    python3-pip python3 python3-dev cmake \
```
al-rigazzi, line 17: Do we still need the pip-installed `cmake`? If that was for SmartRedis, then we can remove it and replace it with `apt-get`/`brew`.

This should have been done as part of the ticket in which we removed `pip install cmake` from SmartRedis. We can open another ticket, as it may go out of scope for this one.

docker/docs/dev/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+LABEL maintainer="Cray Labs"
+
+ARG DEBIAN_FRONTEND="noninteractive"
+ENV TZ=US/Seattle
+
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y build-essential \
+    git gcc make \
+    python3-pip python3 python3-dev cmake pandoc doxygen \
```
al-rigazzi, line 11: Same comment about `cmake` as above.

smartsim/_core/_install/buildenv.py
```diff
@@ -382,19 +382,43 @@ def is_windows(self):
...
+                torch_path = _path / "torch/share/cmake/Torch"
+                if torch_path.is_dir():
+                    return torch_path
+            return None
+
+        torch_path = _torch_import_path()
+        if not torch_path:
+            torch_path = _torch_site_path()
+        if not torch_path:
+            raise SetupError("Could not located torch cmake path")
```
al-rigazzi, line 421: "located"->"locate", and possibly capitalize Torch and CMake.

al-rigazzi (commented):
This PR looks very good and it is a great step towards a better process.

I have just found a couple of typos, noticed a couple of things that might be improved, and left a comment about CMake installation in containers. Maybe let's just decide on what we are going to do for the latter and then it can be merged.

docker/dev/Dockerfile
Spartee: this isn't pip installed, this is apt

doc/requirements-doc.txt
Spartee: nope, shouldv'e updated these too.

docker/dev/Dockerfile
al-rigazzi: Sorry, my bad!

doc/index.rst
Spartee: lets follow up on this one.

doc/conf.py
Spartee: done

al-rigazzi (approved):
OK, thanks for clarifying.

LGTM

PR #134, opened 2022-01-27
Refactor Orchestrator

al-rigazzi:
This PR updates the way the orchestrators are created and launched, especially on WLMs.

The changes are:
- Orchestrator class now is a generic class which can launch any orchestrator on supported launchers, with supported run commands. For example, instead of instantiating a `SlurmOrchestrator()`, users will instantiate `Orchestator(launcher='slurm')` or even `Orchestrator(launcher='auto')`.
- Orchestrator class has a new argument, `single_cmd` which is used to launch all shards with a single command, using the MPMD mechanism available for every run command (`srun`, `jsrun`, `aprun`, `mpirun`).
- All old Orchestrator classes now call the parent Orchestrator `__init__` function with the correct `launcher` arg.
- A new function `Experiment.create_database` can be used to create an Orchestrator, similar to what can be done to create an ensemble, a model, and so on.
- The `Orchestrator.set_hosts` function now only sets the hosts to each DBNode (or to each MPMD run settings) but not to the Orchestrator itself. The reason is that the host of each DBNode is what is used to launch the corresponding redis server, but the name or address needed by the Orchestrator is instead the one linked to the interface the redis server is bound to: if they differ, setting this with the name of the node can result in an error. We now always rely on parsing the output from the `redisstarter.py` script.
- Parsing `redisstarter.py` output for the MPMD case has now two possibilities: either look for _N_ IP addresses in one output (where _N_ is the number of shards) or look for _N_ output files, each one containing one IP address. Most run commands rely on the first mechanism, LSF relies on the second one.
- The size of the Orchestrator is not obtained as `len(Orchestrator)` anymore, as this would not work for MPMD instances. We assign the shard count to it.
- `database_per_host` is now removed, and only 1 db per node is supported.

This PR has to be considered as a draft, because of the following issues which are still open:
- Add a deprecation warning to each old class
- Decide where and how all the Orchestrator arguments should be documented, considering that some Orchestrators won't need all of them (e.g. LSF needs `gpus_per_shard` and `cpus_per_shard`).
- Decide whether tests for old orchestrators should be kept around as long as they are only deprecated.
- Update documentation and examples

Moreover, it makes sense to use this PR to also add the possibility of launching a converged Orchestrator. For that, some changes are needed, but mainly we need to change how each shards binds to a network address, and to pass a different env variable to the processes (e.g. not `SSDB`, or make sure that processes know when the orchestrator is converged). Also, if we want to add the possibility that ensembles are launched together with a converged orchestrator, pinning and/or wrapping might be needed in some cases.


al-rigazzi:
Closing as this has been superseded by the colocated Orchestrator PR.

PR #137, opened 2022-02-03
Add RAI checker for installed backends

al-rigazzi:
This PR mainly adds `CONFIG.get_installed_backends`, a property returning a list of installed Redis AI backends.

This is used e.g. in tests, to check whether the corresponding test should be run. This also helps distinguishing between tests we cannot run because there is no backend, and tests we cannot run because the library is not installed (e.g. `test_freeze_model` which does not need Redis AI).

Notice that for backend tests, we also check whether the module is (still) available, to prevent cases in which a user might have built RedisAI in another env.

### DIFF
--- a/smartsim/_core/config/config.py
+++ b/smartsim/_core/config/config.py
@@ -27,7 +27,6 @@
 import os
 from functools import lru_cache
 from pathlib import Path
-from typing import List
 
 import psutil
 
--- a/smartsim/_core/config/config.py
+++ b/smartsim/_core/config/config.py
@@ -171,20 +171,6 @@ class Config:
         # no account by default
         return os.environ.get("SMARTSIM_TEST_ACCOUNT", "")
 
-    @property
-    def installed_backends(self) -> List[str]:
-        installed = []
-        backends_path = self.lib_path / "backends"
-        for backend in ["tensorflow", "torch", "onnx"]:
-            backend_path = (
-                backends_path / f"redisai_{backend}" / f"redisai_{backend}.so"
-            )
-            backend_so = Path(os.environ.get("RAI_PATH", backend_path)).resolve()
-            if backend_so.is_file():
-                installed.append(backend)
-
-        return installed
-
 
 @lru_cache(maxsize=128, typed=False)
 def get_config():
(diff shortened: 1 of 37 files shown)


Spartee:
Thanks for catching this and taking care of it. My main thought on this is: if it's only used in tests... should it just be in ``conftest.py``? 

I suppose it might be nice to have this functionality internally. I can think of at least one use case which is to be able to discern wether a user has built any backends at all when initializing an ``Orchestrator``. Can you think of others I'm missing? 

If we do want it internally, it's not necessarily a configuration option as much as it is a helper function. ``smartsim/_core/util/helpers.py`` might be a better location. That is unless we want the user to be able to pass in the locations of the backends at some point... This is an option RedisAI supports with ``BACKENDSPATH``. I think if we do leave it in config, we should make sure that's the route we want to go and if so, make a note about it in the code.

I'm personally leaning towards a helper function. thoughts??

Spartee (commented):
See comment above

al-rigazzi:
@Spartee Yeah, I think we can move it to `smartsim/_core/util/helpers.py` and promote it to function (instead of it being a property of `CONFIG` as it is now).

tests/backends/test_tf.py
```diff
@@ -4,23 +4,25 @@
...
 
-should_run = True
+tf_available = True
 try:
     from tensorflow import keras
 
     from smartsim.tf import freeze_model
 except (ImportError, SmartSimError):
-    should_run = False
+    tf_availavble = False
```
MattToast, line 17: `tf_available`

smartsim/_core/config/config.py
```diff
@@ -27,6 +27,7 @@
 import os
 from functools import lru_cache
 from pathlib import Path
+from typing import List
```
MattToast, line 30: Remove unused typing module

MattToast (commented):
Quick fix on unused import and typo-ed variable name that i think is causing tests to error. Otherwise lgtm!!

tests/backends/test_tf.py
al-rigazzi: Thanks, fixed

smartsim/_core/config/config.py
al-rigazzi: Removed

Spartee (approved):
LGTM. Eventually we want to add this to the CLI. maybe a `smart build --check-installed`` command? Either way, glad to have this functionality.

PR #138, opened 2022-02-03
Fix is_active() call to check_cluster_status()

mellis13:
Orchestrator method is_active() had the incorrect set of parameters for check_cluster_status() call.  This PR fixes that and adds a check in WLM tests to check that get_address() (which calls this method) returns correctly sized result.


al-rigazzi (approved):
Thanks for this PR, really a good catch!

LGTM

PR #140, opened 2022-02-08
DataLoader fix

al-rigazzi:
This PR adds fixes for some issues introduced by the DataLoaders, namely:
- DataLoaders were incorrectly linked in docs
- DataLoader notebooks included kernels and that made `make docks` fail
- `test_dataloaders.py` used the standard Redis port for the orchestrators, but that port is already in use on some systems (switched to 6780)


Spartee (approved):
LGTM

PR #143, opened 2022-02-10
Fix build_docs github action + some minor improvements

ben-albrecht:
Fix `build_docs` github action that was previously broken. Some other minor improvements were added as part of this fix.

Changes include:

- `build_docs` action is triggered when we push to develop branch
    - Previously, it was on PR closed, which isn't necessarily a merge, i.e. a contributor could close a PR without merging and the action would still trigger
- Use `GITHUB_TOKEN` instead of personal authentication (PA) credentials to push
- Use "action@github.com" as the git author (user) to distinguish automated commits from human commits
- Check static files from doc branch into develop, so that we can modify
  them in a regular PR to `develop`, rather than have to edit the doc branch
  directly. I'd like for us to get to a point where doc branch pushes are fully automated, so we should never have to modify it directly.
    - Static files include: `.nojekyll` (both top-level and per-sphinx-build), `CNAME`, and top-level `index.html`

### DIFF
--- a/.docs_static/CNAME
+++ b/.docs_static/CNAME
@@ -1,0 +1,1 @@
+www.craylabs.org
--- a/.docs_static/index.html
+++ b/.docs_static/index.html
@@ -1,0 +1,2 @@
+
+<meta http-equiv="refresh" content="0; url=./docs/overview.html" />
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -1,8 +1,9 @@
 name: deploy_dev_docs
 
 on:
-  pull_request:
-    types: [closed]
+  push:
+    branches:
+      - develop
 
 jobs:
   build_docs:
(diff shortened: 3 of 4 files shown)


ben-albrecht:
Example build log here:

https://github.com/ben-albrecht/SmartSim/actions/runs/1824864337

ben-albrecht:
It looks like there is a separate incompatibility issue in our `requirements-docs.txt`, but I am not sure if that actually impacts anything other than showing an error during env build:

https://github.com/ben-albrecht/SmartSim/runs/5143975793?check_suite_focus=true#step:5:961

Spartee (approved):
LGTM

ben-albrecht:
> @Spartee - I noticed this is a different theme than read-the-docs. Is that an intentional change in the recent make docks changes in #133?

For the archives, @Spartee mentioned offline that this was an intentional change.

PR #145, opened 2022-02-11
Update warnings

al-rigazzi:
This PR updates deprecation warnings for `smartsim.constants` and tests thereof. It also introduces an exception to Orchestrator, preventing users from running an aprun-launched orchestrator with MPMD in batch, as this fails.


al-rigazzi:
Close because superseded by other PR

PR #146, opened 2022-02-11
Fixes to mpirun, mpmd, and aprun

al-rigazzi:
This PR adds fixes to issues encountered while testing on specific systems. Specifically:
- mpirun mpmd behavior which concatenated more commands than needed was fixed
- prevent aprun batch launch of orchestrator with single_cmd
- terminate processes with kill to prevent issues with mpirun on slurm-based systems
- update deprecations and tests thereof
- fix names of some generic tests


Spartee (approved):
LGTM

PR #151, opened 2022-02-16
Move docker pip install to requirements-doc.txt

ben-albrecht:
Moved the additional python package dependencies from the docker file to the documentation requirements file.

This resolves a `make docks` bug that showed up on MacOS with python 3.8+ where the dependency resolution resulted in an error (specifically, when trying to resolve version for `appnope`).

Also removed `ipykernel` and `ipython` from requirements, since they are included as dependencies to `nbsphinx`.

### DIFF
--- a/doc/requirements-doc.txt
+++ b/doc/requirements-doc.txt
@@ -3,5 +3,6 @@ breathe==4.31.0
 sphinx-book-theme==0.2.0
 sphinx-fortran==1.1.1
 nbsphinx>=0.8.2
+docutils==0.17
 torch==1.7.1
 tensorflow==2.5.2
--- a/doc/requirements-doc.txt
+++ b/doc/requirements-doc.txt
@@ -3,3 +3,5 @@ breathe==4.31.0
 sphinx-book-theme==0.2.0
 sphinx-fortran==1.1.1
 nbsphinx>=0.8.2
+torch==1.7.1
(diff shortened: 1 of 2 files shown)


Spartee (commented):
So do the ML and nbsphinx dependencies get downloaded? 

I see that we moved them to requirements-doc.txt but that ``pip install -r doc/requirements-doc.txt`` is never called.

Given that it's impossible to build the docs from a pip installed wheel (only possible in a git checkout) I vote that we remove the ``[doc]`` extras_require in ``setup.py`` and maintain only this list in the ``requirements-doc.txt`` inside of the ``doc`` folder.

We will also need to change the dockerfile to use the ``requirements-doc.txt`` file and remove the ``[doc]`` portion of the ``pip install`` for SmartSim.

Spartee:
I think we should also add the following 

```
docutils==0.17
```

as I've seen this in the actions log

```
ERROR: sphinx 4.4.0 has requirement docutils<0.18,>=0.14, but you'll have docutils 0.18.1 which is incompatible.
ERROR: sphinx-book-theme 0.2.0 has requirement docutils<0.17,>=0.15, but you'll have docutils 0.18.1 which is incompatible.
```

And then I think this is ready to go

Spartee (approved):
LGTM

PR #152, opened 2022-02-17
Update CMake version from 3.10 to 3.13

mellis13:
The SmartSim documentation lists CMake version 3.10 as a requirement for SmartSim and SmartRedis.  SmartRedis was upgraded to a minimum CMake version of 3.13.  This PR updates the documentation.  Also, to maintain a consistent build and test environment, the CMake minimum version has also been upgrade in SmartSim.   The current version of CMake is 3.23, so users are unlikely to notice a change unless they are using a very old version of CMake, and newer versions of CMake are easily installed in a conda environment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
 
       - name: Build sdist
         run: |
-          python -m pip install cmake>=3.10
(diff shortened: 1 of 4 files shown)


Spartee (approved):
LGTM

Issue #153, opened 2022-02-17
smart build --device CPU attempts a GPU install


billschereriii:
# Description
The `smart` install is performing case sensitive comparisons when selecting the type of install to perform.  With command line `smart build --device CPU`, it attempted a GPU install.

# How to reproduce
Execute the command line `smart build --device CPU`. We observed the failure on theta, but expect that it is platform independent.

# Expected behavior
With this command line, a CPU installation, not a GPU installation should be attempted. In general, unless a target of `gpu` or `GPU` is given, the default should be for a  CPU installation.

# System
 - OS: Linux (theta)
 - OS version: unknown
 - Commit/version of library: 0.4.0
 - Workload manager (e.g. PBS, Slurm): Cobalt
 - Workload manager version: 1.0.32

PR #154, opened 2022-02-18
Add option to use system-available libtensorflow for RedisAI

al-rigazzi:
RedisAI TF backend requires libtensorflow. The standard RedisAI build script calls `get_deps.sh`, which downloads the appropriate libtensorflow tarball from Google's repository. 

In this PR, we address the situation in which the tarball is not available (e.g. if it is missing for a given architecture), but the user has libtensorflow installed on the system. Notice that until RAI 1.2.3, it was possible to use a `FindTensorflow.cmake` module to use the system-installed tensorflow.

We add the option `--libtensorflow_dir` to `smart build`, and symlink the user-provided libtensorflow dirs into RAI's dependency tree.

Another build issue, raised in #153, i.e. the handling of uppercase device specification (such as `--device CPU`) was fixed in this PR.

### DIFF
--- a/smartsim/_core/_install/builder.py
+++ b/smartsim/_core/_install/builder.py
@@ -1,4 +1,6 @@
+import glob
 import os
+import platform
 import re
 import shutil
 import stat
@@ -197,6 +199,7 @@ class RedisAIBuilder(Builder):
         self,
         build_env={},
         torch_dir="",
+        libtf_dir="",
         build_torch=True,
         build_tf=True,
         build_onnx=False,
@@ -210,6 +213,7 @@ class RedisAIBuilder(Builder):
         self.torch = 1 if build_torch else 0
         self.tf = 1 if build_tf else 0
         self.onnx = 1 if build_onnx else 0
+        self.libtf_dir = libtf_dir
         self.torch_dir = torch_dir
 
     @property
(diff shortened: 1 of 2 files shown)


smartsim/_core/_install/builder.py
```diff
@@ -233,6 +237,43 @@ def copy_tf_cmake(self):
...
+    def symlink_libtf(self, device):
+        """Add symbolic link to available libtensorflow in RedisAI deps.
+
+        :param device: cpu or gpu
+        :type device: str
+        """
+        rai_libtf_path = (
+            self.rai_build_path
+            / "deps"
+            / f"{platform.system().lower()}-{platform.machine()}-{device}"
```
Spartee, line 249: fuzzy match this instead?

smartsim/_core/_install/builder.py
```diff
@@ -233,6 +237,43 @@ def copy_tf_cmake(self):
...
+        if rai_libtf_path.is_dir():
+            shutil.rmtree(rai_libtf_path)
+
+        os.makedirs(rai_libtf_path)
+        libtf_path = Path(self.libtf_dir).resolve()
+        os.symlink(libtf_path / "include", rai_libtf_path / "include")
+        # RedisAI expects to find a lib directory, which is only
+        # available in some distributions.
+        rai_libtf_lib_dir = rai_libtf_path / "lib"
+        os.makedirs(rai_libtf_lib_dir)
```
Spartee, line 262: May want to include some execption handling here for errors that crop up in directory naming and what not.

Spartee (commented):
This is great, I want to touch base on this though as I cannot grasp whats going on in the ``symlink_tf`` function. Questions to follow on slack.

Spartee (approved):
LGTM

PR #156, opened 2022-02-22
Update SmartSim test suite after v0.4.0 release

al-rigazzi:
This PR addresses the SmartSim test suite.

Major features are:
- make `mpirun` available as run command for tests with Cobalt and PBS (fetch `COBALT_NODEFILE` or `PBS_NODEFILE` to get the hosts)
- automatic directory naming for test outputs
- test cleanup, needed after some changes on test systems


conftest.py
```diff
@@ -90,6 +91,31 @@ def kill_all_test_spawned_processes():
     except Exception:
         print("Not all processes were killed after test")
 
+def get_hostlist():
+    global test_hostlist
+    if not test_hostlist:
+        if "COBALT_NODEFILE" in os.environ:
+            try:
+                with open(os.environ["COBALT_NODEFILE"], 'r') as nodefile:
```
Spartee, line 99: Sometimes these node files contain duplicate node names if the allocation was obtained with more than 1 rank per node. we should be sure that there aren't duplicates here as that would mess up the database deployment

conftest.py
```diff
@@ -90,6 +91,31 @@ def kill_all_test_spawned_processes():
...
+            except:
+                return None
+        elif "COBALT_NODEFILE" in os.environ:
+            try:
+                with open(os.environ["PBS_NODEFILE"], 'r') as nodefile:
+                    lines = nodefile.readlines()
+                    test_hostlist = [line.strip() for line in lines]
+            except:
+                return None
+        elif "SLURM_JOB_NODELIST" in os.environ:
```
Spartee, line 111: Should we make generalized functions for these? I imagine this functionality might be a good candidate for ``smartsim.wlm``. Make a ticket? lets chat about this one.

Spartee (approved):
Looks fantastic, I had one comment, but approved otherwise. Thanks for doing this.

conftest.py
al-rigazzi: OK, I removed duplicates using the `list(dictfromkeys(...))` trick.

conftest.py
al-rigazzi: Yeah, totally makes sense, since some of our power users implemented their own versions of it.

PR #157, opened 2022-02-23
Add ipython dep to doc build

Spartee:
One liner, forgot to add ipython back into the build. it's needed separately from nbsphinx as detail in this issue

https://github.com/spatialaudio/nbsphinx/issues/24


al-rigazzi (approved):
LGTM, unless you think we should pin to a specific version.

PR #164, opened 2022-03-02
Fix wrong handling of LSF gpus_per_shard option

al-rigazzi:
This is a hotfix PR, a mishandled option is preventing LSF orchestrators from using more than one GPU per shard.


EricGustin (approved):
LGTM!

PR #165, opened 2022-03-02
Fix re-run smart build bug

### LINKED ISSUE #162: ``smart build`` doesn't inform user which backends are built
# Description
When running ``smart build``, if only some backends are built, say ``torch``, but others are requested, say ``tf`, the command just completes without informing the users that in fact tensorflow is not actually built.

# How to reproduce
 - run ``smart build --no_tf --device cpu``
 - run ``smart build --device cpu`` (with tf build)

``smart build`` will finish but tensorflow backend will not be built.

# Expected behavior
The user should be informed that they need to run ``smart clean`` and then re-run with the args they want to re build the backends.

Ideally we could also just show which backends are built from the ones that are requested. something like

```text
[SmartSim] INFO Running SmartSim build process...
[SmartSim] INFO Checking for build tools...
[SmartSim] INFO Checking requested versions...
[SmartSim] INFO Redis build complete!

ML Backends Requested
-----------------------
    PyTorch 1.7.1: True
    TensorFlow 2.5.2: True
    ONNX 1.9.0: False

Building for GPU support: False

[SmartSim] INFO TensorFlow 2.5.2 installed in Python environment
[SmartSim] INFO Torch 1.7.1 installed in Python environment
[SmartSim] INFO RedisAI installed. Run `smart clean` to remove.
[SmartSim] INFO Torch, TensorFlow backends built
[SmartSim] INFO SmartSim build complete!
```

# System
 - OS: any
 - OS version: any
 - Commit/version of library: 0.4.0

EricGustin:
This PR fixes #162. 
There are four cases that this PR accounts for which have been confirmed to work on local machine and Horizon.
1. User attempts to run `smart build` when `RAI_PATH` is set and backends are installed
   - Inform user that backends are already built, tell them where they are built, and notify the user that there is no reason to build.
2. User attempts to run `smart build` when `RAI_PATH` is set, but backends are not installed
   - Inform user that before running `smart build`, they should unset `RAI_PATH`
3. User attempts to run `smart build` when `RAI_PATH` is not set and backends are installed
   - Inform user that they need to run `smart clean` before running `smart build`
4. User attempts to run `smart build` when `RAI_PATH` is not set and backends are not installed
   - Allow `smart build` to continue

### DIFF
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -320,7 +320,7 @@ class Build:
         if os.environ.get("RAI_PATH"):
             if installed_redisai_backends():
                 logger.error(
-                    f"Backends are already built and loaded at {CONFIG.redisai}. There is no need to build."
+                    f"There is no need to build. Backends are already built and specified in the environment at 'RAI_PATH': {CONFIG.redisai}"
                 )
             else:
                 logger.error(
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -311,7 +311,6 @@ class Build:
         except SetupError as e:
             logger.warning(str(e))
 
-
     def check_backends_install(self):
         """Checks if backends have already been installed.
         Logs details on how to proceed forward
@@ -320,9 +319,13 @@ class Build:
         """
         if os.environ.get("RAI_PATH"):
             if installed_redisai_backends():
-                logger.error(f"Backends are already built and loaded at {CONFIG.redisai}. There is no need to build.")
+                logger.error(
+                    f"Backends are already built and loaded at {CONFIG.redisai}. There is no need to build."
+                )
             else:
-                logger.error(f"Before running 'smart build', unset your RAI_PATH environment variable with 'unset RAI_PATH'.")
+                logger.error(
+                    f"Before running 'smart build', unset your RAI_PATH environment variable with 'unset RAI_PATH'."
+                )
             exit(1)
         else:
             if installed_redisai_backends():
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -154,11 +154,7 @@ class Build:
         print(f"    ONNX {self.versions.ONNX}: {color_bool(onnx)}\n")
         print(f"Building for GPU support: {color_bool(device == 'gpu')}\n")
 
-        if os.path.isdir(CONFIG.lib_path) and os.listdir(CONFIG.lib_path):
-            logger.error(
-                "If you wish to re-run `smart build`, you must first run `smart clean`."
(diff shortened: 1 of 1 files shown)


Spartee (changes requested):
Overall this is great. There is one problem which I did not forsee writing up the ticket. The function ``installed_redisai_backends`` accounts for the fact that the backends may be installed site-wide at the ``RAI_PATH`` which is our plan for smartsim module files (see https://github.com/craylabs/poseidon/issues/176).

This function will stop users from trying to build when they already have the backends built somewhere, but the error message isn't super helpful. They might even try to run ``smart clean`` and then find out they are still greeted with the same message if they still have the modulefile loaded.

Leaning on @al-rigazzi and @ashao to make suggestions here. Requesting changes as I think we need to address this first as site-wide installs are part of the plan.

al-rigazzi:
Yeah, I think the user should be notified if `smart clean` will not be effective. One possibility would be to explicitly tell the user where the backends are installed, and what actions could be useful (i.e. build your own SmartSim or change `RAI_PATH`, maybe?)

Spartee:
@al-rigazzi I agree. I think the best plan is to allow them to build and use the backends they want to build seperately if need be. But the default should be to use the set `RAI_PATH` and `REDIS_PATH` etc, and prompt the user that the backends are already built at a specific location. 

It would be best if there was some way for the user to also install the recommended ML backends provided the RAI version that was installed. this would enabled site-install users to run

```bash
# create python env
module load smartsim-0.4.0-rai-1.2.3
smart build --install-ml
```

this would essentially skip the build process if the backends were module loaded into the environment and then install the correct python dependencies into the user environment.

what do you think @al-rigazzi ?

al-rigazzi:
Yes, I agree with @Spartee, that would be cleaner (and automatically documented through `smart --help`).

EricGustin:
@Spartee @al-rigazzi changes have been made to address your comments.

smartsim/_core/_cli/build.py
```diff
@@ -298,3 +310,26 @@ def check_tf_install(self):
...
+    def check_backends_install(self):
+        """Checks if backends have already been installed.
+        Logs details on how to proceed forward
+        if the RAI_PATH environment variable is set or if
+        backends have already been installed.
+        """
+        if os.environ.get("RAI_PATH"):
+            if installed_redisai_backends():
+                logger.error(
+                    f"Backends are already built and loaded at '{CONFIG.redisai}'. There is no need to build."
```
Spartee, line 323: mention that they are specified in the environment at ``RAI_PATH`` so the user is aware thats where they are specified.

Spartee (approved):
One small change, otherwise good to merge! Feel free to merge after that comment is edited.

PR #168, opened 2022-03-08
Fix run_args issue in create_database

al-rigazzi:
This PR fixes an incorrect usage of `kwargs` in `Orchestrator._build_run_settings`.
Since `run_args` is not popped from `kwargs`, when we use it as argument in `create_run_settings`, two copies are passed, causing an error.

We also eliminated an unused Slurm option setting, which was redundant.


Spartee (approved):
LGTM

PR #169, opened 2022-03-10
Select GPUs via Orchestrator [preview -- do not merge]

billschereriii:
Add orchestrator method to select GPUs, calling through to SmartRedis client

### DIFF
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -520,6 +520,31 @@ class Orchestrator(EntityList):
                 "The SmartSim Orchestrator must be active in order to set the database's configurations."
             )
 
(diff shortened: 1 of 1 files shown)


Spartee:
Given the proposed approach in Pos#131 I think this can be closed. I would keep the branch around just in case we change our minds though.

Spartee:
So Pos#131 will close this as we have decided to use the cli to do model setting within SmartSim so as to not rely on SmartRedis develop for testing purposes. The CLI provides the same funtionality and is also available internally to smartsim through the ``CONFIG`` object. There are also some other reasons around containers, but either way this PR can be closed as this isn't the approach we will be going with.

Issue #161, opened 2022-02-26
Address gaps in RunSettings implementations


Spartee:
# Description
the advent of ``Experiment.create_run_settings`` made it easier for users to write driver scripts that are portable across workload managers and system types. However, not all the interfaces for the RunSettings are the same. 

For example, ``JsrunSettings.set_tasks_per_rs`` only applies to Jsrun. This doesn't mean that we need to implement this in slurm, but we need to grow the list of interface functions that are available across all ``RunSettings`` objects where possible.

The current interface thats common across all implementations is reflected in the base ``RunSettings`` class.

```python
    def set_tasks(self, tasks):
        """Set the number of tasks to launch

        :param tasks: number of tasks to launch
        :type tasks: int
        """
        raise NotImplementedError(
            f"Task specification not implemented for this RunSettings type: {type(self)}"
        )

    def set_tasks_per_node(self, tasks_per_node):
        """Set the number of tasks per node

        :param tasks_per_node: number of tasks to launch per node
        :type tasks_per_node: int
        """
        raise NotImplementedError(
            f"Task per node specification not implemented for this RunSettings type: {type(self)}"
        )

    def set_cpus_per_task(self, cpus_per_task):
        """Set the number of cpus per task

        :param cpus_per_task: number of cpus per task
        :type cpus_per_task: int
        """
        raise NotImplementedError(
            f"CPU per node specification not implemented for this RunSettings type: {type(self)}"
        )

    def set_hostlist(self, host_list):
        """Specify the hostlist for this job

        :param host_list: hosts to launch on
        :type host_list: str | list[str]
        """
        raise NotImplementedError(
            f"Host list specification not implemented for this RunSettings type: {type(self)}"
        )

```

# Justification
Users who want to write portable driver scripts.

# Implementation Strategy

- [ ] research and identify arguments that are common across at least 2 settings classes.
      - Some examples
        - cpu bindings
        - memory
        - verbose launch
        - broadcast
- [ ] Solidify interface functions and present to team.
- [ ] Implement the missing functions in the Settings classes that are relevant.
- [ ] Implement local tests for this run settings additions.
- [ ] For any settings that those arguments do not apply to, throw a warning and pass (do not throw error)
- [ ] remove ``Raise NotImplementedError`` from ``RunSettings`` and others and replace with warnings for it being not applicable.
- [ ] ensure warnings are thrown through tests. Ask @al-rigazzi for how he did this.

PR #170, opened 2022-03-14
Add new RunSettings Methods

MattToast:
Adds new `RunSettings` methods to grow the list of interface functions that are available across all `RunSettings` objects where possible. Methods added for similar arguments that are shared by two or more of the existing `RunSettings` sub-classes.

Do let me know if I am missing any additional methods for arguments that should be included, or implementations for existing `RunSettings`.


smartsim/settings/alpsSettings.py
```diff
@@ -129,6 +129,66 @@ def set_excluded_hosts(self, host_list):
...
+        :type memory_per_node: int
+        """
+        self.run_args["memory-per-pe"] = int(memory_per_node)
+
+    def set_verbose_launch(self, verbose):
+        """Set the job to run in verbose mode
+
+        This sets ``--debug`` arg to the highest level
+
+        :param verbose: Whether the job should be run verbosly
```
al-rigazzi, line 159: "verbosly" -> "verbosely"? (here and other times through the code)

al-rigazzi (commented):
Looks good. I'd just like to understand better what was the criterion to decide which option should get a method and which not. Maybe we chat about it?
Also, I really like the warning approach for missing functions, it will help unifying driver scripts!

smartsim/settings/base.py
```diff
@@ -117,10 +117,70 @@ def set_hostlist(self, host_list):
...
+        :type memory_per_node: int
+        """
+        logger.warning(
+            f"Memory per node specification not implemented for this RunSettings type: {type(self)}"
+        )
+
+    def set_verbose_launch(self, verbose):
+        """Set the job to run in verbose mode
+
+        :param verbose: Whether the job should be run verbosly
```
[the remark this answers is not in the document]
MattToast: Fixed in https://github.com/CrayLabs/SmartSim/pull/170/commits/82114ff67a5118c6ce40455fdff5519a7dcf66b6

smartsim/settings/alpsSettings.py
MattToast: Fixed here and throughout code base in https://github.com/CrayLabs/SmartSim/pull/170/commits/82114ff67a5118c6ce40455fdff5519a7dcf66b6

MattToast:
@al-rigazzi Thanks for the review!

Issue #161 goes into more detail, but basically the deciding factor for whether or not an option got made into method on the base class was whether or not is the the same or very similar functionality existed on at least two of the supported interfaces (`slurm`, `jsrun`, etc).

I did my best to try and parse docs for each of the interfaces and I think I got most of the common options (at least all the options I understood). If you can think of any other arguments I missed, do lmk and I can add them in asap!!

smartsim/settings/lsfSettings.py
```diff
@@ -166,7 +157,27 @@ def set_cpus_per_task(self, cpus_per_task):
         :param cpus_per_task: number of cpus per resource set
         :type cpus_per_task: int
         """
-        self.set_cpus_per_rs(cpus_per_task)
+        self.set_cpus_per_rs(int(cpus_per_task))
+
+    def set_memory_per_rs(self, memory_per_rs):
```
Spartee, line 162: @al-rigazzi do you think we should continue to have these aliases? or just implement the specification?

smartsim/settings/mpirunSettings.py
```diff
@@ -134,6 +137,61 @@ def set_hostlist(self, host_list):
...
+
+        This sets ``--verbose``
+
+        :param verbose: Whether the job should be run verbosely
+        :type verbose: bool
+        """
+        if verbose:
+            self.run_args["verbose"] = None
+            return
+        if "verbose" in self.run_args:
```
Spartee, line 151: this is a bit hard to understand. I get that you like the truthy values, but if you keep it like this, lets make sure there are some comments explaining why.

personally I feel like this would be more readable

```python
if verbose:
    ...
else:
    ...
```

smartsim/settings/mpirunSettings.py
```diff
@@ -134,6 +137,61 @@ def set_hostlist(self, host_list):
...
+
+    def set_timeout(self, time):
+        """Set the maximum number of seconds that a job will run
+
+        This sets ``--timeout``
+
+        :param time: The maximum number of seconds that a job will run in secs
+        :type quiet: int
+        """
+        self.run_args["timeout"] = int(time)
```
Spartee, line 193: on these methods that supply a time constraint, I feel like we can improve upon "number of seconds". 

for example, what about ``hours=1, minutes=0, seconds=0`` as the default and we just convert to seconds. I feel like we could do this in other places as well.

smartsim/settings/mpirunSettings.py
Spartee: given this seems to be in a number of places, please either provide a reasoning why this is better then a traditional conditional or change them to a simple if/else.

smartsim/settings/slurmSettings.py
```diff
@@ -150,6 +150,86 @@ def set_tasks_per_node(self, tasks_per_node):
...
+        """Set the maximum number of seconds that a job will run
+
+        This sets ``--time``
+
+        :param time: The maximum number of seconds that a job will run in secs
+        :type quiet: int
+        """
+        time = int(time)
+
+        if time == 0:
```
Spartee, line 225: see comment about time

Spartee (changes requested):
Good stuff! A couple things

1. there are still gaps between the implementations. ex. ``SrunSettings`` has ``set_nodes`` and few others do each class should have the *exact* same methods available that will raise warnings in the case where it's not implemented.  there should not be an ``AttributeError`` raised if I call the same methods across all classes. this is with the exception of ``JsrunSettings`` which needs weird special methods for resource sets. more on this later

2. I think we should provide a better interface for time across all the settings. I think we should be doing the translation to a string walltime. All the batch settings implement ``set_walltime`` and I think we should also provide ``set_time`` which takes in the suggestion i commented above as arguments and then formats and calls ``set_walltime`` if its a ``BatchSettings`` class.

3. @al-rigazzi why does ``AprunSettings`` randomly implement ``set_walltime``? was there a need for that? you're on the git blame.

4. Instead of ``set_timeout`` lets call it ``set_time`` and implement the suggestion above. One helper method in the base class to do the time conversion and those classes that need to can override. 

5. ``set_hostlist_from_file`` is another I think that could be implemented across all of them.

6. @al-rigazzi Is there a strict reason why we need to continue supporting the resource set semantics of LSF? can we get away with just node/task/etc if we mention that this is an alias for resource sets? I would like to get consistent.

Spartee:
also, whitespace!!!! purge it! ha

smartsim/settings/mpirunSettings.py
MattToast: No good reason! Wasn't sure of if there where coding standards against a `if/elif` chains without a default `else` and I wanted to avoid over-indenting with a
```py
if verbose:
    ...
else:
    if "verbose" in dict:  # avoid KeyExp
        ...
```
so i opted for a guard clause instead.

Whole thing can be avoided by simply using `dict.pop`, but I am bad at my job and forgot that method exists!! Fixed in https://github.com/CrayLabs/SmartSim/pull/170/commits/fb759c0a1100c233e79014e682a3e4ca7e5b82fd

smartsim/settings/mpirunSettings.py
MattToast: Like the idea!! My only concern would be defaulting the `hours` param to be 1. I think it seems a bit un-intuitive that that:
```py
rs: RunSettings = ...
rs.set_time(seconds=5)
```
would set the time limit to 365 seconds from now. We could do scan the parameters and manually correct to the time to be 5 seconds seconds in the future, but I think the default being 0 / unlimited time makes more sense.

MattToast: Basic implementation of `set_time` in [b9aec16](https://github.com/CrayLabs/SmartSim/pull/170/commits/b9aec16118b55f342a31d4b3e77b8e60743ab137)

smartsim/settings/lsfSettings.py
al-rigazzi: @Spartee (cc @MattToast): Answering this and your general comment: I am not really sure we can map everything in LSF to node/task/etc. We can try, and simply say that for everything else, a standard `run_arg` method can be used. 

About `set_walltime` in `AprunSettings`. I cannot think of a reason why we should have it now, I guess it made sense in an old version of the library, possibly.

smartsim/settings/base.py
```diff
@@ -272,6 +486,19 @@ def format_run_args(self):
             formatted.append(str(value))
         return formatted
 
+    def format_env_vars(self):
```
Spartee, line 489: I think we should just return a string that would work with the ``env`` command here. this is overridden in every class anyway.

Spartee (commented):
One small comment

smartsim/settings/base.py
MattToast: Implemented in [e887de9](https://github.com/CrayLabs/SmartSim/pull/170/commits/e887de9b7bfd74f21575428bb3ff2c213ef5f6f8)

Just want to confirm that is okay that function signature are changed by inherited classes (for instance `MpirunSettings.format_env_vars` returns a `list[str]`).

Spartee (approved):
Approving in advance based on conversation about unifying the return type of the ``format_env_vars`` command

smartsim/settings/base.py
MattToast: signatures unified in [2e20b85](https://github.com/CrayLabs/SmartSim/pull/170/commits/2e20b858b143dc048a2806a25b676c0892159c9b)

going to give this a final pass on monday before merging

PR #172, opened 2022-03-16
Fix codecov badge 'unknown' bug

EricGustin:
This is a fix for the codecov badge in the README displaying as 'unknown' rather than displaying the repository's code coverage percentage. This occurs because the coverage report is never uploaded on the develop branch. Instead, the coverage report is ran on the compare branch of a PR. This is a problem because the codecov badge is linked to the develop branch. Since no coverage report is currently being uploaded from the develop branch, the badge is not able to display the repository's code coverage percentage.

The solution introduced in this PR is to run the coverage tests & upload to codecov on pushes to the develop branch, in addition to the already existing event trigger for all PRs.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -1,6 +1,10 @@
 name: run-tests
 
-on: pull_request
+on:
+  pull_request:
+  push:
+    branches:
+      - develop
 
 env:
   HOMEBREW_NO_ANALYTICS: "ON" # Make Homebrew installation a little quicker


ben-albrecht:
Can you clarify why we want it to run on both `pull_request` and `push`? I would think we only want to update the badge when a PR is merged, which would just be `push`.

EricGustin:
@ben-albrecht yeah this is a good point. The purpose of running on `push` is for the sake of the badge whereas the purpose of running on `pull_request` is to ensure all PRs are not introducing any bugs i.e. ensure all tests pass. I guess maybe the discussion we should have is whether running the CI on `push` to develop is worth it if the only purpose is to get the badge working.

ben-albrecht:
@EricGustin - would it make sense to split this into 2 separate jobs, where one is checking coverage before merge with trigger `on: pull_request`, and one is updating the badge post-merge with trigger `on: push: branches: [develop, master]` ?

ben-albrecht (approved):
LGTM

PR #174, opened 2022-03-17
Set test hostlist to None when aprun is available

al-rigazzi:
On certain Cray systems, PBS_NODEFILE does not contain the list of allocated nodes (it actually contains the name MOM node from which it was launched). This breaks some tests where the hostlist is used to launch the orchestrator on PBS. This PR prevents PBS_NODEFILE to be parsed when `aprun` is available (i.e. if the system is a Cray).


al-rigazzi:
Actually closing this as it might be too system-specific.

PR #176, opened 2022-03-20
New ML online training tutorial

al-rigazzi:
This PR substitutes the old ML training tutorials with a new one.

The major features of this tutorial:
- is set to run with local launcher, thus it will be run in the docs page too
- trains a real surrogate model
- does not require Horovod or MPI4PY to run
- runs in 3 minutes on a standard CPU

The whole training is performed in Keras, the PyTorch version will be ready soon.


smartsim/ml/tf/utils.py
```diff
@@ -50,3 +50,33 @@ def freeze_model(model, output_dir, file_name):
...
+        tf.TensorSpec(model.inputs[0].shape, model.inputs[0].dtype)
+    )
+
+    frozen_func = convert_variables_to_constants_v2(full_model)
+    frozen_func.graph.as_graph_def()
+
+    input_names = [x.name.split(":")[0] for x in frozen_func.inputs]
+    output_names = [x.name.split(":")[0] for x in frozen_func.outputs]
+
+    model_serialized = frozen_func.graph.as_graph_def().SerializeToString(deterministic=True)
```
Spartee, line 80: should this be an option? I'm guessing no but want to be sure.

Spartee (approved):
One tiny comment, but I'm approving this. I think we could open 1 more ticket about getting this into the smartsim-tutorials container, but I feel like thats outside the scope of this ticket.

al-rigazzi:
Added a test for the new function, coverage is stable (small increase).

PR #177, opened 2022-03-23
Don't hang on missing tag value

billschereriii:
Prevent the model writer from hanging if a tag is not matched. Also, add code to allow an unmatched tag to be treated as a fatal error instead of just a warning

### DIFF
--- a/smartsim/_core/generation/modelwriter.py
+++ b/smartsim/_core/generation/modelwriter.py
@@ -26,6 +26,8 @@
 
 import re
 
+from smartsim.error.errors import SmartSimError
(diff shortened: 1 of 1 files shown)


Spartee (approved):
LGTM thanks for hopping on this. I like that option you added. @ashao should confirm.

ashao:
Solved my problem! Now I can cleanup the NCAR_ML_EKE notebook even more.

ashao (approved):
I can confirm that this fixed the problem that I observed. Thanks @billschereriii

PR #180, opened 2022-03-30
Keydb support

EricGustin:
This PR adds the ability to use `KeyDB` in place of `Redis` as the database through a new `--keydb` flag for `smart build`. This flexibility enables users to maximize their throughput if they desire.

### DIFF
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -121,6 +121,8 @@ class Build:
         logger.info("SmartSim build complete!")
 
     def build_database(self):
+        if self.keydb:
+            CONFIG.conf_path = Path(CONFIG.core_path, "config", "keydb.conf")
         # check database installation
         database_name = "KeyDB" if self.keydb else "Redis"
         database_builder = builder.DatabaseBuilder(
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -123,16 +123,22 @@ class Build:
     def build_database(self):
         # check database installation
         database_name = "KeyDB" if self.keydb else "Redis"
-        database_builder = builder.DatabaseBuilder(self.build_env(), self.build_env.MALLOC, self.build_env.JOBS, self.verbose)
+        database_builder = builder.DatabaseBuilder(
+            self.build_env(), self.build_env.MALLOC, self.build_env.JOBS, self.verbose
+        )
 
         if not database_builder.is_built:
+            version = Version_("6.2.0") if self.keydb else self.versions.REDIS
             logger.info(
-                f"Building {database_name} version {self.versions.REDIS} from {self.versions.REDIS_URL}"
+                f"Building {database_name} version {version} from {self.versions.REDIS_URL}"
             )
-
-            database_builder.build_from_git(
-                self.versions.REDIS_URL, self.versions.REDIS_BRANCH
+            database_url = (
+                "https://github.com/EQ-Alpha/KeyDB"
+                if self.keydb
+                else self.versions.REDIS_URL
             )
+            database_branch = "v6.2.0" if self.keydb else self.versions.REDIS_BRANCH
+            database_builder.build_from_git(database_url, database_branch)
             database_builder.cleanup()
         logger.info(f"{database_name} build complete!")
 
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -88,9 +88,8 @@ class Build:
             self.build_env = BuildEnv()
 
             if self.verbose:
-                db_name = "KEYDB" if self.keydb else "REDIS"
                 logger.info("Build Environment:")
-                env = self.build_env.as_dict(db_name=db_name)
+                env = self.build_env.as_dict()
                 print(tabulate(env, headers=env.keys(), tablefmt="github"), "\n")
 
             logger.info("Checking requested versions...")
@@ -107,8 +106,9 @@ class Build:
                     )
 
             if self.verbose:
+                db_name = "KEYDB" if self.keydb else "REDIS"
                 logger.info("Version Information:")
-                vers = self.versions.as_dict()
+                vers = self.versions.as_dict(db_name=db_name)
                 print(tabulate(vers, headers=vers.keys(), tablefmt="github"), "\n")
 
             # REDIS/KeyDB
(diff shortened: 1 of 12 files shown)


EricGustin:
Should we add tests for this code with a new github actions job that does `smart build` with varying flags?

MattToast (commented):
lgtm!!

Everything seems to work on my end, only change I'd request is adding `smartsim/_core/bin/keydb-server` and `smartsim/_core/bin/keydb-cli` to the `.gitignore`

MattToast:
> Should we add tests for this code by using a new github actions job that does smart build with varying flags?

I think that seems like a great idea! If not on this PR maybe in a new ticket?

smartsim/_core/_cli/cli.py
```diff
@@ -59,12 +59,12 @@ def site(self):
 
     def dbcli(self):
         install_path = get_install_path()
-        script_path = install_path.joinpath("_core/bin/redis-cli").resolve()
+        script_path = next((install_path / "_core" / "bin").glob("*-cli")).resolve()
```
ashao, line 62: I'm a bit wary of manipulating files with the regexes like this. Safer to be a bit more explicit even if it ends up adding code.

smartsim/_core/_install/builder.py
```diff
@@ -131,8 +131,8 @@ def __init__(self, build_env={}, malloc="libc", jobs=None, verbose=False):
     @property
     def is_built(self):
         """Check if Redis is built"""
-        server = self.bin_path.joinpath("redis-server").is_file()
-        cli = self.bin_path.joinpath("redis-cli").is_file()
+        server = next(self.bin_path.glob("*-server"), None) is not None
```
ashao, line 134: Same comment here; avoid the wildcard and be explicit.

smartsim/_core/_cli/clean.py
```diff
@@ -53,7 +53,7 @@ def clean(self, _all=False):
 
         bin_path = self._core_path / "bin"
         if bin_path.is_dir() and _all:
-            files_to_remove = ["redis-server", "redis-cli"]
+            files_to_remove = ["redis-server", "redis-cli", "keydb-server", "keydb-cli"]
```
ashao, line 56: Probably out of scope for this PR, but something to keep in mind. We could add an artifact manifest to each of the builder objects. This would help us both sanity check that the build completely successfully (i.e. check every artifact and make sure it exists) and also help us simplify the clean stage by simply looping through all the builder objects and removing the artifacts.

Maybe write that up as a ticket?

smartsim/_core/_cli/build.py
```diff
@@ -66,8 +66,15 @@ def __init__(self):
             type=str,
             help="Path to custom libtensorflow directory (ONLY USED IF NEEDED)",
         )
+        parser.add_argument(
```
ashao, line 69: Instead of making it so that we either install Redis or KeyDB, it would be good to be able to build KeyDB as an additional option. We can then switch out which one they use by setting the REDIS_PATH variables and so more easily swap out between the two for development, testing, and benchmarking.

ashao (changes requested):
Nice piece of work and it dovetails well with some of the things I'm trying to do. Main comment is to improve the safety of the code by removing the wildcards when removing files.

Looking forward to having this functionality in.

smartsim/_core/_install/buildenv.py
```diff
@@ -211,10 +211,11 @@ class Versioner:
     ONNX = Version_(REDISAI.onnx)
 
     def as_dict(self):
+        database_name = "REDIS" if "redis" in self.REDIS_URL else "KEYDB"
```
Spartee, line 214: What if a user installed redis from a different url than the standard? their renamed fork? unlikely but possible.

smartsim/_core/_install/builder.py
```diff
@@ -142,16 +142,21 @@ def build_from_git(self, git_url, branch):
         :param branch: branch to checkout
         :type branch: str
         """
-        redis_build_path = Path(self.build_dir, "redis")
+        database_name = "redis" if "redis" in git_url else "KeyDB"
```
Spartee, line 145: same comment

smartsim/_core/_cli/build.py
```diff
@@ -87,13 +94,19 @@ def __init__(self):
             logger.info("Checking requested versions...")
             self.versions = Versioner()
 
+            if self.keydb:
+                self.versions.REDIS = Version_("6.2.0")
+                self.versions.REDIS_URL = "https://github.com/EQ-Alpha/KeyDB"
+                self.versions.REDIS_BRANCH = "v6.2.0"
+                CONFIG.conf_path = Path(CONFIG.core_path, "config", "keydb.conf")
```
Spartee, line 101: probably good to check this path if we don't somewhere else.

Spartee (changes requested):

smartsim/_core/_cli/clean.py
ashao: @Spartee :  Sure, I'll use `redis_builder` as an example:
```
class redis_builder():
  _expected_executable_targets = ['redis-cli', redis-server']
  def check_install(self):
    targets_installed = True
    for target in self._expected_targets:
      targets_installed = targets_installed and shutil.which(target)
      logger.warning(f'{target} could not be found')
  def clean(self):
    for target in _expected_executable_targets:
       shutil.rm(target)
```

smartsim/_core/_cli/build.py
ashao: @Spartee I get what you're saying and within a `modules` system it's easy enough to do. Thinking about this a little more, the origin of this comment is more along the lines of what we think the expected behavior should be if a user does something like:
```
smart build
```
and then later (weeks/months) does
```
smart build --keydb
```
and for whatever reason switching to KeyDB degrades performance in their application (e.g. they haven't futzed with the settings enough). How do they switch back? Right now either the user has to know/remember that they can set `REDIS_PATH` or they can `smart clobber` and build from scratch.

What I think the question boils down to is: are we sure enough that KeyDB is going to be a drop-in improvement over `redis`?

EricGustin: @ashao In the case of a user running `smart build --keydb` weeks/months later, since the implementation of #165, the user will no longer have to know/remember what to do as they will be informed by the cli on how to proceed. This probably doesn't address the entirety of your comment, but hopefully it at least partially does.

smartsim/_core/_cli/clean.py
Spartee: I like it. Thanks for clarifying

smartsim/_core/_cli/build.py
EricGustin: As in check if the path leads to a file that exists?

ashao (approved):

PR #181, opened 2022-03-31
Remove preset environment variables from slurmSettings and mpirunSettings

### LINKED ISSUE #128: Change the way environmental variables are passed (batch mode)
# Description
When in batch mode SmartSim creates the batch script then executed on target machine. Each generated srun execution in the script overwrites PATH, LD_LIBRARY_PATH and PYTHONPATH using export parameter of srun. This can be inconvenient if added preamble (using add_premable function) wishes to modify one of those since the result will never propagate to the target machine.

# Justification
Some use-cases require sourcing scripts before using specialised hardware. Those scripts often tend to modify above environmental variables. Generally user should be able to change those in preamble.

# Implementation Strategy
PATH, LD_LIBRARY_PATH and PYTHONPATH should not be changed by using export parameter of srun but rather in one of two following ways: 

1. Adding export option to generated SBATCH script : _#SBATCH --export_ and removing export from srun parameters of generated script. Srun will propagate those and the user will have a chance to add additional paths.
2. Simply modifying those variables as a part of sbatch script by adding _export PATH=$PATH:smartsim_paths...._ to the script after preamble.

This should not change the way SmartSim works while giving the user more flexibility on preamble scripts.

ben-albrecht:
Remove preset environment variables from slurmSettings and mpirunSettings since they are no longer needed.

Some other clean-ups:

- Check env var value types earlier in the process to ensure types are correct before launching any jobs
- Added a note to the `update_env` docs to inform users on how to have smartsim inherit the user environment
- Only throw `--export` flag if env_vars is not empty, otherwise we get an argument parsing error

Verified full testing passes on an HPC system with `SMARTSIM_TEST_LAUNCHER=slurm`.

Resolves #128

### DIFF
--- a/smartsim/_core/launcher/step/slurmStep.py
+++ b/smartsim/_core/launcher/step/slurmStep.py
@@ -138,7 +138,9 @@ class SrunStep(Step):
                 env_var_str,
                 comma_separated_env_vars,
             ) = self.run_settings.format_comma_sep_env_vars()
-            srun_cmd += ["--export", env_var_str]
+
+            if len(env_var_str) > 0:
+                srun_cmd += ["--export", env_var_str]
 
             if comma_separated_env_vars:
                 srun_cmd = ["env"] + comma_separated_env_vars + srun_cmd
@@ -199,7 +201,8 @@ class SrunStep(Step):
             cmd += mpmd.format_run_args()
             cmd += ["--job-name", self.name]
             (env_var_str, _) = mpmd.format_comma_sep_env_vars()
-            cmd += ["--export", env_var_str]
+            if len(env_var_str) > 0:
+                cmd += ["--export", env_var_str]
             cmd += mpmd.exe
             cmd += mpmd.exe_args
 
--- a/smartsim/settings/base.py
+++ b/smartsim/settings/base.py
@@ -368,7 +368,7 @@ class RunSettings:
         val_types = typing.get_args(val_types_union)
         # Coerce env_vars values to str as a convenience to user
         for (env, val) in env_vars.items():
-            if not isinstance(val, value_types):
+            if not isinstance(val, val_types):
                 raise TypeError(f"env_vars[{env}] was of type {type(val)}, not {val_types_union}")
             else:
                 self.env_vars[env] = str(val)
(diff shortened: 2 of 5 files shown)


smartsim/_core/launcher/step/slurmStep.py
```diff
@@ -138,7 +138,9 @@ def get_launch_cmd(self):
                 env_var_str,
                 comma_separated_env_vars,
             ) = self.run_settings.format_comma_sep_env_vars()
-            srun_cmd += ["--export", env_var_str]
+
+            if len(env_var_str) > 0:
```
al-rigazzi, line 142: Why is this necessary?

smartsim/settings/base.py
```diff
@@ -357,10 +357,25 @@ def run_command(self):
...
+        :raises TypeError: if env_vars values cannot be coerced to strings
         """
-        self.env_vars.update(env_vars)
+        val_types = (str, int, float, bool)
+        # Coerce env_vars values to str as a convenience to user
+        for (env, val) in env_vars.items():
+            if type(val) not in val_types:
+                raise TypeError(f"env_vars[{env}] was of type {type(val)}, not {val_types_union}")
+            else:
+                self.env_vars[env] = val
```
al-rigazzi, line 378: Shouldn't this be `self.env_vars[env] = str(val)`? Or, said differently, how are variables coerced to strings?

al-rigazzi (commented):
A couple of comments that I'd like to see addressed, otherwise looks good, and I'm sure this will improve the user experience a lot!

smartsim/settings/base.py
al-rigazzi: OK, I see, we do it later, that part of the code was collapsed.

ben-albrecht: Yeah, we could potentially cast types here, but I ended up leaving the behavior as is for this PR since we cast elsewhere and tests already exist to lock in the current behavior of not casting.

smartsim/_core/launcher/step/slurmStep.py
al-rigazzi: OK, makes sense.

al-rigazzi (approved):
Thanks for addressing my comments, LGTM!

Spartee (approved):
LGTM, worked on horizon

PR #182, opened 2022-04-02
Support for site-provided dependencies

ashao:
Includes a new function in the `smart` CLI to aid users whose site admins provide most of the packages. Additionally, if the appropriate environment variables are set, SmartSim will be configured to acknowledge the sitewide package installations.


smartsim/_core/_cli/cli.py
```diff
@@ -67,6 +68,8 @@ def dbcli(self):
             print("Redis dependencies not found")
             exit(1)
 
+    def sitebuild(self):
```
Spartee, line 71: I think this was meant to be taken out

smartsim/_core/_install/builder.py
```diff
@@ -39,9 +39,15 @@ def __init__(self, env, jobs=1, verbose=False):
 
         # Find _core directory and set up paths
         _core_dir = Path(os.path.abspath(__file__)).parent.parent
+
+        dependency_path = _core_dir
+        if os.getenv('SMARTSIM_DEP_PATH'):
```
Spartee, line 44: I think you ended up calling this ``SMARTSIM_DEP_INSTALL_PATH``

smartsim/_core/config/config.py
```diff
@@ -87,9 +87,14 @@ class Config:
     def __init__(self):
         # SmartSim/smartsim/_core
         self.core_path = Path(os.path.abspath(__file__)).parent.parent
-        self.lib_path = Path(self.core_path, "lib").resolve()
-        self.bin_path = Path(self.core_path, "bin").resolve()
-        self.conf_path = Path(self.core_path, "config", "redis6.conf")
+
+        dependency_path = self.core_path
+        if os.getenv('SMARTSIM_DEP_PATH'):
```
Spartee, line 92: how about 

```python
dependency_path = os.environ.get('SMARTSIM_DEP_INSTALL_PATH', self.core_path)
```

Spartee (changes requested):
couple comments. Will do another pass after looking through the site install PR

smartsim/_core/_cli/cli.py
ashao: Good catch

smartsim/_core/_install/builder.py
ashao: There's the `SMARTSIM_DEP_INSTALL_PATH` which is set in the builder's environment. On the user side of it, the environment modules set `SMARTSIM_DEP_PATH`.

ashao: Oh yeah, I remember why I called them different things. It's because of the way that environment variables set things. If I made it consistent we would incorrectly replace lines like:
```
set          smartsim_dep_path SMARTSIM_DEP_INSTALL_PATH

prepend-path LD_LIBRARY_PATH   $smartsim_dep_path/lib
prepend-path PATH              $smartsim_dep_path/bin

setenv       SMARTSIM_DEP_PATH SMARTSIM_DEP_INSTALL_PATH
```
While I could write some extra scripting to go around that, IMO it seems more reasonable to keep this the way it is. The rationale is that `SMARTSIM_DEP_INSTALL_PATH` is for a specific compiler chain, whereas `SMARTSIM_DEP_PATH` is the generic environment variable which can be changed by an environment module.

Spartee (approved):
Approving and should be merged once other PR is merged.

PR #185, opened 2022-04-05
Utilities to set ML models and scripts from driver scripts

al-rigazzi:
This PR adds the functionalities needed to add ML models and TorchScript functions to the orchestrators (convergend and not, but see caveat below) from the driver script launching SmartSim.

The full set of available new functionalities:
- set a ML model from a file on converged and non-converged orchestrators
- set a ML model from memory on non-converged orchestrators
- set a script from a file or from a string representation in memory on converged and non-converged orchestrators
- set a function on non-converged orchestrators

Basically, what can not be done is setting a Model or a function from memory on converged orchestrators. These are the reasons for such API gaps:
- an in-memory representation of a model is binary, to pass it to converged orchestrators we need to put it in the launcher script, and there are many problems when putting a binary string on a text file. Moreover, setting from memory should avoid writing to file, but as we would be dumping the model on the script, we are storing it anyhow. Better leave this to the user, so that models can be stored where needed.
- setting a function on the orchestrator requires passing the address of the function itself, which would not work once the context changes (i.e. when the converged setting is deployed).

I decided to use the SmartRedis client to connect to orchestrators and set models and scripts, as spawning `redis-cli` processes was more convoluted, in my opinion.


smartsim/_core/control/controller.py
```diff
@@ -591,3 +599,113 @@ def reload_saved_db(self, checkpoint_file):
             return orc
         finally:
             JM_LOCK.release()
+
+
+    def _enumerate_devices(self, dbobject: DBObject):
```
Spartee, line 604: I'm thinking this might be better as a method within ``DBObject``.

smartsim/_core/control/controller.py
```diff
@@ -591,3 +599,113 @@ def reload_saved_db(self, checkpoint_file):
...
+        if dbobject.device in ["CPU", "GPU"] and dbobject.devices_per_node > 1:
+            for device_num in range(dbobject.devices_per_node):
+                devices.append(f"{dbobject.device}:{str(device_num)}")
+        else:
+            devices = [dbobject.device]
+
+        return devices
+
+
+    def _set_ml_model(self, db_model, address, cluster=False):
```
Spartee, line 622: These two set methods could go into the ``_core/utils/redis.py`` file. Might clean up some of the imports and the base redis client is already imported there in case we want to to use that for validation or something in the future. 

I could also see these functions being used elsewhere. Possibly even with user facing functionality in the future. we should consider that here and split these into self contained functions.

smartsim/_core/control/controller.py
```diff
@@ -591,3 +599,113 @@ def reload_saved_db(self, checkpoint_file):
...
+                            function=db_script.script,
+                            device=device
+                        )
+        
+            except  RedisReplyError as error:
+                logger.error("Error while setting model on orchestrator.")
+                raise error
+
+
+    def _set_dbobjects(self, manifest):
```
Spartee, line 696: this will stay in the ``Controller`` (probably obviously)

smartsim/_core/launcher/colocated.py
```diff
@@ -142,17 +145,86 @@ def _build_colocated_wrapper_cmd(port=6780,
...
+
+def _build_db_model_cmd(db_models):
+    cmd = []
+    for db_model in db_models:
+        cmd.append("+db_model")
+        cmd.append(f"--name={db_model.name}")
+        if db_model.file:
+            cmd.append(f"--file={db_model.file}")
+        else:
+            err_msg = "ML model can not be set from memory for colocated databases.\n"
```
Spartee, line 182: can we detect this prior to launch? The ``Model`` object should be able to detect this when the user adds it to a colocated model and we can raise a ``ValueError``

smartsim/_core/launcher/colocated.py
```diff
@@ -142,17 +145,86 @@ def _build_colocated_wrapper_cmd(port=6780,
...
+
+
+def _build_db_script_cmd(db_scripts):
+    cmd = []
+    for db_script in db_scripts:
+        cmd.append("+db_script")
+        cmd.append(f"--name={db_script.name}")
+        if db_script.func:
+            if not isinstance(db_script.func, str):
+                err_msg = "Functions can not be set from memory for colocated databases.\n"
```
Spartee, line 214: see above comment about catching this earlier

smartsim/entity/__init__.py
```diff
@@ -3,3 +3,4 @@
 from .entity import SmartSimEntity
 from .entityList import EntityList
 from .model import Model
+from .dbobject import DBScript, DBModel, DBObject
```
Spartee, line 6: might be time to define an ``__all__`` here similar to settings when we need to simply import all entity definitions into a file.

smartsim/entity/model.py
```diff
@@ -213,6 +219,134 @@ def params_to_args(self):
...
+        One of either script (in memory string representation) or script_path (file)
+        must be provided
+
+        :param name: key to store script under
+        :type name: str
+        :param script: TorchScript code
+        :type script: str, optional
+        :param script_path: path to TorchScript code
+        :type script_path: str, optional
+        :param device: device for script execution, defaults to "CPU"
```
Spartee, line 301: Do we catch if a user puts in ``GPU:1`` and a ``devices_per_node`` of > 1? I see we already so some name checking but not sure if this case is caught.

Spartee (changes requested):
Great stuff! couple comments and 1 bigger design thought.

### comments
 - whats the workflow for ``Ensemble``? is the user required to write that loop themselves to call the ``Model`` functions?
 - tests for ``Ensembles``?

Should we allow the user to simply pass a ``DBModel`` to ``Experiment.start()`` in addition to the methods in the ``Model`` and ``Ensemble`` classes? ``DBObjects`` would be more user facing, but

smartsim/_core/control/controller.py
al-rigazzi: Done

smartsim/_core/control/controller.py
al-rigazzi: Yep, it stayed there with some mods

smartsim/entity/__init__.py
al-rigazzi: Done

smartsim/_core/launcher/colocated.py
al-rigazzi: Yes, this was easy with models, but it became a bit more difficult with ensembles. Done it for all cases.

smartsim/entity/model.py
al-rigazzi: We do now

smartsim/_core/control/controller.py
al-rigazzi: Moved it

al-rigazzi:
@Spartee re: passing the model to `Experiment.start()`. That would be an interesting idea, but I think that for now, I'd prefer to keep the `DBObject`s a `Model`-related entity. If multiple models need the same `DBObject`, they either belong to an ensemble (and this is addressed now), or they can all put the `DBObject`/the user can just set it to one of the models.

smartsim/entity/ensemble.py
```diff
@@ -298,3 +307,165 @@ def _read_model_parameters(self):
...
+            name=name,
+            script=function,
+            device=device,
+            devices_per_node=devices_per_node
+        )
+        self._db_scripts.append(db_script)
+        for entity in self:
+            self._extend_entity_db_scripts(entity, [db_script])
+
+    def _extend_entity_db_models(self, model, db_models):
```
Spartee, line 447: why aren't we just calling ``add_db_model`` here so that we don't have to do this error handling twice?

smartsim/entity/ensemble.py
```diff
@@ -298,3 +307,165 @@ def _read_model_parameters(self):
...
+                if model.colocated and not db_model.is_file:
+                    err_msg = "ML model can not be set from memory for colocated databases.\n"
+                    err_msg += f"Please store the ML model named {model.name} in binary format "
+                    err_msg += "and add it to the SmartSim Model as file."
+                    raise SSUnsupportedError(err_msg)
+
+                model._db_models.append(db_model)
+
+    def _extend_entity_db_scripts(self, model, db_scripts):
+
```
Spartee, line 461: same as comment above

smartsim/_core/control/controller.py
```diff
@@ -586,3 +594,38 @@ def reload_saved_db(self, checkpoint_file):
...
             JM_LOCK.release()
 
+
+    def _set_dbobjects(self, manifest):
+        db_addresses = self._jobs.get_db_host_addresses()
+        cluster = len(db_addresses) > 1
+        address = db_addresses[0]
+
+        try:
+            client = Client(address=address, cluster=cluster)
```
Spartee, line 604: So this should only be reachable in the case where we think the orchestrator is active but it's actually not active correct? 

could we re-use ``check_cluster_status``? or split off the orchestrator method for checking existance? lets chat about this one today.

Spartee (commented):
Very close, couple comments we should chat about today.

smartsim/entity/ensemble.py
al-rigazzi: I wanted to avoid creating copies of the same `DBModel` or `DBScript`, and just add references to the lists (this makes it also easier to check for duplicates). Maybe I can add a `_append_db_model` function in `Model`, in which I catch the exception, and call it from `add_db_model` both for `Ensemble` and `Model`. Same for script. That way, the try-catch will only happen in one location.

smartsim/_core/control/controller.py
al-rigazzi: Yep, right

smartsim/_core/control/manifest.py
```diff
@@ -194,3 +194,35 @@ def __str__(self):
...
+            has_db_objects |= hasattr(model, "_db_models")
+        has_db_objects |= any([has_db_models(model) | has_db_scripts(model) for model in self.models])
+        if has_db_objects:
+            return True
+
+        ensembles = self.ensembles
+        if not ensembles:
+            return False
+
+        has_db_objects |= any([has_db_models(ensemble) | has_db_scripts(ensemble) for ensemble in ensembles])
```
Spartee, line 220: add comments here to explain what's being done. Love the operator usage but it's not readable.

smartsim/_core/entrypoints/colocated.py
```diff
@@ -55,8 +57,107 @@
...
+                              func,
+                              args.device)
+        else:
+            for device_num in range(args.devices_per_node):
+                client.set_script(args.name, 
+                                  func, 
+                                  args.device+f":{device_num}")
+    elif args.file:
+        if args.devices_per_node == 1:
+            client.set_script_from_file(args.name,
```
Spartee, line 147: What happens if these fail? Have we tried setting a bad model?

smartsim/database/orchestrator.py
```diff
@@ -261,25 +261,8 @@ def is_active(self):
...
-                return False
-        # if a cluster
-        else:
-            try:
-                check_cluster_status(self._hosts, self.ports, trials=1)
-                return True
-            # we expect this to fail if the cluster is not active
-            except SSInternalError:
-                return False
+        return db_is_active(self._hosts, self.ports, self.num_shards)
```
Spartee, line 264: nice

Spartee (approved):
Approving with minor comments. Looks great. Works on my mac.

smartsim/_core/entrypoints/colocated.py
al-rigazzi: The functions are launched in a try-catch statement, I think it should catch such an exception. Do you think we should add a test for that case?

smartsim/_core/control/manifest.py
al-rigazzi: Yep, done.

PR #187, opened 2022-04-08
Kill jobs on KeyboardInterrupt

### LINKED ISSUE #48: Kill jobs on interrupt
# Description
When running jobs through SmartSim, it would be nice if a keyboard interrupt had the ability to cancel running jobs. Currently, when a deployment is being polled and a user interrupts the main thread, the launchers (PBS, Slurm, etc) will purposefully leave those jobs running. This is primarily so that ``Orchestrator``s can be reused between jobs.

# Justification
This feature will allow users to stop jobs by hitting CTRL-c during the execution of their jobs.

# Implementation Strategy
Currently the ``JobManager`` which updates the status of jobs, runs as a daemon thread in the background. This allows the main thread to remain unblocked in interactive settings like jupyter notebooks. 

so suppose the user interrupts the main thread during the execution, a multiprocessing ``Event`` could be used to signal to the ``JobManager`` to prompt the user a question of "Do you want to cancel all running jobs?". 

it is possible that ``atexit`` could be used to set the ``Event`` flag. 

We'd also have to be careful about how this works in Jupyter/Ipython, however, this could solve the issue of an kernel interrupts killing the database in jupyter.

Still more to think about here so labeling as design

EricGustin:
This PR resolves #48 by killing all running jobs whenever a `KeyboardInterrupt` is received. Previously, `KeyboardInterrupt` was handled when launching and polling by warning the user that some jobs may need to be manually cleaned up. This PR extends this behavior by automatically killing all running jobs, regardless of whether they are being polled. This means that when `block=False`, the jobs are still killed. The user still has the option to revert back to the previous behavior by passing `kill_on_interrupt=False` to `Experiment.start` and `Experiment.poll`.

### DIFF
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -26,6 +26,7 @@
 
 import os.path as osp
 import pickle
+import signal
 import threading
 import time
 
@@ -69,16 +70,13 @@ class Controller:
         The controller will start the job-manager thread upon
         execution of all jobs.
         """
-        try:
-            self._launch(manifest)
+        self._jobs.kill_on_interrupt = kill_on_interrupt
+        signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
+        self._launch(manifest)
 
-            # start the job manager thread if not already started
-            if not self._jobs.actively_monitoring:
-                self._jobs.start()
-
-        except KeyboardInterrupt:
-            self._jobs.signal_interrupt(kill_on_interrupt=kill_on_interrupt)
-            raise
+        # start the job manager thread if not already started
+        if not self._jobs.actively_monitoring:
+            self._jobs.start()
 
         # block until all non-database jobs are complete
         if block:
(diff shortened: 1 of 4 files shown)


al-rigazzi (approved):
Thanks for doing this, it will be much easier to interact with running (or about-to-run) jobs and kill them!

LGTM

smartsim/_core/control/controller.py
```diff
@@ -69,23 +70,19 @@ def start(self, manifest, block=True):
...
         execution of all jobs.
         """
-        try:
-            self._launch(manifest)
-
-            # start the job manager thread if not already started
-            if not self._jobs.actively_monitoring:
-                self._jobs.start()
+        self._jobs.kill_on_interrupt = kill_on_interrupt
+        signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
```
ben-albrecht, line 74: I would add a brief comment like `registering custom signal handler for ^C (SIGINT)` to make it clearer and stand out a bit more

smartsim/experiment.py
```diff
@@ -166,12 +166,18 @@ def start(self, *args, block=True, summary=False):
         :param summary: print a launch summary prior to launch,
                         defaults to False
         :type summary: bool, optional
+        :param kill_on_interrupt: flag for killing jobs when SIGINT is received
```
ben-albrecht, line 169: To clarify my own understanding, if `kill_on_interrupt` is `False`, do we exit the smartsim process without cleaning up the jobs? 

It might be helpful to clarify the behavior for True/False in this docstring (and in the previous function).

ben-albrecht (approved):
Looks great Eric. I have a couple nits to consider inline.

PR #191, opened 2022-04-20
Introduce multiversioned docs

EricGustin:
This PR adds a "Versions" section to the documentation's navigation bar so that users who many need to reference older versions of SmartSim will be able to do so. `doc/_static/version_names.txt` will need minor additions for each release. On each new release, the new version's name and documentation url will need to be added to `doc/_static/version_names.txt`.

To view this new feature do
`make docks`
`open docs/develop/index.html`

### DIFF
--- a/doc/_static/version_names.txt
+++ b/doc/_static/version_names.txt
@@ -1,0 +1,11 @@
+let version_names = `
+develop (unstable),
+0.4.0 (stable),
+0.3.2
+`;
+
+let version_urls = `
+https://www.craylabs.org/develop/overview.html,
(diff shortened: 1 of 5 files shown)


doc/_static/version_names.txt
```diff
@@ -0,0 +1,11 @@
+let version_names = `
+develop (unstable),
+0.4.0 (stable),
+0.3.2
+`;
```
Spartee, line 5: could this be a JSON file?

Spartee (approved):
LGTM. One question about the text file but otherwise good!

PR #193, opened 2022-04-26
Orchestrator and DBNode refactor

al-rigazzi:
This PR is a first shot at refactoring Orchestrator and DBNode classes. It mostly removes some if statements and separates code in more meaningful functions.

### DIFF
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -30,17 +30,15 @@ from shlex import split as sh_split
 from warnings import simplefilter, warn
 
 import psutil
-import redis
 from smartredis import Client
 from smartredis.error import RedisReplyError
 
 from .._core.utils import db_is_active
 from .._core.config import CONFIG
-from .._core.utils import check_cluster_status
 from .._core.utils.helpers import is_valid_cmd
 from .._core.utils.network import get_ip_from_host
 from ..entity import DBNode, EntityList
-from ..error import SmartSimError, SSConfigError, SSInternalError, SSUnsupportedError
+from ..error import SmartSimError, SSConfigError, SSUnsupportedError
 from ..log import get_logger
 from ..settings import (
     AprunSettings,
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -23,8 +23,8 @@
 # CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-import sys
 import itertools
+import sys
 from os import getcwd
 from shlex import split as sh_split
 from warnings import simplefilter, warn
@@ -34,8 +34,8 @@ import redis
 from smartredis import Client
 from smartredis.error import RedisReplyError
 
-from .._core.utils import check_cluster_status
 from .._core.config import CONFIG
+from .._core.utils import check_cluster_status
 from .._core.utils.helpers import is_valid_cmd
 from .._core.utils.network import get_ip_from_host
 from ..entity import DBNode, EntityList
@@ -595,9 +595,7 @@ class Orchestrator(EntityList):
             }
 
             if gpus_per_shard > 1:  # pragma: no-cover
-                erf_sets["gpu"] = (
-                    "{" + f"0-{gpus_per_shard-1}" + "}"
-                )
+                erf_sets["gpu"] = "{" + f"0-{gpus_per_shard-1}" + "}"
             elif gpus_per_shard > 0:
                 erf_sets["gpu"] = "{" + str(0) + "}"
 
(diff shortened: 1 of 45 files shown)


smartsim/entity/dbnode.py
```diff
@@ -179,119 +195,36 @@ def _parse_db_hosts(self):
...
-
+        # Find out if all shards' output streams are piped to separate files
+        if len(self._output_files) > 1:
+            for output_file in self._output_files:
+                filepath = osp.join(self.path, output_file)
+                ip = self._parse_db_host(filepath)
         else:
-            filepath = osp.join(self.path, self.name + ".out")
-            trials = 5
+            filepath = osp.join(self.path, self._output_files[0])
```
MattToast, line 204: I know that it is assumed that `len(self._output_files) >= 1`, but it still may be worth it to make an explicit check to protect against future misuse of the class. Perhaps somewhere during its construction?

MattToast (approved):
Looks great!! The new methods are super helpful and definitely help clear up the authors intent. Makes everything much easier to read.

I threw in a couple of comments regarding some minor changes, and depending on how closely we want to adhere to the black style guide, it may be worth running the formatter on orchestrator and dbnode.

Overall LGTM!!

smartsim/database/orchestrator.py
```diff
@@ -635,60 +624,61 @@ def _initialize_entities(self, **kwargs):
...
-                )
-            node = DBNode(db_node_name, self.path, run_settings, [port])
-            node._mpmd = True
-            node._shard_ids = range(self.db_nodes)
-            self.entities.append(node)
+            self.ports = [port]
 
-        self.ports = [port]
+    def _initialize_entities_mpmd(self, **kwargs):
+        port = kwargs.get("port", 6379)
```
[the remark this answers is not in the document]
MattToast: @al-rigazzi Given the number of methods that would need to be changed, I could absolutely see this taking up enough time to warrant a separate issue. I thought I saw this pattern in a few other places outside the DBNode that could be addressed in the new ticket as well.

smartsim/entity/ensemble.py
```diff
@@ -298,3 +307,151 @@ def _read_model_parameters(self):
...
+        orchestrator (converged or not) prior to the execution
+        of every entity belonging to this ensemble
+
+        One of either model (in memory representation) or model_path (file)
+        must be provided
+
+        :param name: key to store model under
+        :type name: str
+        :param model: model in memory
+        :type model: str, optional # TODO figure out what to type hint this as
```
MattToast, line 336: Could you explain what you mean by this?

smartsim/entity/ensemble.py
```diff
@@ -298,3 +307,151 @@ def _read_model_parameters(self):
...
+        )
+        self._db_scripts.append(db_script)
+        for entity in self:
+            self._extend_entity_db_scripts(entity, [db_script])
+
+    def _extend_entity_db_models(self, model, db_models):
+        entity_db_models = [db_model.name for db_model in model._db_models]
+        for db_model in db_models:
+            if not db_model.name in entity_db_models:
+                model._append_db_model(db_model)
```
MattToast, line 451: These new methods are both accessing "protected" methods on `Model`. Given how tightly integrated these two classes are, I am more than willing to call them "friend" classes, but I thought I should bring this to attention to make sure that this is the intended course of action.

MattToast (approved):
Looks great! I found a couple of minor things that I wanted to draw attention to bellow, but otherwise looks ready to go!!

smartsim/entity/dbobject.py
```diff
@@ -0,0 +1,204 @@
...
+            self.file = self._check_filepath(file_path)
+        else:
+            # Need to have this explicitly to check on it
+            self.file = None
+        self.device = self._check_device(device)
+        self.devices_per_node = devices_per_node
+
+    @property
+    def is_file(self):
+        if self.func:
```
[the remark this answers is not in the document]
MattToast: Nvm, I misread the rest of the method

smartsim/entity/ensemble.py
al-rigazzi: Yeah, good point. I think they are kinda C++ friends classes in a sense. The other possibility was to "unprotect" the methods, but I think we can leave it like this for the moment being.

smartsim/entity/ensemble.py
al-rigazzi: I think this was a leftover from old code (which was copied and pasted). In some cases this should be a string, in other cases a byte string, I think. Suggestions on how to express this?

smartsim/entity/dbobject.py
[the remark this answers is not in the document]
al-rigazzi: OK

smartsim/database/orchestrator.py
```diff
@@ -34,8 +34,9 @@
 from smartredis import Client
 from smartredis.error import RedisReplyError
 
-from .._core.utils import check_cluster_status
+from .._core.utils import db_is_active
 from .._core.config import CONFIG
+from .._core.utils import check_cluster_status
```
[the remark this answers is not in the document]
al-rigazzi: Removed in new commit.

smartsim/entity/ensemble.py
MattToast: I know if I where to do this with the typing package it would look something like `Optional[Union[str, bytes]]` or with 3.10 syntax `str | bytes | None`. Perhaps `str, bytes, optional` or directly use the 3.10 syntax?

smartsim/entity/ensemble.py
MattToast: Definitely going to agree that these methods should not be "unprotected." I don't a user should have access to the method. I like the idea of leaving this as is.

smartsim/entity/ensemble.py
al-rigazzi: Let's go for 3.10 syntax!

PR #195, opened 2022-05-09
Orchestrator port parametrized in tests implemented

amandarichardsonn:
Created a parameter that is used for the port argument when instantiating an Orchestrator in the test suite.

Originally, the default port parameter was a hardcoded port number. However, if two local tests run on the same system, local Orchestrators will all be launched on the same port which will result in test failure.

I added the method get_test_port() to the WLMUtils class in conftest.py and replaced each original port argument with the new parameter get_test_port(). Now, local Orchestrators will all be launched on different ports rather than the same.

### DIFF
--- a/smartsim/settings/base.py
+++ b/smartsim/settings/base.py
@@ -367,11 +367,13 @@ class RunSettings:
         :type env_vars: dict[str, Union[str, int, float, bool]]
         :raises TypeError: if env_vars values cannot be coerced to strings
         """
-        val_types_union = (str, int, float, bool)
+        val_types = (str, int, float, bool)
         # Coerce env_vars values to str as a convenience to user
-        for (env, val) in env_vars.items():
-            if type(val) not in val_types_union:
-                raise TypeError(f"env_vars[{env}] was of type {type(val)}, not {val_types_union}") 
+        for env, val in env_vars.items():
+            if not isinstance(val, val_types):
+                raise TypeError(
+                    f"env_vars[{env}] was of type {type(val)}, not {val_types}"
+                )
             else:
                 self.env_vars[env] = val
 
@@ -512,7 +514,6 @@ class RunSettings:
             else:
                 formatted.append(f"{key}={val}")
         return formatted
-                
 
     def __str__(self):  # pragma: no-cover
         string = f"Executable: {self.exe[0]}\n"


tests/test_launch_errors.py
```diff
@@ -31,15 +31,15 @@ def test_model_failure(fileutils):
...
     """Test error when users try to launch second orchestrator"""
     exp_name = "test-orc-error-on-relaunch"
     exp = Experiment(exp_name, launcher="local")
     test_dir = fileutils.make_test_dir()
 
-    orc = Orchestrator(port=6780)
+    orc = Orchestrator(port=wlmutils.get_test_port())
     orc.set_path(test_dir)
-    orc_1 = Orchestrator(port=6790)
+    orc_1 = Orchestrator(port=wlmutils.get_test_port())
```
al-rigazzi, line 42: Here we should use a different port for the second `Orchestrator`, to make sure it fails for the right reason (i.e. it fails because we do not allow for two `Orchestrator`s to be launched, not because the port is already taken). I would just add 1 to the port returned by `get_test_port()` and maybe we need to make an explicit reference to this in the test docs (or in the test headers).

tests/test_orchestrator.py
```diff
@@ -37,12 +37,12 @@ def test_inactive_orc_get_address():
...
 
 
-def test_orc_active_functions(fileutils):
+def test_orc_active_functions(fileutils, wlmutils):
     exp_name = "test_orc_active_functions"
     exp = Experiment(exp_name, launcher="local")
     test_dir = fileutils.make_test_dir()
 
-    db = Orchestrator(port=6780)
+    db = Orchestrator(port=wlmutils.get_test_port())
```
al-rigazzi, line 45: This is OK, but at line 54 we test that the db is running at the correct address. That way was OK as long as we were using a fixed port, can you update the `assert` statement so that it uses the right port?

conftest.py
```diff
@@ -23,6 +23,8 @@
 test_launcher = CONFIG.test_launcher
 test_device = CONFIG.test_device
 test_nic = CONFIG.test_interface
+test_port = CONFIG.test_port
```
al-rigazzi, line 26: Can we add a line about the used port in `print_test_configuration`? And maybe mention that ports `test_port` and `test_port+1` will be used (see another comment in this review)?

al-rigazzi (changes requested):
Good job! A couple of suggestions and changes to make, but all minor.

Also, can you make sure that the tests added with the last merge make use of `test_port`? I think they might need to use that and a couple more ports... I leave it up to you to pick a good strategy and a good way of communicating that to the user in `print_test_configuration`.

tests/test_orchestrator.py
al-rigazzi: Also, there is a problem with this test, if the orchestrator fails to launch, we don't shut it down. It would be great if you could fix this as part of this ticket. 
We implemented a couple workarounds for cases like this in other tests or you could possibly also use the `local_db` fixture... I leave it up to you!

tests/test_smartredis.py
```diff
@@ -31,7 +31,7 @@
 )
```
al-rigazzi, line 31: At the beginning of this file, there is a `REDIS_PORT` constant which we can now delete (and use your new utility wherever it appeared).

al-rigazzi:
All looks very good! I think you might just need to resolve all conflicts and then we are ready to merge.

smartsim/settings/base.py
```diff
@@ -367,12 +367,12 @@ def update_env(self, env_vars):
         :type env_vars: dict[str, Union[str, int, float, bool]]
         :raises TypeError: if env_vars values cannot be coerced to strings
         """
-        val_types = (str, int, float, bool)
+        val_types_union = (str, int, float, bool)
```
al-rigazzi, line 370: The code in this function was updated on `develop`, but it seems that by fixing the conflicts, we reverted to an older version. I'd suggest to checkout the `develop` version, as it is more readable.

al-rigazzi (commented):
Good work! Just one comment (and I think I am responsible for the wrong conflict fix). Once that is addressed, we can merge.

al-rigazzi (approved):
LGTM!

Great stuff!

amandarichardsonn:
Updating `port` argument in summary (#195 )

Replaced the hard coded `port` argument with the get_test_port() method in the WLMUtils class in conftest.py

[ committed by @amandarichardsonn ]
[ reviewed by @al-rigazzi ]

PR #197, opened 2022-05-12
Fix keydb directory name during build

mellis13:
Currently ``database_name`` is set to ``KeyDB``, but when the command ``smart build --keydb`` is executed the following error is encountered:

```bash
[SmartSim] ERROR [Errno 2] No such file or directory: 'SmartSim/smartsim/_core/.third-party/keydb'
```

This PR changes the ``database_name`` to ``keydb``.


al-rigazzi (approved):
LGTM, thanks for catching this!

PR #200, opened 2022-05-18
Updated version of Black

amandarichardsonn:
Added version of Click in set-up.py

Now when users install SmartSim in a new virtual environment and run `pip install -e .\[ml,dev\] `, then `make style`, the updated version of Click will now work with the existing version of Black

`make style` will pass

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,4 +11,5 @@ isort>=5.6.4
 pylint>=2.6.0
 pytest>=6.0.0
 pytest-cov>=2.10.1
+click==8.0.2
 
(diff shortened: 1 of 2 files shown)


al-rigazzi (commented):
Good job! Even though `requirements-dev.txt` is not used in our standard build pipeline, I think it is a good tool for developers to quickly check what dependencies there are. I would suggest to add `click` to that file too.
Also, thanks for catching the conflict which followed from the strategy which was initially proposed!

al-rigazzi (approved):
LGTM!

PR #202, opened 2022-06-03
Avoid checking for alloc_specs if it is not defined

al-rigazzi:
This PR mitigates issues when a Slurm or PBS system is set up so that not all environment variables are populated. 
If the `alloc_specs` fixture is not defined (see commit 4bd5e23), the functions using it won't be testsd.

### DIFF
--- a/tests/full_wlm/test_wlm_helper_functions.py
+++ b/tests/full_wlm/test_wlm_helper_functions.py
@@ -8,6 +8,8 @@ from smartsim.error.errors import LauncherError, SmartSimError, SSUnsupportedErr
 
 
 def test_get_hosts(alloc_specs):
+    if not alloc_specs:
+        pytest.skip("alloc_specs not defined")
     def verify_output(output):
         assert isinstance(output, list)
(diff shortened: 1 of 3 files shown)


tests/full_wlm/test_wlm_helper_functions.py
```diff
@@ -8,6 +8,8 @@
```
ben-albrecht, line 9: Would you mind adding a short comment here explaining why alloc_specs might not be defined? E.g.,

```
# Slurm or PBS system can sometimes be set up such that not all 
#   environment variables associated with 'alloc_specs' are populated,
#   so 'alloc_specs'-related tests only run when the environment variables are set
```

ben-albrecht (approved):
Looks good, barring a request for comment inline

tests/full_wlm/test_wlm_helper_functions.py
al-rigazzi: I'll add it to the tests. This feature is still under-documented, anyhow.

PR #203, opened 2022-06-08
Doc fix

amandarichardsonn:
Creating PR for changes, not done!

### DIFF
--- a/doc/sr_python_walkthrough.rst
+++ b/doc/sr_python_walkthrough.rst
@@ -3,14 +3,13 @@
 Python
 ******
 
-This section will detail the SmartRedis Python client and how to
-use it within SmartSim applications and RedisAI in general.
+This section details the SmartRedis Python client to demonstrate its general use within SmartSim applications and RedisAI.
 
 
 .. note::
-      The Python API examples are written to connect to a
-      database at ``127.0.0.1:6379``.  When running this example,
-      ensure that the address and port of your Redis instance are used.
+      The following Python API examples connect to a
+      database at the address:port ``127.0.0.1:6379``.  When replicating the following example,
+      ensure that you use the address:port of your local Redis instance.
 
 
 
--- a/doc/sr_python_walkthrough.rst
+++ b/doc/sr_python_walkthrough.rst
@@ -16,11 +16,10 @@ This section details the SmartRedis Python client to demonstrate its general use
 Tensors
 =======
 
-The Python client has the ability to send and receive tensors from
-the Redis database.  The tensors are stored in the Redis database
-as RedisAI data structures.  Additionally, Python client API
+The Python client can exchange tensors from
+the Redis database stored as RedisAI data structures.  Additionally, Python client API
 functions involving tensor data are compatible with Numpy arrays
-and do not require any other data types.
+and do not require other data types.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_put_get_tensor.py
   :language: python
@@ -35,7 +34,7 @@ The Python client can store and retrieve tensors and metadata in datasets.
 For further information about datasets, please refer to the :ref:`Dataset
 section of the Data Structures documentation page <data_structures_dataset>`.
 
-The code below shows how to store and retrieve tensors which belong to a ``DataSet``.
+The code below shows how to store and retrieve tensors that belong to a ``DataSet``.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_put_get_dataset.py
   :language: python
@@ -45,9 +44,9 @@ The code below shows how to store and retrieve tensors which belong to a ``DataS
 Models
 ======
 
-The SmartRedis clients allow users to set and use a PyTorch, ONNX, TensorFlow,
+The SmartRedis client allows users to set and use a PyTorch, ONNX, TensorFlow,
 or TensorFlow Lite model in the database. Models can be sent to the database directly
-from memory or from a file. The code below illustrates how a
+from memory or a file. The code below illustrates how a
 jit-traced PyTorch model can be used with the Python client library.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_model_torch.py
@@ -55,7 +54,7 @@ jit-traced PyTorch model can be used with the Python client library.
   :linenos:
   :lines: 26-71
 
-Models can also be set from a file, as in the code below.
+Users can set models from a file, as shown in the code below.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_model_file_torch.py
   :language: python
@@ -66,27 +65,25 @@ Scripts
 =======
 
 Scripts are a way to store python-executable code in the database. The Python
-client can send scripts to the dataset from a file, or directly from memory.
+client can send scripts to the dataset from a file or directly from memory.
 
-As an example, the code below illustrates how a function can be defined and sent
-to the database on the fly, without storing it in an intermediate file.
+The code below illustrates how to avoid storing a function in an intermediate file. 
+With this technique, we can define and send a function to the database on the fly.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_script.py
   :language: python
   :linenos:
   :lines: 26-66
 
-The code below shows how to set a script from a file.  Running the
-script set from file uses the same API calls as the example shown
-above.
+The code below shows how to set a script from a file. Running the script set from 
+the file uses the same API calls like the example shown above.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_script_file.py
   :language: python
   :linenos:
   :lines: 26-41
 
-The content of the script file has to be written
-in Python. For the example above, the file ``data_processing_script.txt``
+The content of the script file requires Python. For the example above, the file ``data_processing_script.txt``
 looks like this:
 
 .. literalinclude:: ../smartredis/examples/serial/python/data_processing_script.txt


doc/sr_cpp_walkthrough.rst
```diff
@@ -3,21 +3,19 @@ C++
...
+This section presents examples that use the SmartRedis C++ API to 
+interact with the RedisAI tensor, model, and script data types. 
+Additionally, this section demonstrates an example of utilizing the SmartRedis ``DataSet`` API.
 
 
 
 .. note::
       The C++ API examples rely on the ``SSDB`` environment
-      variable being set to the address and port of the Redis database.
+      variable set to the address and port of the Redis database.
```
billschereriii, line 14: how about "having been set" -- what we're relying on here is that the variable has to be set before we can proceed

doc/sr_cpp_walkthrough.rst
```diff
@@ -3,21 +3,19 @@ C++
...
 
 
 .. note::
-    The C++ API examples are written
-    to connect to a clustered database or clustered SmartSim Orchestrator.
+    The C++ API examples connect to a clustered database or clustered SmartSim Orchestrator.
```
billschereriii, line 18: Not sure I like this change. It's only a minor tweak to make the examples switch to non-clustered (standalone) Redis

doc/sr_python_walkthrough.rst
```diff
@@ -46,17 +44,17 @@ The code below shows how to store and retrieve tensors which belong to a ``DataS
 Models
 ======
 
-The SmartRedis clients allow users to set and use a PyTorch, ONNX, TensorFlow,
+The SmartRedis client allows users to set and use a PyTorch, ONNX, TensorFlow,
```
billschereriii, line 47: we actually have four clients -- one each for C, C++, Python, Fortran -- so the original statement was correct

billschereriii:
Some really good updates here, thank you! A few of the updates need to be adjusteed or reverted though -- see inline above

doc/sr_python_walkthrough.rst
amandarichardsonn: Ahhhhh I see!

doc/sr_cpp_walkthrough.rst
billschereriii: Getting rid of passive tense is definitely a good thing.  Yes, SSDB has to be manually set by the user (if they set up their own Redis cluster), but it's automatically set up by SmartSim when you go through the launcher.

doc/sr_cpp_walkthrough.rst
amandarichardsonn: What about -> "The C++ API examples connect either to a clustered database or clustered SmartSim Orch"

I am trying to avoid the use of 'are'

amandarichardsonn: wait thats basically the same as before

doc/sr_cpp_walkthrough.rst
amandarichardsonn: Is it worth it adding this short info? -> "If a User sets up their own Redis cluster, then they must manually set the address and port of their Redis database, otherwise it is automatically set through the launcher"

amandarichardsonn: I think I was being extra, and also I am having trouble thinking of another way to put it without it turning into something super wordy

doc/sr_python_walkthrough.rst
```diff
@@ -67,27 +65,25 @@ Scripts
...
 .. literalinclude:: ../smartredis/examples/serial/python/example_script.py
   :language: python
   :linenos:
   :lines: 26-66
 
-The code below shows how to set a script from a file.  Running the
-script set from file uses the same API calls as the example shown
-above.
+The code below shows how to set a script from a file. Running the script set from 
+the file uses the same API calls like the example shown above.
```
billschereriii, line 79: change `like` to `as`

doc/sr_python_walkthrough.rst
```diff
@@ -3,25 +3,23 @@
...
 
 
 Tensors
 =======
 
-The Python client has the ability to send and receive tensors from
-the Redis database.  The tensors are stored in the Redis database
-as RedisAI data structures.  Additionally, Python client API
+The Python client can exchange tensors from
+the Redis database stored as RedisAI data structures.  Additionally, Python client API
```
billschereriii, line 20: How about: "The Python client can send and receive tensors from the Redis database, where they are stored as RedisAI data structures."

doc/sr_python_walkthrough.rst
```diff
@@ -3,25 +3,23 @@
 Python
 ******
 
-This section will detail the SmartRedis Python client and how to
-use it within SmartSim applications and RedisAI in general.
+This section details the SmartRedis Python client to demonstrate its general use within SmartSim applications and RedisAI.
```
billschereriii, line 6: Suggest changing `to demonstrate` to `and demonstrates`

doc/sr_cpp_walkthrough.rst
```diff
@@ -69,8 +67,8 @@ executes a preprocessing script.
 Scripts
 =======
 
-The example in :ref:`SR CPP Models` shows how to store, and use a PyTorch script
-in the database with the C++ Client.  The script is stored a file
+The example in :ref:`SR CPP Models` shows how to store and use a PyTorch script
+in the database with the C++ Client.  The script stores as a file
```
billschereriii, line 71: The script is the file we're talking about here, so while you were right to notice that this text needed updating, I think the correction should be `is stored as`

doc/sr_cpp_walkthrough.rst
```diff
@@ -54,8 +52,8 @@ which belong to a ``DataSet``.
 Models
 ======
 
-The following example shows how to store, and use a DL model
-in the database with the C++ Client.  The model is stored as a file
+The following example shows how to store and use a DL model
+in the database with the C++ Client.  The model stores as a file
```
billschereriii, line 56: Same as the next comment (I'm working from bottom to top)

doc/sr_cpp_walkthrough.rst
```diff
@@ -3,10 +3,9 @@ C++
 ***
 
 
-In this section, examples are presented using the SmartRedis C++
-API to interact with the RedisAI tensor, model, and script
-data types.  Additionally, an example of utilizing the
-SmartRedis ``DataSet`` API is also provided.
+This section presents examples that use the SmartRedis C++ API to 
+interact with the RedisAI tensor, model, and script data types. 
+Additionally, this section demonstrates an example of utilizing the SmartRedis ``DataSet`` API.
```
ashao, line 8: Could you make sure the line lengths follow black convention? (88 characters per line)

doc/sr_cpp_walkthrough.rst
```diff
@@ -16,8 +15,7 @@ SmartRedis ``DataSet`` API is also provided.
 
 
 .. note::
-    The C++ API examples are written
-    to connect to a clustered database or clustered SmartSim Orchestrator.
+    The C++ API examples are written to connect to a clustered database or clustered SmartSim Orchestrator.
```
ashao, line 18: Line length

doc/sr_python_walkthrough.rst
ashao: Tacking on to this one to say `as in`

doc/sr_python_walkthrough.rst
```diff
@@ -67,27 +65,25 @@ Scripts
...
+the file uses the same API calls like the example shown above.
 
 .. literalinclude:: ../smartredis/examples/serial/python/example_script_file.py
   :language: python
   :linenos:
   :lines: 26-41
 
-The content of the script file has to be written
-in Python. For the example above, the file ``data_processing_script.txt``
+The content of the script file requires Python. For the example above, the file ``data_processing_script.txt``
```
ashao, line 86: This was originally awkwardly worded, but the rewrite also sounds a little odd in that it decouples Python from the inherent essence of what the file is. Perhaps change to: `This file must be a valid Python script`

ashao (changes requested):
Some minor suggestions. The main thing is to make sure that the line lengths do not exceed 88 characters. While we can rely on `black` to fix it for us, in between those times, it makes the code nicer to look at.

doc/sr_python_walkthrough.rst
amandarichardsonn: Oop it does sound odd, I will investigate

doc/sr_python_walkthrough.rst
amandarichardsonn: I like that much better

doc/sr_cpp_walkthrough.rst
amandarichardsonn: Okay I think I fixed the line lengths

doc/sr_python_walkthrough.rst
billschereriii: probably

billschereriii (approved):
Looks great! Thanks for pushing through and getting these documentation clarifications in place!

ashao (approved):

amandarichardsonn:
Edits made to SmartSim documentation to improve writing quality (#203 )

Replacing passive tense with present in SmartSim documentation. Also editing grammar and punctuation errors. 

[ committed by @amandarichardsonn  ]
[ reviewed by @billschereriii @ashao  ]

Issue #101, opened 2021-10-19
Singularity launching support 


Spartee:
# Description
SmartSim should be able to use ``singularity``, specifically ``singularity exec`` to launch mpi applications that have been containerized. This should be able to work with a number of different run commands (i.e. ``srun singularity exec``.

# Justification
I will not list out the benefits of putting applications in containers because there are too many to list, however, SmartSim being able to launch such containers would be one more way SmartSim can contribute to reproducible science. Having each stage of a driver script be a containerized application could greatly benefit scientists in terms of sharing and reproducibility.

# Implementation Strategy
I can think of two ways that this can be implemented.

1. as it's own ``RunSettings`` child class.

This approach would create a ``SingularityRunSettings`` class (possibly by a shorter name) that would take in the run command (i.e. ``srun``) and other run arguments. Based on the run command that is passed, a corresponding ``RunSettings`` instance would be created in the class that all class methods that are typical for ``RunSettings`` would pass to. For example:

```python

rs = SingularityRunSettings(run_command="aprun")
rs.set_tasks(10)

```
under the hood this would just pass the argument ``10`` to the ``AprunSettings`` class created and held as an instance variable within the ``SingularityRunSettings`` class.

2. As a base class method for all implementations of ``RunSettings``

this method is likely easier since this method could be promoted to the ``RunSettings`` base class, but essentially this approach is a method that sets a flag for if something is to be launched as a singularity container. 

```python
aprun = AprunSettings(exe="hello_world")
aprun.use_singularity("/path/to/.sif/file", sing_args=["--nv"])
```
this method would do the following
1. check for singularity on the system. (might be delegated to ``Step`` level)
2. change exe for run settings to ``singularity`` and exe_args to start with ``exec`` 
3. add user provided flags. 

# Acceptance Criteria

 - [x] Choose an above implementation strategy write up a design document with a summary included in this ticket

ben-albrecht:
> Choose an above implementation strategy write up a design document with a summary included in this ticket

After discussion within the team, we have reached consensus on the following approach:

Container runtimes will be supported in a class hierarchy as shown:


```python
# container.py
class Container:
  ...

class Singularity(Container):
  ...

class Docker(Container):
  ...

class Podman(Container):
  ...
```


Users will instantiate this class with all the relevant container information, such as path to image, container arguments, and paths to bind/mount (if they don't want to specify them as container args). Below is a simple example:

```python
# user-experiment.py
from smartsim.settings import container

[...]

PR #204, opened 2022-06-08
SmartSim Singularity Integration

ben-albrecht:
This PR adds the ability for SmartSim to launch workloads in Singularity (Apptainer) as described in https://github.com/CrayLabs/SmartSim/issues/101. It also lays the groundwork for supporting other container runtimes such as docker, shifter, and podman in the future, as well as launching the orchestrator in a container.

## Design Variations

During development, it became clear that a few design changes from the original proposal were required. I have documented them below with their rationale:

#### 1. Argument name: `bind_paths` -> `mount`

The terms bind path and mount are mostly used interchangeably across different container runtimes. When writing tests, I kept forgetting if it was `bind_path` or `bind_paths`, which hinted to me that it was not a great arg name, so I swapped it out for the more succinct and easy to remember `mount`.

#### 2. `create_run_settings(..., container: str)` -> `create_run_settings(..., container: Container)`

We originally thought it would be easier for users to get started with containers by allowing them to pass a string into `create_run_settings(container='singularity')` instead of having to create a Container object. While writing tests, I realized that this was potentially very confusing for users because 1) the `container` arg types change between `create_run_settings` and `RunSettings`, and 2) if you need to add other metadata such as container args or container mount paths, you have to switch from using `create_run_settings` to `RunSettings` in your code, which is very annoying. Because creating Containers is so lightweight, I think it is best to keep the container interface consistent across all functions that accept them.

#### 3. dropped getter/setter methods

Because command generation and validation happens upon execution, users can freely modify `Container.args` and `Container.mount` without getter/setter methods to update any other state. Therefore, I dropped these methods from the interface.

## Testing

Added 2 test suites for containers: One for WLM testing and one for local testing. The local testing suite runs in GitHub actions. Due to the added time from building Singularity and pulling the `container-testing` image, the singularity testing only happens on one configuration of the build matrix: python 3.9 + redisai 1.2.5 on linux

### DIFF
--- a/tests/test_containers.py
+++ b/tests/test_containers.py
@@ -1,5 +1,7 @@
 import pytest
 from shutil import which
+from pathlib import Path
+import os
 
 from smartsim import Experiment, status
 from smartsim._core.utils import installed_redisai_backends
@@ -49,6 +51,7 @@ def test_singularity_commands(fileutils):
 
 @pytest.mark.skipif(not singularity_exists, reason="Test needs singularity to run")
 def test_singularity_basic(fileutils):
+    '''Basic argument-less Singularity test'''
     test_dir = fileutils.make_test_dir()
 
     container = Singularity(containerURI)
@@ -65,9 +68,34 @@ def test_singularity_basic(fileutils):
     exp.start(model, summary=False)
 
     # get and confirm status
-    statuses = exp.get_status(model)
-    if not all([stat == status.STATUS_COMPLETED for stat in statuses]):
-        assert False  # client ensemble failed
+    stat = exp.get_status(model)[0]
+    assert stat == status.STATUS_COMPLETED
+
+    print(exp.summary())
+
+
+@pytest.mark.skipif(not singularity_exists, reason="Test needs singularity to run")
+def test_singularity_args(fileutils):
+    '''Test combinations of args and mount arguments for Singularity'''
+    test_dir = fileutils.make_test_dir()
+    hometest_dir = os.path.join(str(Path.home()), 'test') # $HOME/test
+    mount_paths = {test_dir + '/singularity_args': hometest_dir}
+    container = Singularity(containerURI, args='--contain', mount=mount_paths)
+
+    exp = Experiment("singularity_args", launcher="local", exp_path=test_dir)
+
+    run_settings = exp.create_run_settings('python3', 'test/check_dirs.py',
+                                           container=container)
+    model = exp.create_model("singularity_args", run_settings)
+    script = fileutils.get_test_conf_path("check_dirs.py")
+    model.attach_generator_files(to_copy=[script])
+    exp.generate(model)
+
+    exp.start(model, summary=False)
+
+    # get and confirm status
+    stat = exp.get_status(model)[0]
+    assert stat == status.STATUS_COMPLETED
 
     print(exp.summary())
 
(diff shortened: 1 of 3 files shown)


smartsim/settings/slurmSettings.py
```diff
@@ -89,6 +89,10 @@ def make_mpmd(self, srun_settings):
             raise SSUnsupportedError(
                 "Colocated models cannot be run as a mpmd workload"
             )
+        if self.container:
+            raise SSUnsupportedError(
+                "Contanerized MPMD workloads are not yet supported."
```
ashao, line 94: Contanerized -> Containerized

tests/on_wlm/test_containers_wlm.py
```diff
@@ -0,0 +1,70 @@
...
+def test_singularity_wlm_smartredis(fileutils, wlmutils):
+    """Run two processes, each process puts a tensor on
+    the DB, then accesses the other process's tensor.
+    Finally, the tensor is used to run a model.
+
+    Note: This is a containerized port of test_smartredis.py for WLM system
+    """
+
+    launcher = wlmutils.get_test_launcher()
+    if launcher not in ["pbs", "slurm"]:
```
ashao, line 23: Change this to just `pbs` for now OR change launchers to `auto` instead of `slurm` below

tests/on_wlm/test_containers_wlm.py
```diff
@@ -0,0 +1,70 @@
...
+    exp = Experiment(
+        "smartredis_ensemble_exchange", exp_path=test_dir, launcher="slurm"
+    )
+
+    # create and start a database
+    orc = exp.create_database()
+    exp.generate(orc)
+    exp.start(orc, block=False)
+
+    container = Singularity('docker://benalbrecht10/smartsim-testing:latest')
```
ashao, line 38: Could you make this a module file so that it's completely obvious to us whenever we get around to migrating the path to a craylabs one?

tests/test_containers.py
```diff
@@ -0,0 +1,100 @@
...
+    exp = Experiment(
+        "smartredis_ensemble_exchange", exp_path=test_dir, launcher="local"
+    )
+
+    # create and start a database
+    orc = Orchestrator(port=wlmutils.get_test_port())
+    exp.generate(orc)
+    exp.start(orc, block=False)
+
+    container = Singularity('docker://benalbrecht10/smartsim-testing:latest')
```
ashao, line 68: Same as above

tests/on_wlm/test_containers_wlm.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+    test_dir = fileutils.make_test_dir()
+    exp = Experiment(
+        "smartredis_ensemble_exchange", exp_path=test_dir, launcher="slurm"
+    )
+
+    # create and start a database
+    orc = exp.create_database()
+    exp.generate(orc)
```
ashao, line 35: `orc` may not be necessary for this.

tests/on_wlm/test_containers_wlm.py
ben-albrecht: I moved to a global variable per test. Also, we've updated to use a repo that @al-rigazzi has set up now, so there's no need to migrate this anymore (unless we eventually do create a CrayLabs dockerhub org).

tests/on_wlm/test_containers_wlm.py
ben-albrecht: It turned out you are correct.

ashao (approved):
Thanks @ben-albrecht for the changes and for the thoroughness of the tests and implementation. The PR description is also very helpful for us going forward to understand how and why some of the implementation deviated from the original design decisions. They are all fully justified and make sense to me.

PR #205, opened 2022-06-09
Singularity API documentations

ben-albrecht:
Add Singularity docs to SmartSim API docs.

I was on the fence about exposing the base class, `Container`. It has no use for a user, but it is the  type referenced in arguments, e.g. 

```python
  def create_run_settings(self, ..., container: Container, ...):
    ...
```

I opted for leaving `Container` docs out in this PR, but am open to changing that if the reviewer disagrees. This decision can also be revisited in the future, if users are finding the current form confusing.

### DIFF
--- a/smartsim/settings/containers.py
+++ b/smartsim/settings/containers.py
@@ -40,7 +40,8 @@ class Container():
 
 
 class Singularity(Container):
-    '''Singularity (apptainer) container type.
+    '''Singularity (apptainer) container type. To be passed into a
+    ``RunSettings`` class initializer or ``Experiment.create_run_settings``.
 
     .. note::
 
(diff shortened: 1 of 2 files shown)


smartsim/settings/containers.py
```diff
@@ -53,7 +54,7 @@ class Singularity(Container):
         `system administrator <https://apptainer.org/docs/admin/1.0/configfiles.html#bind-mount-management>`_
 
 
-    :param image: local or remote path to container image, e.g. 'docker://sylabsio/lolcow'
+    :param image: local or remote path to container image, e.g. ``docker://sylabsio/lolcow``
```
ashao, line 57: We need to hear more about this `lolcow` image.

ashao (approved):

smartsim/settings/containers.py
ben-albrecht: It's a fun example from their [documentation](https://apptainer.org/docs/user/main/quick_start.html#executing-commands):

```
$ apptainer exec docker://sylabsio/lolcow cowsay "Fresh from the internet!"
 _________________________
< Fresh from the internet! >
 -------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

ben-albrecht:
For the archival records: @ashao agrees we don't need to expose the base `Container` class, since it is not user-facing beyond the fact that users need to know other Containers inherit from the `Container` type, which is stated in the docstring.

PR #206, opened 2022-06-09
Add documentation of new environment variables

billschereriii:
Kill this


billschereriii:
Erroneously created. Discard

PR #207, opened 2022-06-13
Validate redis-server/redis-cli install

billschereriii:
Add validation of redis-server and redis-cli installation, based on how we'll expect to find them in config, to builder class

### DIFF
--- a/smartsim/_core/_install/builder.py
+++ b/smartsim/_core/_install/builder.py
@@ -6,7 +6,7 @@ import subprocess
 import sys
 from pathlib import Path
 from subprocess import SubprocessError
-import ..utils.helpers
+from shutil import which
 
 # NOTE: This will be imported by setup.py and hence no
 #       smartsim related items should be imported into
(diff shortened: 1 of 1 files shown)


billschereriii:
Not happy about wholesale cloning of `expand_exe_path()` -- can someone suggest a better option?

smartsim/_core/_install/builder.py
```diff
@@ -15,6 +16,25 @@
 #   - check cmake version and use system if possible to avoid conflicts
 
 
+def expand_exe_path(exe):
```
ashao, line 19: While duplicating code is no fun, for the case of things that are used for `setup.py` and all that, it's better to err on the side of keeping it clean. As we discussed briefly, the only thing that I can think of is to import the function in the utils from here and make it available. I say keep it as is, unless @al-rigazzi has any objections.

ashao (approved):
I can confirm that this should not impact the site-installs (so long as the user has followed the instructions and imported the modules beforehand). Thanks for getting this in @billschereriii

al-rigazzi (approved):
LGTM, thanks @billschereriii, this will make it easier to spot broken installs!

PR #210, opened 2022-06-15
Fix singularity local launcher

ashao:
When not using Apptainer, the singularity tests were failing on
the local launcher because the SmartSim-specified run directories
were not being mounted. To solve this, when the launch command
is constructed, the working directory is passed to the container's
command constructor and automatically bound. For advanced users,
the working directory can be specified when the container object
is created in which case that directory takes precedence.


tests/test_containers.py
```diff
@@ -63,7 +63,7 @@ def test_singularity_basic(fileutils):
 
     script = fileutils.get_test_conf_path("sleep.py")
     model.attach_generator_files(to_copy=[script])
-    exp.generate(model)
+    exp.generate(model, overwrite=True)
```
al-rigazzi, line 66: Why is `overwrite=True` needed? Theoretically, test directories are wiped out before test execution (and after, if all tests passed). The experiment tree should not be there.

al-rigazzi (approved):
Good stuff, Andrew! I'm so glad you identified and fixed this issue! I just added a comment, if you could address that, then it LGTM

tests/test_containers.py
ashao: While running the tests manually (i.e. `pytest test_containers.py`), the output didn't seem to be consistently cleared. Happy to take this out if this was some weird edge case and/or detrimental to debugging test failures.

ashao: Yep, you were totally right...I must have just phase shifted into a parallel universe briefly.

PR #211, opened 2022-06-16
Skip DBModel test file when TF not available

al-rigazzi:
Currently, running tests when TensorFlow is not installed will result in a failed test. This should not happen. This PR fixes it.

### DIFF
--- a/tests/backends/test_dbmodel.py
+++ b/tests/backends/test_dbmodel.py
@@ -57,7 +57,6 @@ def create_tf_cnn():
     return serialize_model(model)
 
 
-@pytest.mark.skipif(not should_run, reason="Test needs TF to run")
 def test_db_model(fileutils, wlmutils):
     """Test DB Models on remote DB"""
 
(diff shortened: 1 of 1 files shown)


billschereriii (approved):
This solved the problem, thanks for the PR!

tests/backends/test_dbmodel.py
```diff
@@ -17,6 +17,8 @@
 
 should_run &= "tensorflow" in installed_redisai_backends()
 
+if not should_run:
```
ashao, line 20: Could be worth removing the rest of the skip checks after this, since they're made redundant by `allow_module_level=True`

ashao (approved):
One minor point, but not absolutely necessary to close this PR. Thanks @al-rigazzi for tracking this down.

tests/backends/test_dbmodel.py
al-rigazzi: Good catch, done!

al-rigazzi: Good catch, done!

PR #212, opened 2022-06-18
Fix ONNX test

al-rigazzi:
This PR aims at fixing the ONNX test, which has been erratically failing in CI/CD pipelines. 

I identified one possible source of errors in the fact that we overwrite the ONNX model several times, and - possibly - the model stored in the DB is not updated correctly (thus resulting in the error `Number of keys given as OUTPUTS here does not match model definition`). This hints at a bug in one of the used libraries (possibly RedisAI), thus we can not fix it completely: we can just avoid triggering it, by not overwriting models.

I also updated one function to use list of strings for inputs, instead of just a string (which is anyhow accepted by SmartRedis).

While I can not be sure that this fixes the problem, I have run the test 80 times in a row and it never failed.

### DIFF
--- a/tests/backends/run_sklearn_onnx.py
+++ b/tests/backends/run_sklearn_onnx.py
@@ -59,6 +59,7 @@ def run(device):
     linreg = build_lin_reg()
     outputs = run_model(client, device, linreg, X, "X", ["Y"])
     assert len(outputs[0]) == 5
+    print("Linear regression successful")
     print(outputs)
 
     # Kmeans test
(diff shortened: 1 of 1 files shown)


billschereriii (approved):
Thanks for this PR! The code is much cleaner now that we're not overwriting different models with the same name.  Is there any possibility that there is a concurrency issue here -- multiple copies of the test using the same backend database? That could account for some of them sometimes getting the wrong model if someone else has just overwritten it.  If so, we should look into getting Redis into a container with the tests so that there is no competition for the database.

Regardless, the code is much cleaner, with this change, so it is well worth getting this into the codebase.

PR #215, opened 2022-06-23
Update default versions

billschereriii:
Update default versions of SmartSim and SmartRedis for installation


al-rigazzi (changes requested):
I think version numbers occurring in `Makefile` and `docker-compose` can also be updated

al-rigazzi (approved):
LGTM!

PR #220, opened 2022-06-28
Update version_names.json

mellis13:
The version_names.json file needs to be updated with the new release so that future invocations of the github action ``deploy_dev_docs`` do not corrupt the doc version history.


al-rigazzi:
OK, I need more context here. Me and @billschereriii thought this had to be updated to current release versions... apologies if it blocked the docs process.

al-rigazzi (approved):
LGTM, thanks for catching this!

PR #225, opened 2022-07-07
Increase number of trials performed to get task ID from WLM

al-rigazzi:
On slow systems, starting executables in interactive session can take longer than what we hypothesized (>10 seconds). In this PR, we increase the number of trials SmartSim will make to get the task id from the WLM (from 5 to 10, each check is performed every 2 seconds), but also make this a configurable parameter, so that on slower systems, users can even make it larger.

On faster systems, this will not impact execution times, as one of the previous attempts will succeed.


smartsim/settings/containers.py
```diff
@@ -108,7 +108,10 @@ def _container_cmds(self, default_working_directory=''):
             working_directory = self.working_directory
 
         if not (working_directory in serialized_mount):
-            serialized_mount = ','.join([working_directory, serialized_mount])
+            if serialized_mount:
+                serialized_mount = ','.join([working_directory, serialized_mount])
+            else:
+                serialized_mount = working_directory
```
billschereriii, line 114: Is there a risk that this can overwrite multiple directories in the serialized mount with the single working directory? Would an append be better here?

al-rigazzi: So, the `else` path is taken only if `serialized_mount` is an empty string (or it is None, but this cannot happen in this portion of the code), thus I don't think we risk to overwrite it, if it is meaningful.
Here is why I made this change: `serialized_mount` here is a string, but can also be empty (and in our tests it is). If it is empty, we then concatenated `working_directory` and an empty string, separating them with a `,`. This resulted in the not-too-appealing command line `... --bind /your/working/dir, ...`. I was afraid that with shells, the trailing comma in the argument could cause problems (but maybe it is not the case).

PR #233, opened 2022-08-27
Update installation doc after SmartRedis documentation update

ashao:
https://github.com/CrayLabs/SmartRedis/pull/277 removes the `from_source.rst`
file because it was largely redundant. This small PR updates references to that
file with `lib.rst`


MattToast (approved):
LGTM!!

PR #234, opened 2022-09-21
Upgrade RedisAI to 1.2.7

ashao:
With the release of RedisAI 1.2.7, we can bump the supported versions of pyTorch, Tensorflow, and ONNX to more recent versions. Additionally, while updating the dependencies for RedisAI 1.2.7, the installation of the pyTorch backends with GPU support should now be more robust because we pin a more prescriptive version of `torch` and `torchvision`.

The new versions were tested on an internal HPE machine with on V100s.


doc/installation.rst
```diff
@@ -61,7 +61,7 @@ Supported Versions
 
 
 SmartSim supports multiple machine learning libraries through
-the use of RedisAI_. The following libraries are supported.
+the use of RedisAI_ 1.2.7. The following libraries are supported (based on .
```
billschereriii, line 64: Looks like you didn't finish a thought here

doc/installation.rst
```diff
@@ -73,15 +73,15 @@ the use of RedisAI_. The following libraries are supported.
...
      - Yes
    * - Tensorflow_ / Keras_
-     - 2.5.2
+     - 2.8.0
      - 3.7 - 3.9
      - Yes
-   * - ONNX_
-     - 1.9
+   * - ONNX_ Runtime
+     - 1.11.1
```
billschereriii, line 84: If these updates require RAI v1.2.7, we shouldn't claim that we still support older versions. There is some ambiguity here as to whether we're pushing everything forward or not.

billschereriii: run_tests.yml still tries to test with older versions of RedisAI

smartsim/_core/_cli/build.py
```diff
@@ -23,22 +23,29 @@
...
     end_point = None
-    # if we are on linux cpu, use the torch without CUDA
-    if sys.platform == "linux" and device == "cpu":
-        packages.append(f"torch=={versions.TORCH}+cpu")
-        packages.append(f"torchvision=={versions.TORCHVISION}+cpu")
+
+    if sys.platform == "darwin":
+        if device == "gpu":
+            logger.warning("GPU support is not available on Mac OS X")
+        device_suffix = ""
```
billschereriii, line 33: should this be `versions.TORCH_CPU_SUFFIX`?

smartsim/_core/_install/buildenv.py
```diff
@@ -126,6 +129,8 @@ class RedisAIVersion(Version_):
             "onnxmltools": "1.10.0",
             "scikit-learn": "1.0.2",
             "torch": "1.7.1",
+            "torch_cpu_suffix": "+cpu",
+            "torch_gpu_suffix": "+cu110",
```
billschereriii, line 133: If this is a specific (NVIDIA) GPU, please document that here

smartsim/_core/_install/buildenv.py
```diff
@@ -135,8 +140,21 @@ class RedisAIVersion(Version_):
             "onnxmltools": "1.10.0",
             "scikit-learn": "1.0.2",
             "torch": "1.9.1",
+            "torch_cpu_suffix": "+cpu",
+            "torch_gpu_suffix": "+cu111",
```
billschereriii, line 144: same as previous

smartsim/_core/_install/buildenv.py
```diff
@@ -135,8 +140,21 @@ class RedisAIVersion(Version_):
...
         },
+        "1.2.7": {
+            "tensorflow": "2.8.0",
+            "onnx": "1.11.0",
+            "skl2onnx": "1.11.1",
+            "onnxmltools": "1.11.1",
+            "scikit-learn": "1.1.1",
+            "torch": "1.11.0",
+            "torch_cpu_suffix": "+cpu",
+            "torch_gpu_suffix": "+cu113",
```
billschereriii, line 155: and here

billschereriii (changes requested):
Thanks for addressing this. Some feedback that I think needs to be addressed first. It's also unclear to me whether we're trying to support older versions of RedisAI or not -- some of the code does support older versions, but some seems to only support v1.2.7

doc/installation.rst
ashao: Thank you fixed.

doc/installation.rst
ashao: For now, we are keeping 1.2.3, 1.2.5, 1.2.7. I've updated the table to reflect the various versions we support

smartsim/_core/_install/buildenv.py
ashao: Good call. I have renamed this to `torch_cuda_suffix` to allow for the possibility for us to specify an alternate library (e.g. ROCm or another accelerator library)

smartsim/_core/_cli/build.py
ashao: No, since Mac OS X does not have any suffix associated with it. I added a comment to explicitly mention this.

billschereriii (approved):
Looks great! Thanks for taking care of this and cleaning up the code to get more production-ready

MattToast (approved):
Looks fantastic seems to build for me! LGTM!!

PR #237, opened 2022-10-04
Colocated model fix

mellis13:
A bug was reported that uninitialized variables in the ``colocated.py`` entry point were causing an error when utilizing ``add_ml_model()`` for PyTorch models.  This PR fixes that bug and adds tests for the ``add_ml_model()`` methods using PyTorch.


MattToast (approved):
LGTM!! Thanks for taking care of this!

PR #242, opened 2022-11-11
Refactor of OpenMPISettings and improved support for mpiexec

ashao:
mpiexec, mpirun, and orterun are all synonyms in the modern MPI
standards. The mpirunSettings and mpirunStep have been refactored
to the more generic mpiSettings and mpiStep. Within these modules
the classes associated with mpiexec, mpirun, and orterun all
inherit from mpiexec (which historically was the first executable
method that was defined in the MPI standard).

The one exception to this is that the slurm workload manager
provides a wrapper to mpiexec which wraps a subset of the
supported options to mpiexec so that they can be passed to Slurm.
In general, this is a very limited set of options and so we choose
not to support it. Instead, we check to see if mpiexec is actually
mpiexec.slurm and throw an UnsupportedError.

This PR incorporates developments from @rickybalin


smartsim/settings/mpiSettings.py
```diff
@@ -0,0 +1,344 @@
...
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import shutil
+import subprocess
+import re
```
MattToast, line 29: `re` unused

smartsim/settings/mpirunSettings.py
```diff
@@ -24,318 +24,14 @@
...
-            [self.run_command, "-V"], capture_output=True
-        )  # type: subprocess.CompletedProcess
-        version_statement = completed_process.stdout.decode()
-
-        if not re.match(r"orterun\s\(OpenRTE\)\s4.\d+.\d+", version_statement):
-            logger.warning("Non-OpenMPI implementation of `orterun` detected")
+simplefilter("once", DeprecationWarning)
+warn(
+    "mpirunSettings will be deprecated; use mpiSettings instead.",
+    DeprecationWarning
```
MattToast, line 36: add `stacklevel=2`?

tests/test_openmpi_settings.py
```diff
@@ -8,143 +8,49 @@
 
 import pytest
```
MattToast, line 10: This file looks nearly identical to `tests/test_mpi_settings.py`. Do we need them both?

tests/test_mpi_settings.py
```diff
@@ -0,0 +1,272 @@
...
+        ),
+        pytest.param(
+            MpiexecSettings,
+            osp.join("mpi_impl_stubs", "openmpi4"),
+            "mpiexec",
+            id="OpenMPI4-mpiexec",
+        ),
+        pytest.param(
+            OrterunSettings,
+            osp.join("mpi_impl_stubs", "openmpi4"),
```
MattToast, line 50: Should we add parameters to this test to make sure the stubs of intel's of mpiexec/mpirun/orterun do not throw a warning? If not I think we can remove those stubs

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -32,22 +32,34 @@
 from ....log import get_logger
 from .step import Step
 
-logger = get_logger(__name__)
+_supported_launchers = [
+    "PBS",
+    "COBALT",
+    "SLURM",
+    "LSB"
```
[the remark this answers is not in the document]
MattToast: probably worth running `make style` once we're done done reviewing. Looks like there are a lot of style changes that black is making (outside of this PR)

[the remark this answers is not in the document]
ashao: Will do!

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -58,62 +70,52 @@ def get_launch_cmd(self):
         :return: launch command
         :rtype: list[str]
         """
-        mpirun = self.run_settings.run_command
-        mpirun_cmd = [mpirun, "-wdir", self.cwd]
-        # add env vars to mpirun command
-        mpirun_cmd.extend(self.run_settings.format_env_vars())
+        mpi_cmd = [self._run_command, "--wdir", self.cwd]
+        # add env vars to mpi command
+        mpi_cmd.extend(self.run_settings.format_env_vars())
```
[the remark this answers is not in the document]
ashao: 🤦

smartsim/settings/mpirunSettings.py
ashao: At this point, you're allowed to yell at me, since I reneged on my promise to do this earlier

tests/test_mpi_settings.py
ashao: I suppose we could, but at this point I sort of see this as the complementary test of the slurm.mpiexec one. I'm more inclined to do as you suggest and remove the Intel ones since we probably only need to test one MPI implementation.

smartsim/_core/launcher/step/mpiStep.py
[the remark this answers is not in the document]
ashao: Actually, after shoving this through black, I'm actually more tempted to submit a second PR just for the style changes. That'll keep the git blame more informative if we're tracing history

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -34,10 +34,9 @@
 
 logger = get_logger(__name__)
 
-
-class MpirunStep(Step):
+class _BaseMPIStep(Step):
     def __init__(self, name, cwd, run_settings):
-        """Initialize a OpenMPI mpirun job step
+        """Initialize an mpiexec job step
```
MattToast, line 39: a base mpi job step?

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -46,8 +45,20 @@ def __init__(self, name, cwd, run_settings):
...
+        self._supported_launchers = [
+            "PBS",
+            "COBALT",
+            "SLURM",
+            "LSB"
+        ]
+
+
         self.run_settings = run_settings
+        self._run_command = self.run_settings._run_command
```
MattToast, line 60: Thoughts on making this a property?
```py
@property
def run_command(self):  # or _run_command if we prefer it be a "protected" attr
   return self.run_settings._run_command
```
That way its still "un-writable" like the like the protected attr, but that way if the run command were to be changed, we wouldn't have to update it in two locations (`_BaseMPIStep._run_command` and `_BaseMPIStep.run_settings._run_command`)

Slightly on the fence about this suggestion bc I cannot think of why the run command of a step would ever change, but I figured its better to protect it early

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -46,8 +45,20 @@ def __init__(self, name, cwd, run_settings):
         :param run_settings: run settings for entity
         :type run_settings: RunSettings
         """
+
         super().__init__(name, cwd)
+
+        self._supported_launchers = [
```
MattToast, line 51: Can this be a class var so the one list is shared across all instances of the class (assuming that this list is read only)?

smartsim/settings/mpiSettings.py
```diff
@@ -0,0 +1,350 @@
...
+            **kwargs,
+        )
+        self.mpmd = []
+
+        if not shutil.which(self._run_command):
+            msg = (
+                f"Cannot find {self._run_command}. Try passing the "
+                "full path via run_command."
+            )
+            if fail_if_missing_exec:
```
MattToast, line 81: Really like this!! Allows the user to decide how fast to err out

MattToast (commented):
LGTM!!

Flagged a couple of small things, but nothing to withhold approval over. Lmk what you think.

The only reason I'm not explicitly approving is to give Al a chance to look over/approve himself.

smartsim/_core/launcher/step/mpiStep.py
ashao: I think that should be ok, I'm going to meditate on this overnight, but will probably take your suggestion

smartsim/_core/launcher/step/mpiStep.py
ashao: Good call

smartsim/_core/launcher/step/mpiStep.py
ashao: Yep, after my morning coffee; I'm more onboard. Might be a little overly defensive, but better than someone accidentally stumbling their way to this.

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -46,8 +45,22 @@ def __init__(self, name, cwd, run_settings):
         :param run_settings: run settings for entity
         :type run_settings: RunSettings
         """
+
         super().__init__(name, cwd)
+
+        _supported_launchers = [
```
MattToast, line 51: too far indented, is local var

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -46,8 +45,22 @@ def __init__(self, name, cwd, run_settings):
...
+            "PBS",
+            "COBALT",
+            "SLURM",
+            "LSB"
+        ]
+
         self.run_settings = run_settings
+
+        @property
+        def _run_command(self):
```
MattToast, line 61: too far indented, is local func

MattToast (changes requested):
Couple of syntax things i noticed during demos

smartsim/_core/launcher/step/mpiStep.py
ashao: I'm so embarassed

smartsim/_core/launcher/step/mpiStep.py
ashao: I may die of shame

MattToast (approved):
LGTM!! Approving as per yesterday's stand-up

PR #245, opened 2022-12-06
Model as Batch Job

MattToast:
Allows users to launch individual models through an experiment as a batch job with batch settings.

```py
exp = Experiment(...)
rs = exp.craete_run_settings(...)
bs = exp.create_batch_settings(...)
model = exp.create_model("MyModel", rs, batch_settings=bs)
exp.start(model)  # launches the model as a batch job
```

If a model with batch settings is added to a higher order entity (e.g. an ensemble), the batch settings of the higher order entity will be used and the batch settings of the model will be ignored to avoid an "sbatch-in-sbatch" scenario or similar .


al-rigazzi (commented):
Looks very good! The only question I have is whether the changes were tested on a Slurm and a PBS system, as the batch mechanism will not be used on `Local`.

MattToast:
@al-rigazzi Good Question! I implemented against a slurm system, and I just got a chance to run the new `full_wlm` test on a PBS system, and both cases the test is passing. Between that an a couple of dummy runs through I did in the repl against both wlms I fairly confident everything is in working order!!

al-rigazzi (commented):
LGTM -- but let's wait on other reviews.

ashao (approved):
Looks good to me. Thanks for doing this work!

PR #246, opened 2022-12-16
Domain Socket Support for co-located databases

ashao:
`Models` can now connect to a co-located database using a
Unix Domain Socket (UDS). In synthetic benchmarks, this can
lead to significant improvements over using the loopback
interface over TCP/IP. To ensure compatibility with previous
scripts we now provide the following interfaces:

- `colocate_db` (original): Will throw a `DeprecationWarning`
  but otherwise just wraps `colocate_db_tcp`
- `colocate_db_tcp`: Listens for connections over the loopback
- `colocate_db_uds`: Listens for connections over UDS

### DIFF
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -462,8 +462,20 @@ class Controller:
         # Set address to local if it's a colocated model
         if hasattr(entity, "colocated"):
             if entity.colocated:
-                port = entity.run_settings.colocated_db_settings["port"]
-                client_env["SSDB"] = f"127.0.0.1:{str(port)}"
+                port = entity.run_settings.colocated_db_settings.get("port",None)
+                socket = entity.run_settings.colocated_db_settings.get("unix_socket",None)
+                if socket and port:
+                    raise SSInternalError(
+                        "Co-located was configured for both TCP/IP and UDS"
+                    )
+                if port:
+                    client_env["SSDB"] = f"127.0.0.1:{str(port)}"
+                elif socket:
+                    client_env["SSDB"] = f"unix://{socket}"
+                else:
+                    raise SSInternalError(
+                        "Colocated database was not configured for either TCP or UDS"
+                    )
         entity.run_settings.update_env(client_env)
 
     def _save_orchestrator(self, orchestrator):
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -460,8 +460,7 @@ class Controller:
                 client_env["SSKEYOUT"] = entity.name
 
         # Set address to local if it's a colocated model
-        if hasattr(entity, "colocated"):
-            if entity.colocated:
+        if hasattr(entity, "colocated") and entity.colocated:
                 port = entity.run_settings.colocated_db_settings.get("port",None)
                 socket = entity.run_settings.colocated_db_settings.get("unix_socket",None)
                 if socket and port:
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -45,7 +45,7 @@ def write_colocated_launch_script(file_name, db_log, colocated_settings):
     :type colocated_settings: dict[str, Any]
     """
 
-    colocated_cmd = _build_colocated_wrapper_cmd(**colocated_settings, db_log=db_log)
+    colocated_cmd = _build_colocated_wrapper_cmd(db_log, **colocated_settings)
 
     with open(file_name, "w") as f:
         f.write("#!/bin/bash\n")
@@ -75,10 +75,12 @@ def write_colocated_launch_script(file_name, db_log, colocated_settings):
 
 
 def _build_colocated_wrapper_cmd(
+    db_log,
     cpus=1,
     rai_args=None,
     extra_db_args=None,
-    db_log=None,
+    port=6780,
+    ifname=None,
     **kwargs,
 ):
     """Build the command use to run a colocated db application
@@ -113,9 +115,8 @@ def _build_colocated_wrapper_cmd(
         str(cpus),
     ]
     # Add in the interface if using TCP/IP
-    interface = kwargs.get("ifname",None)
-    if interface:
-        cmd.extend(["+ifname", interface])
+    if ifname:
+        cmd.extend(["+ifname", ifname])
     cmd.append("+command")
     # collect DB binaries and libraries from the config
     db_cmd = [CONFIG.database_exe, CONFIG.database_conf, "--loadmodule", CONFIG.redisai]
@@ -127,7 +128,7 @@ def _build_colocated_wrapper_cmd(
             # ex. THREADS_PER_QUEUE=1
             db_cmd.append(f"{arg.upper()} {str(value)}")
 
-    db_cmd.extend(["--port", str(kwargs["port"])])
+    db_cmd.extend(["--port", str(port)])
 
     # Add socket and permissions for UDS
     unix_socket = kwargs.get("unix_socket", None)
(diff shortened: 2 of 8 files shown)


MattToast (changes requested):
Looks good!! I found a couple of small things here and there that need changing so I figured I'd throw up some quick feed back right away.

I'm still not fully convinced deprecating `colocate_db` is needed, but I do not have a better solution off hand. I'm going to think this over and see if I can't come back with a more robust suggestion.

smartsim/_core/control/controller.py
```diff
@@ -462,8 +462,20 @@ def _prep_entity_client_env(self, entity):
         # Set address to local if it's a colocated model
         if hasattr(entity, "colocated"):
             if entity.colocated:
```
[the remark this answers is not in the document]
ashao: I don't think this is possible right because it will fail when trying to evaluate `entity.colocated` if that attribute doesn't exist?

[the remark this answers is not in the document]
MattToast: Good question!!

Should totally be okay change to make. Boolean expressions will short circuit in python. A little proof of concept can be seen bellow


```py
def first_check():
    print("Doing First Check!")
    return False

def second_check():
    print("Doing Second Check")
    raise Exception(
        "This would be the equivilent of ``entity.colocated`` "
        "raising an AttributeError"
    )

if first_check() and second_check():
    #                ^^^^^^^^^^^^^^
    #                Because the first clause returns ``False``
    #                there is no way that result of this function
    #                can change the whole condition from evaluating
    #                to ``False``, so the python interpreter will
    #                save time by simply not running it
    #                (most c-style inspired languages, will do a
    #                similar optimization)
    print("In conditional == TRUE eval block")
else:
    print("In conditional == FALSE eval block")
```

Which prints out exactly what we would expect if we assume a short circuit
```txt
Doing First Check!
In conditional == FALSE eval block
```
Like I mentioned previously, this is really more of a style complaint than anything else (we are using two level of indentation where one will do) so feel free to ignore if you think this change is "out-of-scope".

smartsim/_core/launcher/colocated.py
```diff
@@ -113,26 +107,45 @@ def _build_colocated_wrapper_cmd(
...
             # ex. THREADS_PER_QUEUE=1
             db_cmd.append(f"{arg.upper()} {str(value)}")
 
-    # add port and log information
-    db_cmd.extend(["--port", str(port), "--logfile", db_log])  # usually /dev/null
+    db_cmd.extend(["--port", str(kwargs["port"])])
+
+    # Add socket and permissions for UDS
+    unix_socket = kwargs.get("unix_socket", None)
+    socket_permissions = kwargs.get("socket_permissions", None)
```
MattToast, line 134: Any reason we are reaching into kwargs here? Could these be made named kwargs with default value `None`?

smartsim/_core/launcher/colocated.py
```diff
@@ -113,26 +107,45 @@ def _build_colocated_wrapper_cmd(
...
+                "--unixsocket", str(unix_socket),
+                "--unixsocketperm", str(socket_permissions)
+            ]
+        )
+    elif bool(unix_socket) ^ bool(socket_permissions):
+        raise SSInternalError(
+            "`unix_socket` and `socket_permissions` must both be defined or undefined."
+            )
+
+    db_cmd.extend(["--logfile", db_log])  # usually /dev/null
```
MattToast, line 148: Should `"/dev/null"` be the default value? Would I ever want to be able to extend this list with `["--logfile", None]`?

Maybe guard with a null check?
```py
if db_log is not None:
    db_cmd.extend(...)
```

smartsim/entity/model.py
```diff
@@ -128,21 +129,88 @@ def attach_generator_files(self, to_copy=None, to_symlink=None, to_configure=Non
...
+        with this colocated database by using Unix Domain sockets.
+
+        Extra parameters for the db can be passed through kwargs. This includes
+        many performance, caching and inference settings.
+
+        .. highlight:: python
+        .. code-block:: python
+
+            ex. kwargs = {
+                maxclients: 100000,
```
MattToast, line 164: Can we change the dict name and make the keys strings to make the code block valid python? (e.g. `ex. kwargs -> ex_kwargs` and `maxclients -> "maxclients"`)

smartsim/entity/model.py
```diff
@@ -128,21 +129,88 @@ def attach_generator_files(self, to_copy=None, to_symlink=None, to_configure=Non
...
+        :type socket_permissions: int, optional
+        :param db_cpus: number of cpus to use for orchestrator, defaults to 1
+        :type db_cpus: int, optional
+        :param limit_app_cpus: whether to limit the number of cpus used by the app, defaults to True
+        :type limit_app_cpus: bool, optional
+        :param debug: launch Model with extra debug information about the co-located db
+        :type debug: bool, optional
+        :param kwargs: additional keyword arguments to pass to the orchestrator database
+        :type kwargs: dict, optional
+
```
MattToast, line 185: extra new line

smartsim/entity/model.py
```diff
@@ -162,39 +230,53 @@ def colocate_db(
...
-                "intra_op_parallelism": kwargs.get("intra_op_parallelism", None),
-            },
+        colo_db_config = {}
+        colo_db_config.update(connection_options)
+        colo_db_config.update(common_options)
+        # redisai arguments for inference settings
+        colo_db_config['rai_args'] = {
+            "threads_per_queue": kwargs.get("threads_per_queue", None),
+            "inter_op_parallelism": kwargs.get("inter_op_parallelism", None),
+            "intra_op_parallelism": kwargs.get("intra_op_parallelism", None),
```
MattToast, line 279: I know this was here prior to this PR, so feel free to ignore this comment, but is there any reason these cannot become named kwargs?

MattToast (changes requested):
Looks great!! I just had a few questions regarding about some unnamed but always retrieved kwargs and couple of small style requests, but otherwise looks about ready to go!!

smartsim/_core/launcher/colocated.py
```diff
@@ -113,26 +107,45 @@ def _build_colocated_wrapper_cmd(
...
     # add extra redisAI configurations
     for arg, value in rai_args.items():
         if value:
             # RAI wants arguments for inference in all caps
             # ex. THREADS_PER_QUEUE=1
             db_cmd.append(f"{arg.upper()} {str(value)}")
 
-    # add port and log information
-    db_cmd.extend(["--port", str(port), "--logfile", db_log])  # usually /dev/null
+    db_cmd.extend(["--port", str(kwargs["port"])])
```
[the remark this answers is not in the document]
ashao: Case should always be specified in both the TCP/IP and UDS cases (the latter being a little dumb but recommended by Redis). This was a realization that happened during the testing, so I'll revert the change and make it a named kwarg again

smartsim/_core/launcher/colocated.py
ashao: I'm actually not sure why this is a kwarg and not a required argument. If you look at the (one) reference to it; it is always specified. I'll do a quick refactor and check to make sure it doesn't break anything if I change it to an arg.

smartsim/entity/model.py
ashao: How dare. Valid python? You think we're professionals over here?!

smartsim/entity/model.py
ashao: I think it might be just a strictly organiztional thing? I'm a little loathe to start pulling in this thread since since I'd rather we do a refactor of the entire handling of the **kwargs as opposed to just focusing on this one instance.

smartsim/_core/control/controller.py
```diff
@@ -462,8 +462,20 @@ def _prep_entity_client_env(self, entity):
...
-                port = entity.run_settings.colocated_db_settings["port"]
-                client_env["SSDB"] = f"127.0.0.1:{str(port)}"
+                port = entity.run_settings.colocated_db_settings.get("port",None)
+                socket = entity.run_settings.colocated_db_settings.get("unix_socket",None)
```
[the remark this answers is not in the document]
ashao: I'll do a `make style` in the followup PR because it will touch a LOT of files and I'd rather the changes here be fairly encapsulated.

smartsim/entity/model.py
MattToast: Very fair! I could easily see this being the subject of a separate PR. Consider this resolved!!

MattToast (approved):
Changes look great! LGMT, pending tests!!

PR #247, opened 2022-12-21
Set number of tasks for PALS mpiexec

rickybalin:
Added an explicit call to set the number of tasks for the PALS mpiexec settings. This now sets the number of tasks with "--np", whereas before it was being set with "--n", which was giving and error on Polaris.


MattToast (approved):
LGTM! Thanks for catching and fixing this!!

PR #252, opened 2023-01-28
Update style to conform with black

ashao:
This PR makes no substantive changes, but ensures that the codebase
follows `black` style.


MattToast (approved):
LGTM!! Thanks for doing this, I think it's nice to have everything once again conform to pep8!

PR #253, opened 2023-02-02
Fix Setup Script Syntax

MattToast:
Address bad `setup.py` syntax no longer accepted by `setuptools>=67.0.0`


ashao (approved):
Thanks for tracking this down.

Spartee:
aight come on it wasn't that bad.

MattToast:
omg the original smartsim bdfl returns!!

PR #255, opened 2023-02-08
Better Exit Usage

MattToast:
Removes use of `exit` and replaces with `sys.exit` to handle the unlikely case where users (foolishly) try to use the `smart` cli without fully initializing `site` module


al-rigazzi (approved):
LGTM, good catch, thanks for the fast fix

PR #259, opened 2023-02-21
Fix Mpiexec Test on Slurm Based Machines

MattToast:
Adds new test for testing mpiexec settings that is able to handle slurm's `mpiexec.slurm` wrapper.

### DIFF
--- a/tests/test_run_settings.py
+++ b/tests/test_run_settings.py
@@ -1,6 +1,7 @@
 import contextlib
 import logging
 from shutil import which
+import subprocess
 
 import pytest
 
@@ -12,7 +13,6 @@ from smartsim.settings import (
 )
 from smartsim.settings.settings import create_run_settings
 from smartsim.error.errors import SSUnsupportedError
-from smartsim.wlm import detect_launcher
 
 
 def test_create_run_settings_local():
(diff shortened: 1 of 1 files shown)


tests/test_run_settings.py
```diff
@@ -47,7 +49,23 @@ def test_create_run_settings_returns_settings_subclasses(run_cmd, Settings):
...
         assert settings.run_command == _run_cmd
-        assert type(settings) == Settings
+        assert isinstance(settings, Settings)
+
+
+def test_create_run_settings_handles_mpiexec_settings_correctly():
+    with contextlib.ExitStack() as ctx:
+        # naively assume that if slurm is present, then the mpiexec.slurm
+        # wrapper is being used
+        if detect_launcher() == "slurm":
```
ashao, line 59: So here's where this gets REALLY strange (and why there's that stupid help-string parsed check in the `MpiexecSettings` source code. A machine might have Slurm, but ALSO have an actual MPI library that provides a 'real' version of mpiexec. In that circumstance, this test would incorrectly fail. I think you're going to have to do a query here as well on the help strings for mpiexec to see if the `mpiexec` in the path is the Slurm one or not.

ashao (changes requested):
One tweak to the Slurm check because Slurm is 'cool and normal' with how it handles `mpiexec` /s

tests/test_run_settings.py
MattToast: I was afraid of that...

It's unfortunate that this test may need to dissolve into making the same check in both the application and test code, but at this point in time I don't really see a way around it.

ashao (approved):
Great thanks Matt!

PR #260, opened 2023-02-21
Update copyright headers

MattToast:
Update all existing copyrights headers to use `2021-2023`, add copyright headers to all application and testing files that were previously missing them.


tests/test_local_restart.py
```diff
@@ -1,10 +1,38 @@
...
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 
 """
 Test restarting ensembles and models.
 """
 
 
+from smartsim import Experiment, status
```
billschereriii, line 33: I would have expected this to maintain relative position compared to the file comment above

billschereriii (approved):
Thank you for a very comprehensive update! This ended up being far more work than anyone would have guessed as the copyrights were not in as good a shape as we would have imagined.

The .github/workflows YAML files are still missing copyrights, but everything else looks really great. Please address that before merging.

tests/test_local_restart.py
MattToast: whoops, this was actually a change intended for a different branch that I must put on this one accidentally! Consider it revereted

PR #266, opened 2023-03-01
Fix Unreachable Block on Colo Lsf Settings

MattToast:
An object of type `dict[str, Any] | None` is never equivalent to `True`. This needs to be cast to a bool or make a container-is-not-empty check. This PR does the later.


ashao (approved):
Good catch!

PR #268, opened 2023-03-08
exclude csv-valued env values from --export, enable csv-valued in mpmd

ankona:
This PR fixes two defects related to passing environment variables to slurm jobs.

1. CSV-valued environment variables are not passed to MPMD executions
2. CSV-valued environment variables are passed twice

For (1), add populating source env with `env k1=v1,v2 {srun} --export k1`

For (2), perform key-only export of csv-valued vars:
- old: `env k1=v1,v2 srun --export k1=v1,v2`
- new: `env k1=v1,v2 srun --export k1`

Sample demonstrating "double export" overwrite & solution in non-mpmd:

```
mcb@h:~> env abc=123,456 def=234,345 srun --output /home/users/mcbridch/outsolo.txt --export=abc=123,456 /usr/bin/printenv
mcb@h:~> cat ./outsolo.txt | grep abc                                                                                    
abc=123
mcb@h:~> env abc=123,456 def=234,345 srun --output /home/users/mcbridch/outsolo.txt --export=abc /usr/bin/printenv 
mcb@h:~> cat ./outsolo.txt | grep abc                                                                             
abc=123,456
```

Sample mpmd launch_cmd with no env:

image

Sample mpmd launch_cmd with env:

image

### DIFF
--- a/smartsim/_core/launcher/step/slurmStep.py
+++ b/smartsim/_core/launcher/step/slurmStep.py
@@ -141,7 +141,7 @@ class SrunStep(Step):
             ) = self.run_settings.format_comma_sep_env_vars()
 
             if len(env_var_str) > 0:
-                srun_cmd += ["--export", env_var_str]
+                srun_cmd += ["--export", f"ALL,{env_var_str}"]
 
             if comma_separated_env_vars:
                 compound_env = compound_env.union(comma_separated_env_vars)
@@ -213,7 +213,7 @@ class SrunStep(Step):
 
             (env_var_str, csv_env_vars) = mpmd.format_comma_sep_env_vars()
             if len(env_var_str) > 0:
-                cmd += ["--export", env_var_str]
+                cmd += ["--export", f"ALL,{env_var_str}"]
             if csv_env_vars:
                 compound_env_vars.extend(csv_env_vars)
             cmd += mpmd.exe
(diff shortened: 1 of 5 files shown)


al-rigazzi (approved):
Good catch! LGTM

ashao (changes requested):
Can you confirm that passing just the name of a variable (e.g. just `foo` with no value) leads to an export argument that looks like
```
--export foo,k=1
```

ankona:
> Can you confirm that passing just the name of a variable (e.g. just `foo` with no value) leads to an export argument that looks like
> 
> ```
> --export foo,k=1
> ```

I thought this could be a fairly simple addition. Unfortunately, it seems like it's not a currently supported feature and would require me to make more changes than expected. 

For example, check out this [type checking](https://github.com/CrayLabs/SmartSim/blob/develop/smartsim/settings/base.py#L383-L386). This bit of code would make it impossible to add variables without values without changing & potentially weakening this feature.

Another example is the [default formatting behavior](https://github.com/CrayLabs/SmartSim/blob/develop/smartsim/settings/base.py#L520-L526). It assumes a key-value pair and would need to be updated.

I added a new [task ](https://github.com/CrayLabs/SmartSim-EE/issues/301) to review this. I'd like to solve this, but not in the existing task

al-rigazzi (changes requested):
As discussed, let's update comments and add the `ALL` keyword to `env_var_str` (I think this could happen in the format function, but I leave it up to you). Sorry for the misunderstanding!

al-rigazzi (approved):
LGTM, thanks for the last-minute updates!

ashao (approved):
Looks great! Thanks for making those changes and having such complete test coverage.

PR #269, opened 2023-03-15
Relax codecov threshold

ashao:
The updated codecov integration will occasionally flag PRs as failing due to decreases in code coverage. Investigating the logs points to this being sporadic changes due to occasional failures when uploading code coverage. To avoid false positives and a stopgap solution, the threshold for code coverage is set to 0.5% instead of the default 0%


ankona (approved):
Great catch!

PR #271, opened 2023-03-31
Fix Type Hint in Doc for `create_run_settings`

MattToast:
The factory method `Experiment.create_run_setttings` lists the `run_args` param to be a `list[str]`, but every RS constructor expects a `dict[str, str]` without any casting being done behind the scenes.


ashao (approved):
Thanks for fixing it @MattToast and thanks for finding it @amandarichardsonn

PR #282, opened 2023-05-03
Adding affinity scripts to PALS mpiexec settings

rickybalin:
Added an option to specify an affinity script for the PALS mpiexec settings and add it to the run command after the run arguments, with no "--" prefix. This feature is needed when running multiple applications on the same nodes with the co-located deployment and the applications have to share the GPU on each node.


rickybalin:
I was wondering if there were any updates on this pull request. The ability to specify a GPU affinity script to the run settings is very useful when sharing the GPU on each node between multiple applications.

smartsim/settings/palsSettings.py
```diff
@@ -185,6 +185,16 @@ def set_walltime(self, walltime):
         """
         logger.warning("set_walltime not supported under PALS")
 
+    def set_gpu_affinity_script(self, affinity, *args):
```
al-rigazzi, line 188: So, here the idea would be to let the user add more than one file?
Also, I think you will need to add typehints, so possibly the declaration will be:
```python
def set_gpu_affinity_script(self, affinity: str, *args: str) -> None:

al-rigazzi (changes requested):
Thanks for this contribution, @rickybalin! A couple of minor things to address before we can merge:
1. Can you resolve merge conflicts and address my comment?
2. Then you should be able to run `make check-mypy` and `make check-lint` to make sure all tests will pass
3. May I ask you to add a test to check that the affinity parameter(s) are correctly added to this kind of Run Settings?

Thanks (and sorry for catching up so late)!

al-rigazzi (commented):
Everything looks good! If you could just add a test for the new feature, I think this is ready to go!

rickybalin:
> Everything looks good! If you could just add a test for the new feature, I think this is ready to go!

Done.

PR #290, opened 2023-05-25
Update doc strings

mellis13:
Per group discussion, updating doc strings to the correct form of colocated.


tutorials/ml_inference/Inference-in-SmartSim.ipynb
```diff
@@ -811,20 +824,21 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "id": "5daa7402-f62b-4710-a269-e078b2ce08ac",
    "metadata": {},
    "source": [
-    "# Co-Located Deployment\n",
+    "# CoLocated Deployment\n",
```
MattToast, line 832: Should we also make this "Colocated" (lower "l") just to be consistent with the rest of the titles/capitalizations of colocated?

tutorials/ml_inference/Inference-in-SmartSim.ipynb
```diff
@@ -811,20 +824,21 @@
...
-    "A co-located Orchestrator is a special type of Orchestrator that is deployed\n",
+    "A colocated Orchestrator is a special type of Orchestrator that is deployed\n",
     "on the same compute hosts an a Model instance defined by the user. In this\n",
     "deployment, the database is not connected together in a cluster and each shard\n",
     "of the database is addressed individually by the processes running on that compute\n",
     "host. This is particularly important for GPU-intensive workloads which require\n",
     "frequent communication with the database.\n",
     "\n",
-    "<img src=\"https://www.craylabs.org/docs/_images/co-located-orc-diagram.png\" alt=\"lattice\" width=\"600\"/>\n"
+    "<img src=\"https://www.craylabs.org/docs/_images/colocated-orc-diagram.png\" alt=\"lattice\" width=\"600\"/>\n"
```
MattToast, line 841: Just confirming that this link does correspond with the `colocated-orc-diagram.png` whose name was changed in this pr. Do we care that this may not render if a user were to pull this notebook from `develop` as opposed to from a release?

MattToast (approved):
Two quick questions regarding the jupyter changes, but otherwise lgtm!

tutorials/ml_inference/Inference-in-SmartSim.ipynb
mellis13: Actually, I think I'm in favor of just not updating it at all.  It's an internal filename and if we change it previous versioned releases will be broken because once we publish this to the main docs the file won't exist anymore.  Do you agree with that?  It's an edge case but I guess not much is gained by fixing an internal file name.

MattToast: I think that seems like a perfectly reasonable justification to just leave the current filename! Like you said, its strictly internal, so I'm totally fine with letting this be one place where the "co-loctated" spelling still exists

PR #291, opened 2023-05-25
Fix test suite failure when no tf wheel

MattToast:
Resolves test module `test_dbmodel.py` using symbols from a TF import statement that may not be present at test time (e.g. if a user installs w/o ml backends and then runs `make test-full`)


mellis13 (approved):
LGTM -- thanks for fixing this

mellis13 (approved):
LGTM -- thanks for fixing this

mellis13:
Just a quick question on this -- is it worth noting in the change log that we fixed this bug?

MattToast:
> Just a quick question on this -- is it worth noting in the change log that we fixed this bug?

Certainly couldn't hurt in case anyone user was running into this issue. I'll throw in a quick mention

PR #292, opened 2023-05-30
Migrate from redis-py-cluster to redis-py.

mellis13:
This PR migrates SmartSim from using ``redis-py-cluster`` to ``redis-py`` to check cluster status.  ``redis-py-cluster`` is no longer maintained and ``redis-py`` has picked up the cluster functionality in ``redis-py-cluster``.


ashao (approved):
Thanks for getting to this! Looks good to me

MattToast (approved):
LGTM!!

smartsim/_core/utils/redis.py
```diff
@@ -90,20 +90,20 @@ def check_cluster_status(hosts, ports, trials=10):  # cov-wlm
...
     """
-    host_list = []
+    cluster_nodes = []
     for host in hosts:
         for port in ports:
-            host_dict = dict()
-            host_dict["host"] = get_ip_from_host(host)
-            host_dict["port"] = port
-            host_list.append(host_dict)
+            cluster_nodes.append(ClusterNode(get_ip_from_host(host), port))
```
MattToast, line 96: super minor style thing: If you want to make this a hair more declarative, you could do something like:
```py
cluster_nodes = [ClusterNode(get_ip_from_host(host), port)
                 host, port in itertools.product(hosts, ports)]
```

mellis13: Good idea -- fixed.

MattToast: One last thing before merge: gotta import `itertools` or this will raise an undefined symbol error at runtime

mellis13: Great catch.  I just realized that our existing code isn't covered by the CI/CD or local tests because we only launch clusters on nightly WLM tests.  I can't think of an easy way around it though given the constraints of the ``Orchestrator`` constructor.  I'm going to post an issue to fix this in the future... we should really catch broken code around clusters before they get to ``develop``.

PR #293, opened 2023-06-01
Update redis conf file

mellis13:
This PR changes the redis conf file name from ``redis6.conf`` to ``redis.conf`` to align with the upgrade from redis v6 to redis v7.  Also, the contents of the redis file were merged with the updated redis conf file associated with redis v7.    It is worth noting that the following entries in the redis conf file are new with redis v7 (see below).  The defaults were taken for these new entries.

```bash
# By default, Redis modifies the process title (as seen in 'top' and 'ps') to
# provide some runtime information. It is possible to disable this and leave
# the process name as executed by setting the following to no.
set-proc-title yes

# When changing the process title, Redis uses the following template to construct
# the modified title.
#
# Template variables are specified in curly brackets. The following variables are
# supported:
#
# {title}           Name of process as executed if parent, or type of child process.
# {listen-addr}     Bind address or '*' followed by TCP or TLS port listening on, or
#                   Unix socket if only that's available.
# {server-mode}     Special mode, i.e. "[sentinel]" or "[cluster]".
# {port}            TCP port listening on, or 0.
# {tls-port}        TLS port listening on, or 0.
# {unixsocket}      Unix domain socket listening on, or "".
# {config-file}     Name of configuration file used.
#
proc-title-template "{title} {listen-addr} {server-mode}"



# When diskless replication is enabled with a delay, it is possible to let
# the replication start before the maximum delay is reached if the maximum
# number of replicas expected have connected. Default of 0 means that the
# maximum is not defined and Redis will wait the full delay.
repl-diskless-sync-max-replicas 0


# FLUSHDB, FLUSHALL, SCRIPT FLUSH and FUNCTION FLUSH support both asynchronous and synchronous
# deletion, which can be controlled by passing the [SYNC|ASYNC] flags into the
# commands. When neither flag is passed, this directive will be used to determine
# if the data should be deleted asynchronously.

lazyfree-lazy-user-flush no


# For convenience, Redis stores all persistent append-only files in a dedicated
# directory. The name of the directory is determined by the appenddirname
# configuration parameter.

appenddirname "appendonlydir"


# Redis supports recording timestamp annotations in the AOF to support restoring
# the data from a specific point-in-time. However, using this capability changes
# the AOF format in a way that may not be compatible with existing AOF parsers.
aof-timestamp-enabled no

```

### DIFF
--- a/smartsim/_core/config/redis.conf
+++ b/smartsim/_core/config/redis.conf
@@ -32,8 +32,17 @@
 # If instead you are interested in using includes to override configuration
 # options, it is better to use include as the last line.
 #
+# Included paths may contain wildcards. All files matching the wildcards will
+# be included in alphabetical order.
+# Note that if an include path contains a wildcards but no files match it when
+# the server is started, the include statement will be ignored and no error will
+# be emitted.  It is safe, therefore, to include wildcard files from empty
+# directories.
+#
 # include /path/to/local.conf
 # include /path/to/other.conf
+# include /path/to/fragments/*.conf
+#
 
 ################################## MODULES #####################################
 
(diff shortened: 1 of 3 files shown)


smartsim/_core/config/redis.conf
```diff
@@ -49,43 +58,81 @@
...
+# When protected mode is on and the default user has no password, the server
+# only accepts local connections from the IPv4 address (127.0.0.1), IPv6 address
+# (::1) or Unix domain sockets.
 #
 # By default protected mode is enabled. You should disable it only if
 # you are sure you want clients from other hosts to connect to Redis
-# even if no authentication is configured, nor a specific set of interfaces
-# are explicitly listed using the "bind" directive.
+# even if no authentication is configured.
 protected-mode no
```
billschereriii, line 111: We may want to revisit this decision some day. Maybe make a spike to investigate it?

smartsim/_core/config/redis.conf
```diff
@@ -105,7 +152,7 @@ tcp-backlog 4096
 # incoming connections. There is no default, so Redis will not listen
 # on a unix socket when not specified.
 #
-# unixsocket /tmp/redis.sock
+# unixsocket /run/redis.sock
```
billschereriii, line 155: Do we use UDS from SmartSim? Might want to align the file here with what we use if so

smartsim/_core/config/redis.conf
```diff
@@ -269,44 +363,74 @@ logfile ""
...
-#   Will save the DB if both the given number of seconds and the given
-#   number of write operations against the DB occurred.
+# Snapshotting can be completely disabled with a single empty string argument
+# as in following example:
 #
-#   In the example below the behavior will be to save:
-#   after 900 sec (15 min) if at least 1 key changed
-#   after 300 sec (5 min) if at least 10 keys changed
-#   after 60 sec if at least 10000 keys changed
+# save ""
```
billschereriii, line 424: I thought we wanted snapshotting disabled by default, so this line should be uncommented

smartsim/_core/config/redis.conf
```diff
@@ -1343,8 +1706,54 @@ cluster-require-full-coverage yes
...
+# cluster-link-sendbuf-limit 0
+
+# Clusters can configure their announced hostname using this config. This is a common use case for
+# applications that need to use TLS Server Name Indication (SNI) or dealing with DNS based
+# routing. By default this value is only shown as additional metadata in the CLUSTER SLOTS
+# command, but can be changed using 'cluster-preferred-endpoint-type' config. This value is
+# communicated along the clusterbus to all nodes, setting it to an empty string will remove
+# the hostname and also propagate the removal.
+#
+# cluster-announce-hostname ""
```
billschereriii, line 1736: We may want to use this for our eventual multiDB support

billschereriii (approved):
Generally looks very good. A few comments on Redis features that we may want to follow up on in subsequent tickets -- please open for any that you agree merit investigation
Thanks for taking care of this one!

smartsim/_core/config/redis.conf
mellis13: done

smartsim/_core/config/redis.conf
mellis13: great catch -- fixed

smartsim/_core/config/redis.conf
mellis13: Good idea -- done

PR #295, opened 2023-06-01
Typehints

ankona:


pyproject.toml
```diff
@@ -65,3 +65,41 @@ ignore_errors = true
 
 [tool.coverage.html]
 directory = "htmlcov"
+
+[tool.mypy]
+python_version = "3.9"
```
billschereriii, line 70: We actually support a larger range. Need to confirm that the type hints work as expected with other versions of Python, especially the older ones

pyproject.toml
```diff
@@ -65,3 +65,41 @@ ignore_errors = true
...
+# Ignore packages that are not used or not typed
+module = [
+  "coloredlogs",
+  "smartredis",
+  "smartredis.error"
+]
+ignore_missing_imports = true
+ignore_errors = true
+
+[[tool.mypy.overrides]]
```
billschereriii, line 88: Does it cause problems to have two tool.mypy.overrides sections?

requirements-dev.txt
```diff
@@ -4,3 +4,5 @@ pylint>=2.6.0
 pytest>=6.0.0
 pytest-cov>=2.10.1
 click==8.0.2
+mypy>=1.3.0
```
billschereriii, line 7: How did you decide which requirements to put here vs requirements-mypy.txt? Seems like they should all be in dev at this point

smartsim/_core/control/job.py
```diff
@@ -25,8 +25,10 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import time
+import typing as t
```
billschereriii, line 28: as far as I can see you're not using this import

smartsim/_core/control/manifest.py
```diff
@@ -82,13 +84,13 @@ def models(self):
...
     @property
-    def ensembles(self):
+    def ensembles(self) -> t.List[EntityList]:
         """Return Ensemble instances in Manifest
 
         :return: list of ensembles
         :rtype: List[Ensemble]
         """
-        _ensembles = []
+        _ensembles: t.List[EntityList] = []
```
billschereriii, line 93: You seem to have declared type on some, but not all, local variables. What's the defining point for when it's needed?

smartsim/_core/launcher/util/launcherUtil.py
```diff
@@ -24,24 +24,26 @@
...
+import typing as t
+
 
 class ComputeNode:  # cov-slurm
     """The ComputeNode class holds resource information
     about a physical compute node
     """
 
-    def __init__(self, node_name=None, node_ppn=None):
+    def __init__(self, node_name: t.Optional[str] = None, node_ppn: t.Optional[int]  =None) -> None:
```
billschereriii, line 35: spacing on the second None

billschereriii (changes requested):
A few small questions, but the vast majority of this PR looks really good. Thanks for all the hard work!

smartsim/_core/control/controller.py
```diff
@@ -408,7 +418,7 @@ def _launch_step(self, job_step, entity):
             logger.debug(f"Launching {entity.name}")
             self._jobs.add_job(job_step.name, job_id, entity, is_task)
 
-    def _create_batch_job_step(self, entity_list):
+    def _create_batch_job_step(self, entity_list: EntityList) -> t.Any:
```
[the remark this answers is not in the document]
ankona: That call-tree is out of scope for the public API. I use the minimum to avoid mypy complaining about non-typed methods called in a typed context for now. 

It will be fixed in my next PR when I add the typehints for the remaining code.

smartsim/_core/control/controller.py
```diff
@@ -56,7 +60,7 @@ class Controller:
     underlying workload manager or run framework.
     """
 
-    def __init__(self, launcher="local"):
+    def __init__(self, launcher: str="local") -> None:
```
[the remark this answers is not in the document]
ankona: Applied black styling

requirements-dev.txt
ankona: reqs.txt -> project deps
reqs-dev.txt -> deps for development machines
reqs-mypy.txt -> stubs required by mypy 

I can combine mypy/dev if the -mypy convention isn't desired

pyproject.toml
ankona: Nope, they allow us to apply different settings to different bits of target code. Here I'm just ignoring 3rd parties without stubs in one. The other is using a whitelist of the modules I want to include as required items to pass mypy.

In my next PR, I'll reverse the global ignore/local include pattern and the second overrides here will go away since it won't be necessary.

smartsim/_core/control/job.py
ankona: good 👁️ !

smartsim/_core/control/manifest.py
ankona: mypy identifies the items that can't be inferred. I err'd on the side of less clutter rather than adding a hints on variables where it wasn't necessary

smartsim/_core/launcher/util/launcherUtil.py
ankona: Applied black

smartsim/entity/strategies.py
```diff
@@ -40,31 +42,18 @@ def create_all_permutations(param_names, param_values):
...
     permutations = []
     for p in zip(*param_values):
         permutations.append(dict(zip(param_names, p)))
     return permutations
 
 
-def random_permutations(param_names, param_values, n_models):
-    import random
-
+def random_permutations(param_names: t.List[str], param_values: t.List[t.List[str]], n_models: int = 0) -> t.List[t.Dict[str, str]]:
```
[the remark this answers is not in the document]
ankona: Applied black.

Some of the types are ugly... I could alias some to clean this up "inline" but I'd hope we iterate on the API and refactor methods instead. Leaving them ugly was my incentive to revisit but I can solve however you prefer.

requirements-dev.txt
billschereriii: I'm agnostic on whether they're combined, but should mypy itself be in the mypy section?

smartsim/entity/strategies.py
```diff
@@ -40,31 +42,18 @@ def create_all_permutations(param_names, param_values):
...
-            if str(model_dict) not in permutation_strings:
-                permutation_strings.add(str(model_dict))
-                permutations.append(model_dict)
-        return permutations
+    permutations = create_all_permutations(param_names, param_values)
+    
+    if n_models and n_models < len(permutations):
+        permutations = random.sample(permutations, n_models)
+    
+    return permutations
```
billschereriii, line 65: Flagging this to come back and review more carefully -- big change to the implementation

billschereriii: From MattD:
looks like that should be roughly equivalent, if not 1-to-1 with the current lib call. We don’t do any kind of setting a seed for random any where in this file, right? The only thing that would make me hold off is if this needs to be “reproducibly random”, but iirc we don’t do have that constraint for the the random perms strategy

billschereriii: I think that this is ok, but let's document it somewhere. A comment here in the file would be sufficient

requirements-dev.txt
ankona: Combined -mypy.txt into -dev.txt

pyproject.toml
ankona: Removed `python_version` per conversation to rely on default behavior

smartsim/entity/strategies.py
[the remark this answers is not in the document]
ankona: Aliased the worst offender (StrategyFunction)

smartsim/entity/strategies.py
ankona: pulled this refactor out into a separate PR.

billschereriii (approved):
LGTM. Thanks for all your hard work on this one.
Btw, I really like the recent change of defining type alias names for come of the more complex types. We should replicate that elsewhere as it makes sense to do so to take advantage of the readability gains it provides

Issue #294, opened 2023-06-01
Starting an experiement with a `summary=True` delays execution


b-fg:
When calling an experiment to start with `summary=True` enabled, a count down is forced before running the experiment here
https://github.com/CrayLabs/SmartSim/blob/5817084bee1cb96a46e60b9a8821dc9896d3ffe8/smartsim/experiment.py#L826-L827
It would be nice to be able to print the summary without having to wait.

ashao:
Nice spot! I vaguely recall that we had a delay at some point (years ago) to allow time for the orchestrator to spin up, but I think we fixed that problem in a much better way than hardcoding a sleep time and just neglected to remove these lines from the codebase

ashao:
Ok, so checking in with others on the team was that the wait was a "last chance" cancel the launch of the experiment. If you want to see the summary before you start you can always call the `summary` method on your `Experiment` object. Does that suffice in the interim?

b-fg:
Okay! That `summary` produces a slightly different output though. So maybe that can be updated with more info as when using `summary=True`:

`Experiment.summary`:
```
'|    | Name                | Entity-Type   | JobID   | RunID   | Time   | Status    | Returncode   |\n|----|---------------------|---------------|---------|---------|--------|-----------|--------------|\n| 0  | Sod2D_RL_ensemble_0 | Model         | 860440  | 0       | 5.7321 | Completed | 0            |'
```

`summary=True`:
```
=== Launch Summary ===
Experiment: sod2d_exp
Experiment Path: /home/b-fg/Workspace/bsc1/SOD2D_DRL/relexi_sod/examples/bl/sod2d_exp
Launcher: local
Database Status: active

=== Ensembles ===
Sod2D_RL_ensemble
Members: 1
Batch Launch: False
```

b-fg:
Not blocking at all - no rush. I just pointed it out because I thought it could be improved. Thank you for the rapid response!

mellis13:
Closing this issue as it was addressed with #298

PR #298, opened 2023-06-05
Remove wait associated with Experiment start summary

mellis13:
Currently there is a 10 second pause if ``summary=true`` is provided to ``Experiment.start()``.  After receiving user feedback, we have decided to remove that pause (#294).  A separate issue has been posted to enable a preview of the experiment launch summary without invoking ``Experiment.start()``.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -25,6 +25,7 @@ Description
 
 A full list of changes and detailed notes can be found below:
 
+- Remove wait time associated with Experiment launch summary
 - Update and rename Redis conf file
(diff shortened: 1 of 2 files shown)


billschereriii:
I thought we were planning to add a dry run mode as well that would let users see the path of execution before committing to a run?

mellis13:
My understanding was that the ``Experimernt`` preview would be a separate PR that is out-of-scope of this release.  But, in the meantime the pause would be removed to address user feedback in #294

billschereriii (approved):
LGTM. Thanks for taking care of this!

PR #300, opened 2023-06-06
refactor random_permutations for simplicity

ankona:
Small refactor of `random_permutations` to simplify code.

Notes: 

- Given that`list(product(*param_values))` was called prior to the sampling code, I assume that the permutation count is manageable.
- Since the original code used a set to identify already retrieved permutations, random.sample (sample without replacement) should work identically


billschereriii (approved):
LGTM, just don't forget to update the changelog

PR #302, opened 2023-06-07
Better test port usage clean up

MattToast:
Fix a test that can leak an orc using a port


MattToast:
Problem was reproduced despite this change, this appears to be a rehearing. Closing as an unnecessary change.

PR #306, opened 2023-06-21
Allow pinning of the co-located database

ashao:
This updates the strategy for adding a co-located deployment for simulation and database. Prior, the co-located database was always pinned to the last N logical processors on a machine. In the case of a Slurm machine, user-specified bind settings were being overwritten leading to a general inflexibility for users trying to maximize performance. Additionally, `limit_app_cpus` was not working as intended because only the launcher command was being pinned and not the application launched by the launcher command.

The changes here provide two options, `limit_db_cpus` and `db_cpu_list` to control the pinning of the co-located database. `limit_db_cpus` enables the pinning using `taskset` and `db_cpu_list` is a user-provided string that specifies which processors the orchestrator should be bound to.  If `db_cpu_list` is not provided, the database is automatically pinned to the first N logical processors.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -27,7 +27,6 @@
 import json
 import os
 import inspect
-import shutil
 import pytest
 import psutil
 import shutil
@@ -44,6 +43,7 @@ from smartsim.settings import (
 from smartsim._core.config import CONFIG
 from smartsim.error import SSConfigError
 from subprocess import run
+import sys
 
 
 # Globals, yes, but its a testing file
@@ -115,7 +115,8 @@ def pytest_sessionfinish(session, exitstatus):
     returning the exit status to the system.
     """
     if exitstatus == 0:
-        shutil.rmtree(test_dir)
+        # shutil.rmtree(test_dir)
+        pass
     else:
         # kill all spawned processes in case of error
         kill_all_test_spawned_processes()
@@ -597,3 +598,38 @@ class MLUtils:
     @staticmethod
     def get_test_num_gpus():
         return test_num_gpus
+
+@pytest.fixture
+def coloutils():
+    return ColoUtils
+
+class ColoUtils:
+    def setup_test_colo(fileutils, db_type, exp_name, db_args, launcher="local"):
+        """Setup things needed for setting up the colo pinning tests"""
+
+        exp = Experiment(exp_name, launcher=launcher)
+
+        # get test setup
+        test_dir = fileutils.make_test_dir()
+        sr_test_script = fileutils.get_test_conf_path("send_data_local_smartredis.py")
+
+        # Create an app with a colo_db which uses 1 db_cpu
+        colo_settings = exp.create_run_settings(exe=sys.executable, exe_args=sr_test_script)
+        colo_model = exp.create_model(f"colocated_model", colo_settings)
+        colo_model.set_path(test_dir)
+
+        if db_type in ['tcp', "deprecated"]:
+            db_args["port"] = 6780
+            db_args["ifname"] = "lo"
+
+        colocate_fun = {
+            "tcp": colo_model.colocate_db_tcp,
+            "deprecated": colo_model.colocate_db,
+            "uds":colo_model.colocate_db_uds
+        }
+        colocate_fun[db_type](**db_args)
+
+        # assert model will launch with colocated db
+        assert colo_model.colocated
+        # Check to make sure that limit_db_cpus made it into the colo settings
+        return exp, colo_model
--- a/conftest.py
+++ b/conftest.py
@@ -543,7 +543,7 @@ class FileUtils:
         return dir_path
 
     @staticmethod
-    def make_test_dir(caller_function=None, caller_fspath=None, level=2):
+    def make_test_dir(caller_function=None, caller_fspath=None, level=1):
         """Create test output directory and return path to it.
 
         This function should be called without arguments from within
--- a/conftest.py
+++ b/conftest.py
@@ -604,11 +604,8 @@ def coloutils():
     return ColoUtils
 
 class ColoUtils:
-    def setup_test_colo(fileutils, db_type, exp_name, db_args, launcher="local"):
+    def setup_test_colo(fileutils, db_type, exp, db_args):
         """Setup things needed for setting up the colo pinning tests"""
-
-        exp = Experiment(exp_name, launcher=launcher)
-
         # get test setup
         test_dir = fileutils.make_test_dir()
         sr_test_script = fileutils.get_test_conf_path("send_data_local_smartredis.py")
@@ -632,4 +629,4 @@ class ColoUtils:
         # assert model will launch with colocated db
         assert colo_model.colocated
         # Check to make sure that limit_db_cpus made it into the colo settings
-        return exp, colo_model
+        return colo_model
--- a/conftest.py
+++ b/conftest.py
@@ -513,7 +513,7 @@ class FileUtils:
         return dir_path
 
     @staticmethod
-    def get_test_dir(caller_function=None, caller_fspath=None):
+    def get_test_dir(caller_function=None, caller_fspath=None, level=1):
         """Get path to test output.
 
         This function should be called without arguments from within
@@ -532,7 +532,7 @@ class FileUtils:
         :rtype: str
         """
         if not caller_function or not caller_fspath:
-            caller_frame = inspect.stack()[1]
+            caller_frame = inspect.stack()[level]
             caller_fspath = caller_frame.filename
             caller_function = caller_frame.function
 
@@ -543,7 +543,7 @@ class FileUtils:
         return dir_path
 
     @staticmethod
-    def make_test_dir(caller_function=None, caller_fspath=None):
+    def make_test_dir(caller_function=None, caller_fspath=None, level=2):
         """Create test output directory and return path to it.
 
         This function should be called without arguments from within
@@ -560,7 +560,7 @@ class FileUtils:
         :rtype: str
         """
         if not caller_function or not caller_fspath:
-            caller_frame = inspect.stack()[1]
+            caller_frame = inspect.stack()[level]
             caller_fspath = caller_frame.filename
             caller_function = caller_frame.function
 
@@ -603,11 +603,10 @@ def coloutils():
     return ColoUtils
 
 class ColoUtils:
-    @staticmethod
     def setup_test_colo(fileutils, db_type, exp, db_args):
         """Setup things needed for setting up the colo pinning tests"""
         # get test setup
-        test_dir = fileutils.make_test_dir()
+        test_dir = fileutils.make_test_dir(level=2)
         sr_test_script = fileutils.get_test_conf_path("send_data_local_smartredis.py")
 
         # Create an app with a colo_db which uses 1 db_cpu
@@ -625,7 +624,6 @@ class ColoUtils:
             "uds":colo_model.colocate_db_uds
         }
         colocate_fun[db_type](**db_args)
-
         # assert model will launch with colocated db
         assert colo_model.colocated
         # Check to make sure that limit_db_cpus made it into the colo settings
--- a/conftest.py
+++ b/conftest.py
@@ -115,8 +115,7 @@ def pytest_sessionfinish(session, exitstatus):
     returning the exit status to the system.
     """
     if exitstatus == 0:
-        # shutil.rmtree(test_dir)
-        pass
+        shutil.rmtree(test_dir)
     else:
         # kill all spawned processes in case of error
         kill_all_test_spawned_processes()
@@ -604,6 +603,7 @@ def coloutils():
     return ColoUtils
 
 class ColoUtils:
+    @staticmethod
     def setup_test_colo(fileutils, db_type, exp, db_args):
         """Setup things needed for setting up the colo pinning tests"""
         # get test setup
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -142,7 +142,7 @@ def _build_colocated_wrapper_cmd(
     # collect DB binaries and libraries from the config
 
     db_cmd = []
-    if limit_db_cpus:
+    if limit_db_cpus and db_cpu_list:
         db_cmd.extend([
             'taskset', '-c', db_cpu_list
         ])
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -81,8 +81,7 @@ def _build_colocated_wrapper_cmd(
     extra_db_args: t.Optional[t.Dict[str, str]] = None,
     port: int = 6780,
     ifname: t.Optional[t.Union[str, t.List[str]]] = None,
-    limit_db_cpus: bool = False,
-    db_cpu_list: t.Optional[str] = None,
+    custom_pinning: t.Optional[str] = None,
     **kwargs: t.Any,
 ) -> str:
     """Build the command use to run a colocated DB application
@@ -99,8 +98,6 @@ def _build_colocated_wrapper_cmd(
     :type port: int
     :param ifname: network interface(s) to bind DB to
     :type ifname: str | list[str], optional
-    :param limit_db_cpus: If True, limit the cpus that the database can run on
-    :type limit_db_cpus: bool, optional
     :param db_cpu_list: The list of CPUs that the database should be limited to
     :type db_cpu_list: str, optional
     :return: the command to run
@@ -114,16 +111,11 @@ def _build_colocated_wrapper_cmd(
     # the lock on the file.
     lockfile = create_lockfile_name()
 
-
-
-
     # create the command that will be used to launch the
     # database with the python entrypoint for starting
     # up the backgrounded db process
 
-    cmd = []
-    cmd.extend(
-        [
+    cmd = [
             sys.executable,
             "-m",
             "smartsim._core.entrypoints.colocated",
@@ -132,7 +124,6 @@ def _build_colocated_wrapper_cmd(
             "+db_cpus",
             str(cpus),
         ]
-    )
     # Add in the interface if using TCP/IP
     if ifname:
         if isinstance(ifname, str):
@@ -142,9 +133,9 @@ def _build_colocated_wrapper_cmd(
     # collect DB binaries and libraries from the config
 
     db_cmd = []
-    if limit_db_cpus and db_cpu_list:
+    if custom_pinning:
         db_cmd.extend([
-            'taskset', '-c', db_cpu_list
+            'taskset', '-c', custom_pinning
         ])
     db_cmd.extend([
         CONFIG.database_exe,
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -138,9 +138,10 @@ def _build_colocated_wrapper_cmd(
             'taskset', '-c', custom_pinning
         ])
     db_cmd.extend([
-        CONFIG.database_exe,
-        CONFIG.database_conf,
-        "--loadmodule", CONFIG.redisai]
+            CONFIG.database_exe,
+            CONFIG.database_conf,
+            "--loadmodule", CONFIG.redisai
+        ]
     )
 
     # add extra redisAI configurations
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -137,10 +137,12 @@ def _build_colocated_wrapper_cmd(
         db_cmd.extend([
             'taskset', '-c', custom_pinning
         ])
-    db_cmd.extend([
+    db_cmd.extend(
+        [
             CONFIG.database_exe,
             CONFIG.database_conf,
-            "--loadmodule", CONFIG.redisai
+            "--loadmodule",
+            CONFIG.redisai
         ]
     )
 
--- a/smartsim/_core/launcher/colocated.py
+++ b/smartsim/_core/launcher/colocated.py
@@ -122,9 +122,6 @@ def _build_colocated_wrapper_cmd(
     # up the backgrounded db process
 
     cmd = []
-    if limit_db_cpus:
-        cmd.extend(["taskset -c", f"{db_cpu_list}"])
-
     cmd.extend(
         [
             sys.executable,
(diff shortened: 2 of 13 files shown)


tests/on_wlm/test_colocated_model.py
```diff
@@ -31,95 +31,123 @@
...
 
+launcher = pytest.test_launcher
+supported_dbs = ["uds", "tcp", "deprecated"]
+
 # retrieved from pytest fixtures
-if pytest.test_launcher not in pytest.wlm_options:
+if launcher not in pytest.wlm_options:
     pytestmark = pytest.mark.skip(reason="Not testing WLM integrations")
 
+def _setup_test_colo(fileutils, db_type, exp_name, db_args, launcher):
```
mellis13, line 41: Maybe the code would be simpler if the experiment was created outside of this function and the experiment was passed in by reference?  Then no need to return tuple, etc.

tests/test_colo_model_local.py
```diff
@@ -37,55 +37,116 @@
...
+    exp, colo_model = _setup_test_colo(
+        fileutils,
+        db_type,
+        "colocated_model_pinning_auto_2cpu",
+        db_args,
+        launcher
+    )
+    assert colo_model.run_settings.colocated_db_settings["db_cpu_list"] == "0-1"
+    assert colo_model.run_settings.colocated_db_settings["limit_db_cpus"]
+    exp.start(colo_model)
```
mellis13, line 131: Should we test that all of these have a completed status?

mellis13 (changes requested):
Just some initial comments as you work on getting the CI/CD working

tests/on_wlm/test_colocated_model.py
ashao: I'm inclined to keep this as is since we do need both and otherwise we have two very small functions. Happy to revisit though

mellis13: Either way is fine.  What I was thinking wouldn't involve two functions -- I was thinking that the Experiment should be created as part of the test (see snippet below).  ``launcher`` and ``exp_name`` are passed in only to call the experiment constructor.  It might make it closer to the ideal of a function only doing one task so the code is more reusable later if this function wasn't also creating an experiment.

```python
@pytest.mark.parametrize("db_type", supported_dbs)
def test_colocated_model_pinning_auto_1cpu(fileutils, db_type):
     exp = Experiment(name, launcher)
     _setup_colo_model(fileutils, db_args, db_args, exp)
```

ashao: Ohhh I see what you mean. That makes sense.

smartsim/_core/launcher/colocated.py
```diff
@@ -100,6 +99,10 @@ def _build_colocated_wrapper_cmd(
     :type port: int
     :param ifname: network interface(s) to bind DB to
     :type ifname: str | list[str], optional
+    :param limit_db_cpus: If True, limit the cpus that the database can run on
+    :type limit_db_cpus: bool, optional
```
MattToast, line 103: Can we remove this parameter and just assume that if a user provides a `db_cpu_list` that is not None that they want to limit to those cpus?

smartsim/_core/launcher/colocated.py
```diff
@@ -111,26 +114,43 @@ def _build_colocated_wrapper_cmd(
     # the lock on the file.
     lockfile = create_lockfile_name()
 
+
+
+
```
MattToast, line 119: pedantic style thing: extra white space added

smartsim/_core/launcher/colocated.py
```diff
@@ -111,26 +114,43 @@ def _build_colocated_wrapper_cmd(
...
-        "-m",
-        "smartsim._core.entrypoints.colocated",
-        "+lockfile",
-        lockfile,
-        "+db_cpus",
-        str(cpus),
-    ]
+
+    cmd = []
+    cmd.extend(
```
MattToast, line 125: is there a reason for the change from declaring a populated list to declaring an empty list and then extending?

smartsim/_core/launcher/step/step.py
```diff
@@ -81,15 +83,11 @@ def get_colocated_launch_script(self) -> str:
...
-        # (not allowed b/c MacOS doesn't support it)
-        # TODO: support this only on linux
-        if (
-            self.__class__.__name__ == "LocalStep"
-            and db_settings["limit_app_cpus"] is True
-        ):  # pragma: no cover
-            logger.warning("Setting limit_app_cpus=False for local launcher")
-            db_settings["limit_app_cpus"] = False
+        if sys.platform == 'darwin':
+            if isinstance(self, LocalStep) and db_settings["limit_db_cpus"] is True:
```
MattToast, line 87: style thing: If you want to save a layer of nesting:
```py
if sys.platform == "darwin" and isinstance(self, LocalStep) and ...:
```

MattToast: follow up question: should we even bother with the downcast to local step? If a user is on macos, I'm pretty sure we are safe to assume that they are using a local launcher (or if they aren't, they're likely to have much bigger issues, hahaha)

smartsim/entity/model.py
```diff
@@ -152,12 +154,11 @@ def attach_generator_files(
 
     def colocate_db(self, *args: t.Any, **kwargs: t.Any) -> None:
         """An alias for ``Model.colocate_db_tcp``"""
-        warnings.warn(
+        logger.warning(
```
MattToast, line 157: is there a reason for the change from `warnings` to `logger` here? Only reason I ask is that in the past all deprecations were done via `warnings`

smartsim/entity/model.py
```diff
@@ -198,8 +200,11 @@ def colocate_db_uds(
         :type socket_permissions: int, optional
         :param db_cpus: number of cpus to use for orchestrator, defaults to 1
         :type db_cpus: int, optional
-        :param limit_app_cpus: whether to limit the number of cpus used by the app, defaults to True
-        :type limit_app_cpus: bool, optional
+        :param limit_db_cpus: whether to limit the number of cpus used by the database defaults to True
+        :type limit_db_cpus: bool, optional
```
MattToast, line 204: Similar question here: Is this parameter strictly needed? Can we assume that if `db_cpu_list is not None` that a user wants to limit db cpus to that list

smartsim/entity/model.py
```diff
@@ -223,7 +230,8 @@ def colocate_db_tcp(
         port: int = 6379,
         ifname: str = "lo",
         db_cpus: int = 1,
-        limit_app_cpus: bool = True,
+        limit_db_cpus: bool = True,
+        db_cpu_list: t.Optional[str] = None,
```
MattToast, line 234: same question here

smartsim/entity/model.py
```diff
@@ -291,7 +303,39 @@ def _set_colocated_db_settings(
         if hasattr(self.run_settings, "_prep_colocated_db"):
             self.run_settings._prep_colocated_db(common_options["cpus"])
 
+        if "limit_app_cpus" in common_options:
+            raise SSUnsupportedError(
+                "Pinning of app CPUs via limit_app_cpus is no supported. Modify RunSettings " +
```
MattToast, line 308:
```py
"Pinning of app CPUs via limit_app_cpus is no supported. Modify RunSettings " +
#                                          ^^ no -> not
```
Also style thing: The trailing `+` is not strictly needed for concatenation of two str literals, but it gets optimized away at the byte code level, so feel free to keep or leave depending on your own personal style

smartsim/entity/model.py
```diff
@@ -291,7 +303,39 @@ def _set_colocated_db_settings(
...
+
         # TODO list which db settings can be extras
+        cpus = common_options["cpus"]
+        # Deal with cases where the database should be pinned to cpus
+        # (1) if the user set a db_cpu_list, but not limit_db_cpus
+        if common_options["db_cpu_list"] and not common_options["limit_db_cpus"]:
+            logger.warning(
+                "limit_db_cpus is False, but db_cpu_list is not None. Setting limit_db_cpus=True"
+            )
+            common_options["limit_db_cpus"] = True
```
MattToast, line 320: ^^ Really seems like we can drop `limit_db_cpus` and just assume that if `db_cpus_list is not None` that we should limit to those cpus

tests/conftest/setup_test_colo/colocated_model.out
```diff
@@ -0,0 +1,4 @@
+SmartRedis Library@19-19-36:WARNING: Environment variable SR_LOG_FILE is not set. Defaulting to stdout
+SmartRedis Library@19-19-36:WARNING: Environment variable SR_LOG_LEVEL is not set. Defaulting to INFO
+default@19-19-36:ERROR: Unable to connect to backend database: Failed to connect to Redis: Connection refused
+Test worked! Sent and received array: [1 2 3 4]
```
MattToast, line 4: Output file from test? Can this be checked out of source control?

smartsim/entity/model.py
```diff
@@ -291,7 +303,39 @@ def _set_colocated_db_settings(
...
+        # (2) limit_db_cpus, but not db_cpu_list is specified automatically set
+        #     pin to the cpus 0:db_cpus-1
+        if common_options["limit_db_cpus"] and not common_options["db_cpu_list"]:
+            if cpus > 1:
+                cpu_list = f"0-{cpus-1}"
+            elif cpus == 1:
+                cpu_list = "0"
+            else:
```
MattToast, line 328: Should we add logic to handle the case were a user requests that the `db_cpu_list` contain more cpus than the number of requested cpus (e.g. `cpus == 1` and `cpu_list == "0,99"`)? Or should we chalk that up to "user error"?

MattToast (changes requested):
Gave some initial thoughts in a quick first pass while the cli was getting sorted out; lmk what you think!

smartsim/entity/model.py
MattToast: Just realized this is a bit vague. To be clear can we do something like this:
```py
    def colocate_db_uds(
        self,
        unix_socket: str = "/tmp/redis.socket",
        socket_permissions: int = 755,
        db_cpus: int = 1,
        limit_db_cpus: bool | str = True,
        # NOTE: no `db_cpu_list` param
        debug: bool = False,
        **kwargs: t.Any,
    ) -> None:
        ...
        if isinstance(limit_db_cpus, bool):
            if limit_db_cpus:
                db_cpu_list = "0" if db_cpus == 1 else f"0-{db_cpus-1}"
            else: 
                db_cpu_list = None
        else:
            db_cpu_list = limit_db_cpus
        # type(db_cpu_list) is now str | None
        ...
        common_options = {
            "cpus": db_cpus,
            # NOTE: no `limit_db_cpus` key
            "db_cpu_list": db_cpu_list,
            "debug": debug,
        }
        ...
```
That way we do not have to pass both parameters deeper into the call stack, when in reality we can infer all of the info we need from `db_cpu_list`

conftest.py
```diff
@@ -591,3 +592,38 @@ class MLUtils:
     def get_test_device():
         global test_device
         return test_device
+
+@pytest.fixture
+def coloutils():
+    return ColoUtils
```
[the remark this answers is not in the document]
ashao: Like all class-fixtures here, this is intended to return just the type. It should have all static methods

conftest.py
```diff
@@ -591,3 +592,38 @@ class MLUtils:
...
+
+class ColoUtils:
+    def setup_test_colo(fileutils, db_type, exp_name, db_args, launcher="local"):
```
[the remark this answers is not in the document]
ashao: Yes (facepalm)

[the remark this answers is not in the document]
ashao: While I agree with the cleanup overall (shoving all the fixtures potentially into nested classes), I'm a little hesitant about breaking the pattern right now. There's a lot of cleanup that we can/should do for the testing

conftest.py
```diff
@@ -113,7 +113,8 @@ def pytest_sessionfinish(session, exitstatus):
     returning the exit status to the system.
     """
     if exitstatus == 0:
-        shutil.rmtree(test_dir)
+        # shutil.rmtree(test_dir)
+        pass
```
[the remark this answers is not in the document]
ashao: Mistake to include in the commit. Was because I was debugging something that was only pretend succeeding

smartsim/_core/launcher/colocated.py
ashao: Good point and the answer is no.

smartsim/_core/launcher/colocated.py
ashao: Not too keen on removing `limit_db_cpus` as I think there's enough people who are ok with the 'automatic' creation of `db_cpu_list`. What I'd prefer to do is to raise an exception if `limit_db_cpus` is False but `db_cpu_list` isn't None. It's potentially ambiguous what the user wants to do.

ashao: I can see removing it in this routine though since this shouldn't be touched by the user.

smartsim/_core/launcher/step/step.py
ashao: Good point. It does seem odd to check whether a step is a LocalStep in just the colocated part of the code.

smartsim/entity/model.py
ashao: Yah so this is where I disagree, I think there's only one unallowed pathway which we should make sure that the users are aware of and fail fast on, which is `limit_db_cpus=False` and `db_cpus_list` is something.

ashao: Will followup offline

smartsim/entity/model.py
ashao: That's actually an allowed case because taskset just tells the application what CPUs it CAN be deployed in. So it's entirely possible for cpus=1 and then to have len(cpu_list)>1

smartsim/entity/model.py
ashao: I think it's because I didn't see any other usage of warnings in the codebase and got self-conscious. Agree that `warnings` is better.

tests/test_colo_model_local.py
```diff
@@ -24,68 +24,135 @@
...
 
+def test_create_pinning_string():
+    # Automatic creation of pinned cpu list
+    assert Model._create_pinning_string(None,2) == '0,1'
+    # Individual ids only
+    assert Model._create_pinning_string([1,2],2) == '1,2'
+    # Mixed ranges and individual ids
+    assert Model._create_pinning_string([range(2),3],3) == '0,1,3'
+    # Range only
+    assert Model._create_pinning_string(range(3),3) == '0,1,2'
```
MattToast, line 47: This such a good test case to parameterize!! I demand it be parameterized!!
```suggestion
@pytest.mark.parametrize("pin_list, num_cpus, expected", [
    pytest.param(None, 2, "0,1", id="Automatic creation of pinned cpu list"),
    pytest.param([1,2], 2, "1,2", id="Individual ids only"),
    pytest.param([range(2),3], 3, "0,1,3", id="Mixed ranges and individual ids"),
    pytest.param(range(3), 3, "0,1,3", id="Range only"),
    pytest.param([range(8, 10), range(6, 1, -2)], 4, "8,9,6,4,2", id="Multiple ranges"),
])
def test_create_pinning_string(pin_list, num_cpus, expected):
    assert Model._create_pinning_string(pin_list, num_cpus) == expected
```

smartsim/entity/model.py
```diff
@@ -315,6 +333,30 @@ def _set_colocated_db_settings(
...
+        """Create a comma-separated string CPU ids. By default, None returns
+        0,1,...,cpus-1; an empty iterable will disable pinning altogether,
+        and an iterable constructs a comma separate string (e.g. 0,2,5)
+        """
+        if pin_ids is None:
+            return ','.join(str(i) for i in range(cpus))
+        elif not pin_ids:
+            return None
+        elif isinstance(pin_ids, collections.abc.Iterable):
+            pin_list = []
```
MattToast, line 350: Thoughts on adding an
```suggestion
        elif isinstance(pin_ids, collections.abc.Iterable):
            if "darwin" in sys.platform:
                raise ValueError("Custom DB CPU pinning not available on darwin")
            pin_list = []
```
block here?

MattToast (changes requested):
Some more initial thoughts for you while we sort out the CI

smartsim/entity/model.py
```diff
@@ -315,6 +333,30 @@ def _set_colocated_db_settings(
...
+    def _create_pinning_string(
+        pin_ids: t.Optional[t.Iterable[t.Union(int, t.Iterable[int])]],
+        cpus: int
+        ):
+        """Create a comma-separated string CPU ids. By default, None returns
+        0,1,...,cpus-1; an empty iterable will disable pinning altogether,
+        and an iterable constructs a comma separate string (e.g. 0,2,5)
+        """
+        if pin_ids is None:
+            return ','.join(str(i) for i in range(cpus))
```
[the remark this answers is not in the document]
ashao: I'd prefer not to change the behavior of a given value based on a change of OS. I implemented a different default value for the `custom_pinning` argument above. Do you and @mellis13 agree?

smartsim/entity/model.py
ashao: Changed it so that it warns if pin_ids isn't an empty list on MacOS. It then follows the normal logic below.

smartsim/entity/model.py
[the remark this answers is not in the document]
mellis13: Tough choice.... I don't like changing the behavior of an input value based on the OS, but I'm also scared of disregarding user input based on OS.  As coded, will the user  get a warning that it has been disregarded?  If not, I think we should throw a warning (or even an error) at the top level of the API for Mac OS to maximize transparency.

smartsim/entity/model.py
```diff
@@ -315,6 +333,30 @@ def _set_colocated_db_settings(
...
+        elif isinstance(pin_ids, collections.abc.Iterable):
+            pin_list = []
+            for i in pin_ids:
+                if isinstance(i, collections.abc.Iterable):
+                    pin_list.extend([str(j) for j in i])
+                else:
+                    pin_list.append(str(i))
+            return ','.join(pin_list)
+        else:
+            raise TypeError("pin_ids must be an iterable of ints")
```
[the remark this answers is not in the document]
ashao: Agree. Check out the change to the warning below.

smartsim/entity/model.py
```diff
@@ -315,6 +333,30 @@ def _set_colocated_db_settings(
...
+            return ','.join(str(i) for i in range(cpus))
+        elif not pin_ids:
+            return None
+        elif isinstance(pin_ids, collections.abc.Iterable):
+            pin_list = []
+            for i in pin_ids:
+                if isinstance(i, collections.abc.Iterable):
+                    pin_list.extend([str(j) for j in i])
+                else:
+                    pin_list.append(str(i))
```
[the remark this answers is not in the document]
ashao: LOL yep I definitely probably would do that at some point. Made a new `_stringify_id` to deal with this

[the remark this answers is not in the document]
ashao: The two above are functionally the same, however I agree that it's probably easier to debug a colodb launch command if it at least doesn't have repetitions. Wrapped the resulting pin_list in a `set` before the `join`

tests/test_colo_model_local.py
ashao: LOL fair

tests/on_wlm/test_colocated_model.py
```diff
@@ -25,101 +25,154 @@
...
 import pytest
 
 from smartsim import Experiment, status
+from smartsim.entity import Model
+
+if sys.platform == "darwin":
+    supported_dbs = ["tcp", "deprecated"]
+else:
+    supported_dbs = ["uds", "tcp", "deprecated"]
+    supported_dbs = ["uds"]
```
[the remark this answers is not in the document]
ashao: Missed commit thanks!

smartsim/_core/launcher/step/step.py
```diff
@@ -81,21 +83,16 @@ def get_colocated_launch_script(self) -> str:
...
-        if (
-            self.__class__.__name__ == "LocalStep"
-            and db_settings["limit_app_cpus"] is True
-        ):  # pragma: no cover
-            logger.warning("Setting limit_app_cpus=False for local launcher")
-            db_settings["limit_app_cpus"] = False
+        if sys.platform == 'darwin' and db_settings["db_cpu_list"]:
+                logger.warning(
+                    "DB pinning is not supported on MacOS, setting limit_db_cpus = False")
+                db_settings["limit_db_cpus"] = False
```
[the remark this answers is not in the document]
ashao: Resolved higher up the stack.

smartsim/_core/launcher/step/step.py
```diff
@@ -69,6 +70,7 @@ def get_step_file(self, ending: str = ".sh", script_name: t.Optional[str] = None
         return osp.join(self.cwd, self.entity_name + ending)
 
     def get_colocated_launch_script(self) -> str:
+        from .localStep import LocalStep
```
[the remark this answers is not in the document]
ashao: Nice catch!

smartsim/_core/launcher/colocated.py
```diff
@@ -114,23 +114,34 @@ def _build_colocated_wrapper_cmd(
...
+    db_cmd = []
+    if custom_pinning:
+        db_cmd.extend([
+            'taskset', '-c', custom_pinning
+        ])
+    db_cmd.extend([
+        CONFIG.database_exe,
+        CONFIG.database_conf,
+        "--loadmodule", CONFIG.redisai]
+    )
```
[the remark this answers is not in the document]
ashao: I'm glad, glad I say, to cause you pain via clashing styles [ref: previous conversations about drip]

smartsim/entity/model.py
```diff
@@ -291,7 +298,18 @@ def _set_colocated_db_settings(
         if hasattr(self.run_settings, "_prep_colocated_db"):
             self.run_settings._prep_colocated_db(common_options["cpus"])
 
+        if "limit_app_cpus" in common_options:
```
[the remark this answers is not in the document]
MattToast: Re-rasing this issue as this block is still unreachable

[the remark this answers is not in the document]
ashao: Your wish is my command.

smartsim/_core/launcher/colocated.py
```diff
@@ -114,23 +114,37 @@ def _build_colocated_wrapper_cmd(
...
+            'taskset', '-c', custom_pinning
+        ])
+    db_cmd.extend(
+        [
+            CONFIG.database_exe,
+            CONFIG.database_conf,
+            "--loadmodule",
+            CONFIG.redisai
+        ]
+    )
```
MattToast, line 147: Ahh perfectly styled! :)

MattToast (approved):
LGTM (pending CI and Docs)!! Thanks for all the hard-work on this one!

PR #308, opened 2023-06-23
Refactor smart cli into subparsers

ankona:
Separate creation of the sub-command parsers from parser execution and decouple CLI definition & instantiation.


smartsim/_core/_cli/cli.py
```diff
@@ -28,70 +28,84 @@
...
+# from pkg_resources import require
+
+# import smartsim._core._cli as cli
+from smartsim._core._cli.utils import MenuItemConfig
+from smartsim._core._cli.clean import (
+    configure_parser as clean_parser,
+    execute as clean_execute,
+    execute_all as clobber_execute,
+)
+import smartsim._core._cli.clean as clean
```
MattToast, line 42: Is `clean` ever used? Looks like you should be free to rm this import

smartsim/_core/_cli/cli.py
```diff
@@ -28,70 +28,84 @@
...
-                print(option)
-                sys.exit(0)
-        print("Database (Redis or KeyDB) dependencies not found")
-        sys.exit(1)
-
-
-def main() -> None:
-    SmartCli()
+        self.parser = parser
+        self.args: t.Optional[argparse.Namespace] = None
```
MattToast, line 55: Purely stylistic: If we are going to have the `self.parsed_args` propery, can we make `self.args` into "protected" `self._args`, or do we need to keep public write access to that attr for some reason?

smartsim/_core/_cli/utils.py
```diff
@@ -81,3 +85,100 @@ def pip_install(packages: t.List[str], end_point: t.Optional[str] = None, verbos
...
+def get_db_path() -> t.Optional[Path]:
+    bin_path = get_install_path() / "_core" / "bin"
+    for option in bin_path.iterdir():
+        if option.name in ("redis-cli", "keydb-cli"):
+            return option
+    return None
+
+
+CliHandler = t.Callable[[Namespace], int]
+CliParseConfigurator = t.Callable[[ArgumentParser], None]
```
MattToast, line 140: Incredibly pedantic request: If these types are not meant to be used as annotations outside of this file can we `_` prefix them (not that it matters to an end user bc this is all in a "protected " `_core` dir anyway)

smartsim/_core/_cli/utils.py
```diff
@@ -81,3 +85,100 @@ def pip_install(packages: t.List[str], end_point: t.Optional[str] = None, verbos
...
+    def handler(self, value: CliHandler) -> None:
+        self._handler = value    
+
+    @property
+    def configurator(self) -> t.Optional[CliParseConfigurator]:
+        return self._config
+
+    @configurator.setter
+    def configurator(self, value: t.Optional[CliParseConfigurator]) -> None:
+        self._config = value
```
MattToast, line 183: Are these properties strictly needed if we are going to add both read and write access to the underlying data anyway? Would you be against just doing something like
```py
class MenuItemConfig:
    def __init__(...):
        # no "protected" attrs
        self.command = ...
        self.help = ...
        self. handler = ...
        self.configurator = ...

    # No properties on the class
```
The only reason I mention is because this looks like a pretty simple wrapper class around some data  and might be a good candidate to turn into a `dataclass` or something.

smartsim/_core/_cli/cli.py
```diff
@@ -28,70 +28,84 @@
...
+        return self.args
+
+
+def default_cli() -> SmartCli:
+    build = MenuItemConfig("build",
+                           "Build SmartSim dependencies (Redis, RedisAI, ML runtimes)",
+                           build_execute,
+                           build_parser)
+    
+    cleanx = MenuItemConfig("clean",
```
MattToast, line 92: Not wrong, just curious: Any reason for the name `cleanx`? Is it just to avoid the name conflict with the global `clean` namespace?

MattToast (changes requested):
Looks amazing! All of the help statements are now correct, and I love the clean up to remove the (imo unnecessary) classes and shift this to a more procedural/functional style! It should make testing much easier!!

I found a couple small knit picks, the most major being that there seems to be a type error being raised when installing for GPU. Otherwise, all other functionality seems to be working as I would expect it to!!

Thanks for the hard work on this one!!

smartsim/_core/_cli/build.py
```diff
@@ -77,358 +76,382 @@ def _install_torch_from_pip(versions: Versioner, device: str = "cpu", verbose: b
...
+    verbose = args.v
+    keydb = args.keydb
+    device = str(args.device)
+
+    # torch and tf build by default
+    pt = not args.no_pt
+    tf = not args.no_tf
+    onnx = args.onnx
+
+    build_env = BuildEnv(checks=False) if args.only_python_packages else BuildEnv()
```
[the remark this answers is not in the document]
ankona: ternary was gross for sure. i don't like nested negation so i split the difference into:

```
do_checks = not args.only_python_packages
build_env = BuildEnv(checks=do_checks)
```

smartsim/_core/_cli/cli.py
ankona: 100%. victim of my own iteration.

smartsim/_core/_cli/cli.py
ankona: great eye. this helps me get rid of my `cleanx` rename, too~

smartsim/_core/_cli/utils.py
ankona: Done! I was originally thinking that I might import these into the implementations of each menu item and that evaporated.

smartsim/_core/_cli/utils.py
ankona: I think I went through a couple iterations where I was computing/reusing some properties and ended up simplifying it so these have no value. Great eye.

smartsim/_core/_cli/cli.py
ankona: 100% meant to avoid that collision. Getting rid of unused import solved it so this is no longer the `kleenex` variable

smartsim/_core/_cli/build.py
```diff
@@ -77,358 +76,382 @@ def _install_torch_from_pip(versions: Versioner, device: str = "cpu", verbose: b
...
-                msg += "1.2.5 or "
-            msg += "1.2.7."
-            raise SetupError(msg)
-        try:
-            if not self.build_env.check_installed("onnx", self.versions.ONNX):
-                msg = (
-                    f"ONNX {self.versions.ONNX} not installed in python environment. "
-                    + f"Consider installing onnx=={self.versions.ONNX} with pip"
+        if device == "gpu":
+            gpu_env = build_env.get_cudnn_env()
```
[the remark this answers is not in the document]
ankona: found & fixed bug introduced when removing clean from class (self.build_env collision with local var of same name in `build_redis_ai`)

smartsim/_core/_cli/build.py
```diff
@@ -77,358 +76,382 @@ def _install_torch_from_pip(versions: Versioner, device: str = "cpu", verbose: b
...
-                logger.warning(msg)
             else:
-                logger.info(
-                    f"ONNX {self.versions.ONNX} installed in Python environment"
-                )
-        except SetupError as e:
-            logger.warning(str(e))
+                build_env.update(gpu_env)  # type: ignore
+        # update RAI build env with cudnn env vars
+        rai_builder.env = build_env # type: ignore
```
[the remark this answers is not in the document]
ankona: fixed with above change & removed suppression

smartsim/_core/_cli/build.py
```diff
@@ -77,358 +76,382 @@ def _install_torch_from_pip(versions: Versioner, device: str = "cpu", verbose: b
...
+        # NOTE: have the option to add other builds here in the future
+        # like "from_tarball"
+        rai_builder.build_from_git(
+            versions.REDISAI_URL, versions.REDISAI_BRANCH, device
+        )
+        logger.info("ML Backends and RedisAI build complete!")
+
+
+def infer_torch_device() -> str:
+    backend_torch_path = f"{CONFIG.lib_path}/backends/redisai_torch"
```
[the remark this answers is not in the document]
ankona: created an issue for followup

MattToast (approved):
LGTM!! Thanks for this refactor, this should be much nicer to work with!!

PR #309, opened 2023-06-24
Move ML dep installation to smart build

MattToast:
Moves all ml python package installation from `python setup.py`/`pip install` time to `smart build` time where we can be confident that users have configured their environment for to get the version of RedisAI that they would like to work with.

In order to prevent the CLI from unexpectedly modifying the python environment out from underneath a user. `smart build` will ONLY attempt to install/re-install/handle package conflicts if supplies the `--modify_python_env` flag (i.e `smart build --modify_python_env`).

This is with the notable exception of when a user requests to build the `torch` backend, does not provide a `--torch_dir`, and no existing `torch` package could be found. This is to match `smart`'s current behavior.


smartsim/_core/_cli/build.py
```diff
@@ -126,7 +104,17 @@ def __init__(self) -> None:
...
-            help="If true, only install the python packages (i.e. skip backend builds)",
+            help="If true, only validate the python packages (i.e. skip backend builds)",
+        )
+        parser.add_argument(
+            "--modify_python_env",
+            action="store_true",
+            default=False,
+            help=(
+                "If true, `smart` will use `pip` to attempt to modifiy the current "
+                "python enviornment to statify the package dependencies of smartsim "
```
ankona, line 115: statify != satisfy

ankona: enviornment != environment

billschereriii:
The behavior changes (new flags, etc.) should be mentioned in the docs

smartsim/_core/_cli/build.py
billschereriii: smartsim != SmartSim

smartsim/_core/_cli/build.py
```diff
@@ -248,14 +244,14 @@ def build_redis_ai(
...
-        # to download them if they are not installed. this should not break
-        # the build however, as we use onnx and tf directly from RAI instead
-        # of pip like we do PyTorch.
+        # Check for onnx and tf in user python environemnt or download
+        # them if requested. If they are not present and the user did not request
+        # they be downloaded, we still should not break the build, as we use
+        # onnx and tf directly from RAI instead of pip like we do PyTorch.
         if onnx:
-            self.check_onnx_install()
+            self.check_py_onnx_version(handle_conflict=self.modify_py_env)
```
billschereriii, line 252: Minor quibble with the name: check_version doesn't sound like it should be doing an installation

billschereriii: (applies to all backends)

smartsim/_core/_cli/build.py
```diff
@@ -265,12 +261,13 @@ def build_redis_ai(
                     # we will always be able to find a torch version downloaded by
                     # pip so if we can't find it we know the user suggested a torch
                     # installation path that doesn't exist
-                    logger.error("Could not find requested user Torch installation")
-                    sys.exit(1)
+                    raise SetupError("Could not find requested user Torch installation")
```
billschereriii, line 264: Do exceptions get captured into logfiles? If not, we may be losing the opportunity to share with users what went wrong

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
-                # if torch version if x.x.x then we need to install the cpu version
-                if "cpu" not in installed.patch and not self.build_env.is_macos():
-                    msg = (
-                        "Torch GPU installed in python environment but requested Torch CPU. "
-                        + " Run `pip uninstall torch torchvision` and run `smart build` again"
-                    )
-                    logger.error(msg)  # error because this is usually fatal
-            logger.info(f"Torch {self.versions.TORCH} installed in Python environment")
+                device_suffix = self.versions.TORCH_CPU_SUFFIX
+            elif device in ["gpu", "cuda"]:
```
billschereriii, line 345: We don't recognize "cuda" as a device in SmartRedis, which may be confusing to users. Is this an artifact of adding support for different vendors' GPUs?  Suggest opening a ticket against SmartRedis to trigger a group conversation on this

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
             device = "gpu"
         return device
 
-    def install_torch(self, device: str = "cpu") -> None:
+    def check_py_torch_version(
+        self, device: str = "cpu", handle_conflict: bool = False
```
billschereriii, line 332: looks like the handle_conflict parameter isn't used?

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+            # Previous check only looks at major/minor version numbers,
+            # Torch requires we look at the patch as well
+            installed = self.build_env.get_package_version(package)
+            if device_suffix and device_suffix not in installed.patch:
+                # Torch requires that we compare to the patch
+                raise VersionConflictError(
+                    f"{package}=={installed} does not satisfy device "
+                    f"suffix requirement: {device_suffix}"
+                )
+            return True
```
billschereriii, line 367: Nesting functions inside functions can make code hard to read. Are you doing this just to limit access scope? Might be better to take it out and just throw a few underscores at its name

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+        conflicts: t.List[str] = []
+        to_uninstall: t.List[str] = []
+        to_install: t.List[str] = []
+        validator = validator or self.build_env.check_installed
+        verbose_info: t.Callable[[str], t.Any] = lambda s: ...
+        if self.verbose:
+            verbose_info = logger.info
+
+        for name, version in packages.items():
+            spec = f"{name}=={version}" if version else name
```
billschereriii, line 429: How are we going to handle requests for package versions 3.4.x or for version >= 2.6.0?

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+            try:
+                if validator(name, version):
```
billschereriii, line 431: me being pedantic with naming again, something like "is_installed" or "validate_installed_version" might make more sense than "validator" here

smartsim/_core/_install/buildenv.py
```diff
@@ -487,13 +481,31 @@ def _torch_site_path() -> t.Optional[Path]:
...
+        dist = importlib.metadata.distribution(dist_pkg)
+        imp_pkg = imp_pkg or dist_pkg  # Many distributions ship an import of same name
+        imp_file = os.path.sep.join(imp_pkg.split("."))
+        path = Path(dist.locate_file(imp_file))
+        if path.is_dir():
+            return path
+        path = path.parent / (path.name + ".py")
+        if path.is_file():
+            return path
+        raise FileNotFoundError("Could not locate `{imp_pkg}` in dist `{dist_pkg}`")
```
ankona, line 507: i think you wanted an f string here

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+                # Incompatible version found
+                if install_on_conflict:
+                    verbose_info(f"{e}: Queueing `{name}` for reinstall")
+                    to_uninstall.append(name)
+                    to_install.append(spec)
+                else:
+                    conflicts.append(spec)
+
+        if missing or conflicts:
+            indent = "\n    "
```
billschereriii, line 453: Does Python support the \t escape sequence? Recommend using it if so

smartsim/_core/_install/buildenv.py
```diff
@@ -539,20 +557,26 @@ def check_build_dependency(self, command: str) -> None:
...
-            return True
-        except pkg_resources.DistributionNotFound:
+            installed = cls.get_package_version(package)
+        except importlib.metadata.PackageNotFoundError:
             return False
+        if version:
+            if not isinstance(version, Version_):
+                version = Version_(version)
+            # detect if major or minor versions differ
+            if installed.major != version.major or installed.minor != version.minor:
```
ankona, line 572: Could we create a comparison operator on Version_ to make this accessible in other places?

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+            fmt_list: t.Callable[[str, t.List[str]], str] = (
```
billschereriii, line 454: I would argue that the type hint here makes the code harder to read. Maybe just use t.Callable?

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+                lambda n, l: f"{n}:{indent}{indent.join(l)}" if l else ""
+            )
+            missing_str = fmt_list("Missing", missing)
+            conflict_str = fmt_list("Conflicting", conflicts)
+            sep = "\n" if missing_str else ""
```
billschereriii, line 459: should be `missing_str and conflict_str`

smartsim/_core/_cli/build.py
```diff
@@ -248,14 +244,14 @@ def build_redis_ai(
...
+        # Check for onnx and tf in user python environemnt or download
+        # them if requested. If they are not present and the user did not request
+        # they be downloaded, we still should not break the build, as we use
+        # onnx and tf directly from RAI instead of pip like we do PyTorch.
         if onnx:
-            self.check_onnx_install()
+            self.check_py_onnx_version(handle_conflict=self.modify_py_env)
         if tf:
-            self.check_tf_install()
+            self.check_py_tf_version(handle_conflict=self.modify_py_env)
```
ankona, line 254: It looks like we might be repeating these `self.check_py_XXX_version(...)` calls in `Build.__init__` and `Build.install_redis_ai`

Can we do this once?

- [install_redis_ai](https://github.com/CrayLabs/SmartSim/blob/f812630adfa41a6caffe91615d35cc12e66c0023/smartsim/_core/_cli/build.py#L251-L271)

smartsim/_core/_cli/utils.py
```diff
@@ -56,28 +57,50 @@ def color_bool(trigger: bool = True) -> str:
...
     if end_point:
-        cmd.extend(["-f", end_point])
+        cmd.extend(("-f", str(end_point)))
 
-    cmd_arg = " ".join(cmd)
+    cmd.extend(packages)
 
     if verbose:
-        logger.info(f"Installing packages {packages}...")
+        logger.info(f"Installing packages: {', '.join(packages)}")
```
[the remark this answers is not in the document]
billschereriii: We're assuming that the order of installation of packages won't matter, right? That should be the case unless two packages spec conflicting versions of a common dependency

smartsim/_core/_cli/utils.py
```diff
@@ -56,28 +57,50 @@ def color_bool(trigger: bool = True) -> str:
...
-        cmd_arg, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
     )
     _, err = proc.communicate()
     returncode = int(proc.returncode)
     if returncode != 0:
-        error = f"{packages} installation failed with exitcode {returncode}\n"
-        error += err.decode("utf-8")
+        error = (
+            f"'{' '.join(cmd)}' installation failed with exitcode {returncode}\n"
```
billschereriii, line 83: It's cool to be able to kick off all the installations with a single command, but it would be more useful for users to be able to tell them which package failed to install

smartsim/_core/_cli/utils.py
```diff
@@ -56,28 +57,50 @@ def color_bool(trigger: bool = True) -> str:
...
+            f"{err.decode('utf-8')}"
+        )
         raise BuildError(error)
     if verbose:
-        logger.info(f"{packages} installed successfully")
+        logger.info(f"{', '.join(packages)} installed successfully")
+
+
+def pip_uninstall(packages: t.List[str], verbose: bool = False) -> None:
```
billschereriii, line 91: really minor: renaming to "packages_to_remove" would help visually differentiate from the variable with the same name in the previous routine

smartsim/_core/_cli/utils.py
```diff
@@ -56,28 +57,50 @@ def color_bool(trigger: bool = True) -> str:
...
+        logger.info(f"Attempting to uninstall: {', '.join(packages)}")
+    cmd = [sys.executable, "-m", "pip", "uninstall", "-y"] + packages
+    proc = subprocess.Popen(
+        cmd, shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+    _, err = proc.communicate()
+    retcode = int(proc.returncode)
+    if retcode != 0:
+        raise BuildError(
+            f"'{' '.join(cmd)}' uninstall failed with exitcode {retcode}\n"
```
billschereriii, line 102: Same comment as before: Being able to say which package couldn't be removed would be more useful than "cleanup failed"

smartsim/_core/_install/buildenv.py
```diff
@@ -58,7 +69,11 @@ class SetupError(Exception):
     See setup.py for more
     """
 
-    pass
+
+class VersionConflictError(SetupError):
+    """An error for when version numbers of some library/package/program/etc
+    do not match and build may not be able to continue
+    """
```
billschereriii, line 76: Is there any potential benefit to stashing name, current_version, target_version in this class? The message could be automatically generated from those three

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+        install_on_absent: bool = False,
+        install_on_conflict: bool = False,
+    ) -> None:
+        missing: t.List[str] = []
+        conflicts: t.List[str] = []
+        to_uninstall: t.List[str] = []
+        to_install: t.List[str] = []
+        validator = validator or self.build_env.check_installed
+        verbose_info: t.Callable[[str], t.Any] = lambda s: ...
+        if self.verbose:
```
ankona, line 425: Do we expect the verbosity to change between method calls? If not, we can avoid putting these `if self.verbose: verbose_info = ...` lines anywhere but the object constructor and just use `self.verbose_info(...)`

smartsim/_core/_install/buildenv.py
```diff
@@ -388,18 +385,17 @@ class BuildEnv:
     JOBS = int(os.environ.get("BUILD_JOBS", 1))
 
     # check for CC/GCC/ETC
-    CHECKS = int(os.environ.get("NO_CHECKS", 0))
+    NO_CHECKS = int(os.environ.get("NO_CHECKS", 0))
```
billschereriii, line 388: thank you for this change!

smartsim/_core/_install/buildenv.py
```diff
@@ -487,13 +481,31 @@ def _torch_site_path() -> t.Optional[Path]:
...
+            Currently this function is only suited to finding namespace dirs or *.py
+            files. It is unable to handle other python importables (*.so, *.zip, etc.)
+        """
+        dist = importlib.metadata.distribution(dist_pkg)
+        imp_pkg = imp_pkg or dist_pkg  # Many distributions ship an import of same name
+        imp_file = os.path.sep.join(imp_pkg.split("."))
+        path = Path(dist.locate_file(imp_file))
+        if path.is_dir():
+            return path
+        path = path.parent / (path.name + ".py")
```
billschereriii, line 504: f"{path.name}.py"

smartsim/_core/_cli/build.py
ankona, line 453: If we separate out the output formatting, we could more easily write independent tests like `foo = _check_py_package_version(...); assert foo.conflicts; assert not foo.to_uninstall;`  (instead of capturing stdout and looking for output strings that will be less reliable)

smartsim/_core/_cli/utils.py
```diff
@@ -56,28 +57,50 @@ def color_bool(trigger: bool = True) -> str:
...
+def pip_install(
+    packages: t.List[str], end_point: t.Optional[str] = None, verbose: bool = False
+) -> None:
     """Install a pip package to be used in the SmartSim build
     Currently only Torch shared libraries are re-used for the build
     """
     # form pip install command
-    cmd = ["python", "-m", "pip", "install"]
-    cmd.extend(packages)
+    cmd = [sys.executable, "-m", "pip", "install"]
```
ankona, line 67: +1billion

smartsim/_core/_install/buildenv.py
```diff
@@ -539,20 +557,26 @@ def check_build_dependency(self, command: str) -> None:
...
+            installed = cls.get_package_version(package)
+        except importlib.metadata.PackageNotFoundError:
             return False
+        if version:
+            if not isinstance(version, Version_):
+                version = Version_(version)
+            # detect if major or minor versions differ
+            if installed.major != version.major or installed.minor != version.minor:
+                msg = (
+                    f"Incompatible version for {package} detected: "
```
billschereriii, line 574: do you need a `+` to concatenate these strings?

smartsim/_core/_cli/build.py
ankona, line 332: Do you have a `DeviceType = Literal['cpu','gpu']` lying around to tighten this hint?

smartsim/_core/_install/buildenv.py
MattToast: Good catch! best to save the extra mem alloc

smartsim/_core/_install/buildenv.py
MattToast: I did want an f-str there!

smartsim/_core/_install/buildenv.py
MattToast: You do not for string literals or f-strs. No difference in performance either; at the byte code level string concatenation of str literals and f-strs, both with and without  `+`, are optimized to be equivalent.

I simply prefer the no `+` just bc its less to type. I don't think we ever agreed upon a specific style for SS, but I imagine the with `+` is more prevalent. I'd be willing to change if we think it is best to match the existing style!

smartsim/_core/_install/buildenv.py
MattToast: Brilliant! That also sounds like useful info should one want to catch and handle that err

smartsim/_core/_cli/utils.py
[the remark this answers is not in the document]
MattToast: Good idea with the `'\n\t'`, and can confirm that the order we in which we install/uninstall packages should not matter! (assuming we trust `pip`'s dependency resolving)

smartsim/_core/_cli/utils.py
MattToast: hahaha anything to try and be a little more safe!

smartsim/_core/_cli/build.py
MattToast: good catch!

smartsim/_core/_cli/build.py
MattToast: fair point! It's pretty locally scoped. Changing!

smartsim/_core/_cli/build.py
MattToast: It does! I just reverted to the old pep8 prefer 4 spaces over tabs... that probably doesn't apply for formatting shell msgs, lol. Updated to use `\t`

MattToast: Brilliant idea!! Breaking logic out into a new helper fn and keeping this one around for formatting!

smartsim/_core/_cli/build.py
MattToast: Totally willing to do that!! This may even give up the ability to clean up some of the other `if verbose:` checks down the line!

smartsim/_core/_cli/build.py
```diff
@@ -331,84 +328,146 @@ def infer_torch_device(self) -> str:
...
+            validator=torch_validator,
+            install_on_absent=True,
+            install_on_conflict=handle_conflict,
+        )
 
-    def check_onnx_install(self) -> None:
-        """Check Python environment for ONNX installation"""
+    def check_py_onnx_version(self, handle_conflict: bool = False) -> None:
+        """Check Python environment for a compatible ONNX installation"""
+        logger.info(f"Searching for a compatible ONNX install...")
```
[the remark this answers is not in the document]
MattToast: 🤦

smartsim/_core/_cli/build.py
MattToast: Good point! This was just to limit scope, but probably not worth the cost of readability. Moved to a "private" static method on the Build class!

smartsim/_core/_cli/build.py
MattToast: Good point! This was holdover code from the previous `_install_torch_from_pip` top level function that I stole. Not sure why `"cuda"` is treated as an acceptable value...

Removing and added a more strict `choices=['cpu', 'gpu']` param to our argparser

smartsim/_core/_cli/build.py
MattToast: 🤦 Spelling is hard....

MattToast: modifiy != modify

smartsim/_core/_install/buildenv.py
MattToast: We actually do have the comparison operators for `Version_` implemented which just differ to the `packaging.Version` comparison ops. The problem is that those operations compare patch as well as the major and minor which we do not want.

Totally willing to revisit changing this behavior, but I would argue that it is out of scope for this ticket. Thoughts on putting up a new ticket instead?

smartsim/_core/_cli/build.py
MattToast: I can add one!

smartsim/_core/_cli/build.py
MattToast: Smart! Changing to `validate_installed_version`

smartsim/_core/_cli/build.py
billschereriii: If the message is getting logged, then I'm good. Removing sys.exit() calls seems like a good plan to me

smartsim/_core/_cli/build.py
MattToast: On a second pass, the reason for this was also in part because I was catching the `device_suffix` var in the closure. I broke out the construction of the torch version validation function into a separate factory method which is hopefully a bit more readable.

Let me know if you think this is now sufficiently clear, otherwise I can take another pass to try and clean it up in a more substantial way!

smartsim/_core/_cli/utils.py
MattToast: Done!

smartsim/_core/_cli/utils.py
MattToast: Agreed!! Changed!

smartsim/_core/_cli/build.py
MattToast: Good point! Now that we know that the RAI versions are more suggestions than hard pins, we could probably work on allowing more flexible versions. I did some preliminary work to get the ball rolling on this, but more thought could probably go into this! Adding a ticket!

smartsim/_core/_cli/build.py
MattToast: better named!!

smartsim/_core/_cli/build.py
MattToast: Potentially? I am tempted to say yes, but it world required changing the order in which things are built (e.g. instead of building `redis -> torch -> rai` it would need to be `torch -> redis -> rai`).

Given the time to code freeze, and the fact that some others are currently working on this section of the code, I'm a little hesitant to squeze that change into this PR. Would you be okay with a follow on ticket?

smartsim/_core/_cli/build.py
```diff
@@ -68,6 +72,7 @@ def __init__(self) -> None:
             "--device",
             type=str.lower,
             default="cpu",
+            choices=["cpu", "gpu"],
```
billschereriii, line 75: Will this fail with a weird error message if someone puts in CPU (uppercase)?

smartsim/_core/_cli/build.py
```diff
@@ -352,31 +361,43 @@ def check_py_torch_version(
...
+            if with_suffix and with_suffix not in installed.patch:
                 # Torch requires that we compare to the patch
                 raise VersionConflictError(
-                    f"{package}=={installed} does not satisfy device "
-                    f"suffix requirement: {device_suffix}"
+                    package,
+                    installed,
+                    version or f"X.X.X{with_suffix}",
+                    msg=(
+                        f"{package}=={installed} does not satisfy device "
```
billschereriii, line 390: Just a note that the `==` here will eventually have to be tweaked

billschereriii: Also, can we get this message down with the _check_py_package_version() method somehow?

billschereriii (approved):
Aside from a few minor notes, looks really great! Thanks for your hard work in getting this up to snuff.

ankona (approved):

smartsim/_core/_cli/build.py
ankona: `+1 love it` to removing those sys.exit calls. i think of them like the "view" in MVC - they make it harder to test behavior alone

smartsim/_core/_install/buildenv.py
ankona: I concur. Good on ya for refusing my feature creep. Maybe a modulo operator feels nice? `if installed % version is None: ...`

smartsim/_core/_cli/build.py
ankona: 100% with you on either skipping or doing follow up. I didn't realize the order change would have to occur. Up to your discretion

smartsim/_core/_cli/utils.py
```diff
@@ -65,26 +66,51 @@ def pip_install(
...
-        cmd.extend(["-f", end_point])
+        cmd.extend(("-f", str(end_point)))
+
+    packages.sort()
+    pkg_str_list = "\n\t".join(packages)
+
+    if verbose:
+        logger.info(f"Attempting to Install Packages:\n\t{pkg_str_list}")
+    for package in packages:
+        proc = subprocess.Popen(
```
ankona, line 80: the content of the install/uninstall loops are pretty similar. you could yank the bodies out & parameterize them (e.g. `cmd: str, pkg_name: str, action: str = "Install",`)... that would give you the ability to mock the actual installation and test that the build does try to install the packages you expect

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+        end_point = None
+        device_suffix = ""
+    else:  # linux
+        end_point = "https://download.pytorch.org/whl/torch_stable.html"
+        if device == "cpu":
+            device_suffix = Versioner.TORCH_CPU_SUFFIX
+        elif device == "gpu":
+            device_suffix = Versioner.TORCH_CUDA_SUFFIX
+        else:
+            raise BuildError("Unrecognized device requested")
```
ankona, line 252: will the linter complain here about an unnecessary else?

can we shift the validation to the start and short circuit out earlier?

```
if device not in ['cpu', 'gpu']:
    raise BuildError("Unrecognized device requested")
```

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+        conflict_str = fmt_list("Conflicting", conflicts)
+        sep = "\n" if missing_str and conflict_str else ""
+        logger.warning(
+            "Python Env Status Warning!\n"
+            "Requested Packages are Missing or Conflicting:\n\n"
+            f"{missing_str}{sep}{conflict_str}"
+            "\n\nConsider installing packages at the requested versions via "
+            "`pip` or re-running `smart build` with `--modify_python_env`"
+        )
+    if to_uninstall:
```
ankona, line 377: Does taking the install & uninstall actions belong in a method called `_confirm_package_in_python_env`?

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+    install_on_absent: bool,
+    install_on_conflict: bool,
+    verbose: bool,
+) -> t.Tuple[t.List[str], t.List[str], t.List[str], t.List[str]]:
+    missing: t.List[str] = []
+    conflicts: t.List[str] = []
+    to_uninstall: t.List[str] = []
+    to_install: t.List[str] = []
+
+    verbose_info: t.Callable[[str], t.Any] = lambda s: ...
```
ankona, line 396: this looks like something you could pull out into "global space" and reuse. Are we planning on adding any verbose logging elsewhere that we could gain from not repeating these lines?

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+        try:
+            if validate_installed_version(name, version):
+                # Installed at the correct version, nothing to do here
+                verbose_info(f"{spec} already installed in environment")
+            else:
+                # Not installed! Install preferred version or warn user
+                if install_on_absent:
+                    verbose_info(f"Package not found: Queueing `{spec}` for install")
+                    to_install.append(spec)
+                else:
```
ankona, line 411: unnecessary else from linter?

smartsim/_core/_cli/build.py
ankona: I'm thinking this might also be cause for a custom log level instead of this extra set of declarations?

smartsim/_core/_cli/utils.py
```diff
@@ -68,28 +69,55 @@ def pip_install(
...
-    )
-    _, err = proc.communicate()
-    returncode = int(proc.returncode)
-    if returncode != 0:
-        error = f"{packages} installation failed with exitcode {returncode}\n"
-        error += err.decode("utf-8")
-        raise BuildError(error)
+        logger.info(f"Attempting to Uninstall Packages:\n\t{pkg_str_list}")
+    for package in packages_to_remove:
+        _pip("uninstall", "-y", package)
```
ankona, line 90: could we avoid the hardcoded strings and make it easier to use by having:

```
def _pip(..., interactive=True):
   ...
```

this avoids yet another string constant to make sure case/format matches.

smartsim/_core/_install/buildenv.py
```diff
@@ -61,6 +72,29 @@ class SetupError(Exception):
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        current_version: "Version_",
+        target_version: "Version_",
+        msg: t.Optional[str] = None,
+    ) -> None:
+        if msg is None:
```
ankona, line 87: Do we want to override `__str__` and `__repr__` on SetupError children instead of having msg built in a constructor?

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+
+    return check_torch_version
+
+
+def install_py_onnx_version(
+    fetch_if_missing: bool = False,
+    verbose: bool = False,
+) -> None:
+    """Check Python environment for a compatible ONNX installation"""
+    logger.info("Searching for a compatible ONNX install...")
```
ankona, line 301: What do you think about adding a logging decorator to put on all the methods we do the immediate logging for?

```
@auto_logger("Searching for a compatible ONNX install...", )
def install_py_onnx_version(...):
  if not Versioner.ONNX:
    ....
```

I know that returns are more difficult unless we refactor to reliably raise exceptions, but at least we could start to avoid logger.info @ the start of half the methods..

smartsim/_core/_cli/build.py
ankona: nope. I'm forgetting I'm in a loop. my bad!

smartsim/_core/_cli/build.py
```diff
@@ -391,20 +492,46 @@ def execute(args: argparse.Namespace) -> int:
...
+    """Install python ML package dependencies via pip"""
+    if with_onnx:
+        install_py_onnx_version(
+            fetch_if_missing=fetch_if_missing,
+            verbose=verbose,
+        )
+    if with_tf:
+        install_py_tf_version(
+            fetch_if_missing=fetch_if_missing,
+            verbose=verbose,
```
ankona, line 527: I hate passing verbose around. It's not a parameter... it's a cross cutting configuration value. If we did custom log level, couldn't we remove the parameter and just log?

smartsim/_core/_cli/verify.py
```diff
@@ -0,0 +1,189 @@
...
+def verify_install(
+    location: str, port: int, with_tf: bool, with_pt: bool, with_onnx: bool
+) -> None:
+    exp = Experiment("Verification", exp_path=location, launcher="local")
+    with _make_managed_orc(exp, port) as client:
+        client.put_tensor("plain-tensor", np.ones((1, 1, 3, 3)))
+        client.get_tensor("plain-tensor")
+        if with_tf:
+            logger.info("Verifying TensorFlow Backend")
+            _verify_tf_install(client)
```
ankona, line 99: ditto on logging decorator possibly littering calling methods with logging like this, too.

smartsim/_core/_cli/build.py
MattToast: Good point! Will change!

smartsim/_core/_cli/build.py
MattToast: Love the concept of a custom "verbose log" log level, but a little bit worried about squeezing into this ticket. Maybe make a new one?

smartsim/_core/_install/buildenv.py
MattToast: I'd argue that we should still pass he message through the `super` constructor just so the the `Exception.message` attr gets set correctly. Iirc the default `__str__` basically amounts to
```py
def __str__(self):
    return str(self.message)
```
so I'm not sure that I would see the benefit of the override.

That said, I could be convinced that the construction of the message string that gets passed to `super().__init__` does clutter up the constructor. Maybe we should just move it into its own static helper method?

smartsim/_core/_cli/build.py
MattToast: Love it! Little worried about feature creep. Maybe added to the ticket with the "verbose mode log level"?

smartsim/_core/_cli/build.py
MattToast: Totally agree!!

smartsim/_core/_cli/utils.py
MattToast: change "-y" to kwarg

.github/workflows/run_tests.yml
```diff
@@ -108,15 +108,15 @@ jobs:
       - name: Install SmartSim (with ML backends)
         run: |
           python -m pip install git+https://github.com/CrayLabs/SmartRedis.git@develop#egg=smartredis
-          python -m pip install .[dev,ml]
```
ashao, line 111: Really happy to see this go away

smartsim/_core/_cli/build.py
```diff
@@ -30,104 +30,38 @@
...
+    pip_install,
+    pip_uninstall,
+)
 from smartsim._core._install import builder
 from smartsim._core._install.buildenv import (
     BuildEnv,
+    DbEngine,
     SetupError,
     Version_,
+    VersionConflictError,
```
ashao, line 48: This is not necessarily a "you" problem, but should these `Error`s actually be moved to the central `error.py`?

.github/workflows/run_tests.yml
```diff
@@ -108,15 +108,15 @@ jobs:
...
+          python -m pip install .[dev]
```
ashao, line 111: Could we throw a deprecation warning and still do a no-op if someone requests 'ml', since this is a change in behavior for users

smartsim/_core/_cli/build.py
```diff
@@ -285,73 +223,236 @@ def build_redis_ai(
...
+    package_pinning: _TPinningStr = "==",
+    end_point: t.Optional[str] = None,
+    validate_installed_version: t.Optional[
+        t.Callable[[str, t.Optional[Version_]], bool]
+    ] = None,
+    install_on_absent: bool = False,
+    install_on_conflict: bool = False,
+    verbose: bool = False,
+) -> None:
+    # TODO: Do not like how the defualt validation function will always look for
```
ashao, line 351: Small typo `defualt` -> `default`

smartsim/_core/_cli/verify.py
```diff
@@ -0,0 +1,189 @@
...
+
+from smartsim import Experiment
+from smartsim._core.utils.helpers import installed_redisai_backends
+from smartsim._core._cli.utils import SMART_LOGGER_FORMAT
+from smartsim.log import get_logger
+
+SMART_LOGGER_FORMAT = "[%(name)s] %(levelname)s %(message)s"
+logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
+
+# Many of the functions in this module will import optional
```
ashao, line 45: Update the comments in here to describe the purpose of this verify module.

smartsim/_core/_cli/verify.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+        return 2
+    return 0
+
+
+def verify_install(
+    location: str, port: int, with_tf: bool, with_pt: bool, with_onnx: bool
+) -> None:
+    exp = Experiment("Verification", exp_path=location, launcher="local")
+    with _make_managed_orc(exp, port) as client:
```
ashao, line 94: Since this verify check is always going to happen on the local. Can we add a check to make sure that the port is actually open. I'm a little bit worried about hitting false negatives just because it can't find a port. Throwing a more informative error that at least says, 'Eh things might work, but we can't verify because the port is being used'. 

```
import socket

with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as redis_socket:
  if redis_socket.connect_ex(('localhost',port)) ~= 0:
    raise SSConnectionError(f'Port {port} cannot be bound. Cannot verify RedisAI installation.')
  redis_socket.close()
```
     



If you wanted to be really fancy, give a list of a few non-commonly used ports and iterate through them to find one that's actually open.

tests/test_cli.py
```diff
@@ -668,3 +670,71 @@ def mock_operation(*args, **kwargs) -> int:
...
+    print("LGTM")
+
+def _bad_build(*args, **kwargs):
+    raise Exception
+
+@contextmanager
+def _totally_real_temp_dir(*a, **kw):
+    yield "/path/to/a/real/temp/dir"
+
+@pytest.mark.parametrize(
```
ashao, line 685: I like the compactness of the test, but it took me a couple of minutes to figure out to determine the validity of the test because of the monkeypatches. Please expand out the docstring for this test a bit.

ashao (changes requested):
This is an excellent update that demystifies our bizarre 2-step process. The docs definitely need to be updated to reflect the change in behavior of `smart build` and the surface the new `smart verify` functionality.

smartsim/_core/_cli/build.py
```diff
@@ -30,104 +30,38 @@
...
 from smartsim._core.config import CONFIG
 from smartsim._core.utils.helpers import installed_redisai_backends
 from smartsim.error import SSConfigError
 from smartsim.log import get_logger
 
-SMART_LOGGER_FORMAT = "[%(name)s] %(levelname)s %(message)s"
 logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 # NOTE: all smartsim modules need full paths as the smart cli
 #       may be installed into a different directory.
```
ashao, line 61: Would you mind putting a brief description for this module? It's definitely one of the weird bits that whenever we ask a dev to touch it's not immediately understandable

ashao (changes requested):
One additional comment

MattToast:
After further discussion offline, it was decided to move in a different direction for unifying SmartSim ML dependency installation behavior. Instead SmartSim will move the installation of the PyTorch package into the `pip install` step as an optional dependency and the RedisAI backend fetched separately during `smart build`, similar to how Tensorflow and Onnx are installed now. #336

PR #310, opened 2023-06-27
casting issues for nightlies

ankona:
Fix two defects created in the typehinting PRs:
1. The job-id was cast to an integer and should have been a string
2. A test for task_id having a null-or-zero value was added to incorrectly guard against adding steps without task_id from being added to history


MattToast (approved):
lgtm

smartsim/entity/dbnode.py
```diff
@@ -176,7 +176,7 @@ def _parse_db_host(self, filepath: t.Optional[str] = None) -> str:
             try:
                 ip = self._parse_ips(filepath, 1)[0]
             # suppress error
-            except FileNotFoundError:
+            except (FileNotFoundError, IndexError):
```
MattToast, line 179: Since this was a bit of red herring, should we remove this handle for `IndexError`s just to revert to the original behavior?

MattToast (approved):
One small question, but otherwise lgtm!

mellis13 (approved):

smartsim/entity/dbnode.py
ankona: Done!

PR #311, opened 2023-06-30
Update, Apply, and Automate Python Linting

ankona:
This PR mitigates linting issues that have slipped in over time. It includes the following subtasks:

- Updates pylint configuration to newer format.
- Applies pylint as a step in CI workflow (initially non-blocking)
- Mitigates as many issues as possible
- Where mitigation is not possible (API break, etc), disable linter.


.github/workflows/run_tests.yml
```diff
@@ -118,6 +118,16 @@ jobs:
...
+      - name: Run mypy
+        # TF 2.6.2 has a dep conflict with new mypy versions
+        if: (matrix.rai != '1.2.5')
+        run: |
+          python -m pip install .[mypy]
+          make check-mypy
+
+      - name: Run Pylint
+        run: make check-lint
+
```
billschereriii, line 130: You might need to update the changes to the yml file here based on what happened in SmartRedis

.pylintrc
```diff
@@ -1,3 +1,11 @@
+# Copyright (c) 2022, Hewlett Packard Enterprise Development Company LP
```
billschereriii, line 1: Copyright year needs update

billschereriii: Might want to check where you cloned the copyright statement from as well

smartsim/_core/_cli/build.py
```diff
@@ -170,15 +184,18 @@ def __init__(self) -> None:
                     CONFIG.conf_path = Path(CONFIG.core_path, "config", "keydb.conf")
                     if not CONFIG.conf_path.resolve().is_file():
                         raise SSConfigError(
-                            "Database configuration file at REDIS_CONF could not be found"
+                            f"Database configuration file at {CONFIG.conf_path} "
+                            + "could not be found"
```
billschereriii, line 188: Per discussions with MattD, the + should be omitted in multiline string constructions

smartsim/_core/_cli/build.py
```diff
@@ -214,7 +231,10 @@ def build_database(self) -> None:
         )
         if not database_builder.is_built:
             logger.info(
-                f"Building {database_name} version {self.versions.REDIS} from {self.versions.REDIS_URL}"
+                "Building {} version {} from {}",
+                database_name,
+                self.versions.REDIS,
+                self.versions.REDIS_URL,
```
billschereriii, line 237: Seems like it would be better to just do a multiline string here:
```python
f"Building {database_name} version {self.versions.REDIS} "
f"from {self.versions.REDIS_URL}"
```

smartsim/_core/_cli/build.py
```diff
@@ -349,17 +369,19 @@ def install_torch(self, device: str = "cpu") -> None:
                 # if torch version is x.x.x+cpu
                 if "cpu" in installed.patch:
                     msg = (
-                        "Torch CPU is currently installed but torch GPU requested. Uninstall all torch packages "
-                        + "and run the `smart build` command again to obtain Torch GPU libraries"
+                        "Torch CPU is currently installed but torch GPU requested. "
+                        + "Uninstall all torch packages and run the `smart build` "
+                        + "command again to obtain Torch GPU libraries"
```
billschereriii, line 374: delete + (x2)

smartsim/_core/_cli/build.py
```diff
@@ -349,17 +369,19 @@ def install_torch(self, device: str = "cpu") -> None:
...
 
             if device == "cpu":
                 # if torch version if x.x.x then we need to install the cpu version
                 if "cpu" not in installed.patch and not self.build_env.is_macos():
                     msg = (
-                        "Torch GPU installed in python environment but requested Torch CPU. "
-                        + " Run `pip uninstall torch torchvision` and run `smart build` again"
+                        "Torch GPU installed in python environment but requested "
+                        + "Torch CPU. Run `pip uninstall torch torchvision` and "
+                        + "run `smart build` again"
```
billschereriii, line 384: delete +

smartsim/_core/_cli/build.py
```diff
@@ -399,18 +421,22 @@ def check_tf_install(self) -> None:
                 "tensorflow", self.versions.TENSORFLOW
             ):
                 msg = (
-                    f"TensorFlow {self.versions.TENSORFLOW} not installed in Python environment. "
-                    + f"Consider installing tensorflow=={self.versions.TENSORFLOW} with pip"
+                    "TensorFlow %s not installed in Python environment. "
+                    + "Consider installing tensorflow==%s with pip",
```
billschereriii, line 425: %s should be {}?

billschereriii: delete +

smartsim/_core/_cli/build.py
```diff
@@ -399,18 +421,22 @@ def check_tf_install(self) -> None:
...
                 logger.info(
-                    f"TensorFlow {self.versions.TENSORFLOW} installed in Python environment"
+                    "TensorFlow {} installed in Python environment",
+                    self.versions.TENSORFLOW,
                 )
         except SetupError as e:
             logger.warning(str(e))
 
-    def check_backends_install(self) -> None:
+    @staticmethod
```
billschereriii, line 438: You made this a `staticmethod` because `self` is never accessed?

smartsim/_core/_cli/build.py
```diff
@@ -419,16 +445,29 @@ def check_backends_install(self) -> None:
...
+                    ),
+                    CONFIG.redisai,
                 )
             else:
                 logger.error(
-                    f"Before running 'smart build', unset your RAI_PATH environment variable with 'unset RAI_PATH'."
+                    (
+                        "Before running 'smart build', unset your RAI_PATH environment "
+                        "variable with 'unset RAI_PATH'."
+                    )
```
billschereriii, line 459: parens unneeded?

smartsim/_core/_cli/build.py
```diff
@@ -419,16 +445,29 @@ def check_backends_install(self) -> None:
...
+            if installed := installed_redisai_backends():
+                logger.info(
+                    f"Installed redis backends ({redis_install_base()}): {installed}",
+                )
                 logger.error(
-                    "If you wish to re-run `smart build`, you must first run `smart clean`."
+                    (
+                        "If you wish to re-run `smart build`, you must first "
+                        "run `smart clean`."
+                    )
```
billschereriii, line 471: parens unneeded?

smartsim/_core/_cli/cli.py
```diff
@@ -67,23 +67,28 @@ def __init__(self) -> None:
             sys.exit(0)
         getattr(self, args.command)()
 
-    def build(self) -> None:
+    @staticmethod
```
billschereriii, line 70: making build/clean/clobber/site() staticmethods keeps the self pointer from being passed to the Build/Clean/Clobber/Site methods. I think this is a dangerous change

smartsim/_core/_cli/utils.py
```diff
@@ -33,18 +34,21 @@
...
+logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 
 def get_install_path() -> Path:
-    try:
-        import smartsim as _
-    except (ImportError, ModuleNotFoundError):
+    module_spec = importlib.util.find_spec("smartsim")
+
+    if module_spec is None:
```
billschereriii, line 44: You handled the same issue in SmartRedis differently (there, by disabling the lint check for non-top-of-file lint check).
While this change is very similar functionally, I'm not sure whether it's identical

smartsim/_core/_cli/utils.py
```diff
@@ -33,18 +34,21 @@
...
         raise SetupError("Could not import SmartSim") from None
 
     # find the path to the setup script
-    package_path = Path(_.__path__[0]).resolve()
+    package_dir = ""
+    if module_spec.submodule_search_locations:
+        package_dir = module_spec.submodule_search_locations[0]
+    package_path = Path(package_dir).resolve()
```
billschereriii, line 51: I'm having a lot of difficulty seeing whether this code is equivalent to what it's replacing. Need to have other reviewers check it more carefully.

smartsim/_core/_cli/utils.py
```diff
@@ -70,6 +76,7 @@ def pip_install(packages: t.List[str], end_point: t.Optional[str] = None, verbos
 
     if verbose:
         logger.info(f"Installing packages {packages}...")
+    # pylint: disable-next=consider-using-with
     proc = subprocess.Popen(
         cmd_arg, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
billschereriii, line 81: We shouldn't have any shell=True commands. Feel free to just open a ticket rather than any more immediate addressing of the comment

smartsim/_core/_install/buildenv.py
```diff
@@ -58,24 +61,28 @@ class SetupError(Exception):
     See setup.py for more
     """
 
-    pass
```
billschereriii, line 61: We don't need the pass as an explicit declaration that the class is a name-only overload?

smartsim/_core/_install/buildenv.py
```diff
@@ -58,24 +61,28 @@ class SetupError(Exception):
...
 
 # so as to not conflict with pkg_resources.packaging.version.Version
+# pylint: disable-next=invalid-name
 class Version_(str):
     """A subclass of pkg_resources.packaging.version.Version that
     includes some helper methods for comparing versions.
     """
 
-    def _convert_to_version(self, vers: t.Union[str, Iterable[packaging.version.Version], packaging.version.Version]) -> t.Any:
+    @staticmethod
```
billschereriii, line 72: This one would be better as a classmethod so you can use the class parameter in the isinstance checks

smartsim/_core/_install/buildenv.py
```diff
@@ -126,6 +133,9 @@ def __le__(self, cmp: t.Any) -> bool:
         except InvalidVersion:
             return super().__le__(cmp)
 
+    def __hash__(self) -> int:
+        return hash(str(self))
```
billschereriii, line 137: Functional change?

smartsim/_core/_install/builder.py
```diff
@@ -174,7 +182,7 @@ def run_command(self,
             if proc.returncode != 0:
                 raise BuildError(error)
         except (OSError, SubprocessError) as e:
-            raise BuildError(e)
+            raise BuildError(e) from e
```
billschereriii, line 185: This change seems redundant

smartsim/_core/config/config.py
```diff
@@ -105,7 +105,8 @@ def redisai(self) -> str:
         redisai = Path(os.environ.get("RAI_PATH", rai_path)).resolve()
         if not redisai.is_file():
             raise SSConfigError(
-                "RedisAI dependency not found. Build with `smart` cli or specify RAI_PATH"
+                "RedisAI dependency not found. Build with `smart` cli "
+                + "or specify RAI_PATH"
```
billschereriii, line 109: delete +

smartsim/_core/control/controller.py
```diff
@@ -566,14 +566,14 @@ def _orchestrator_launch_wait(self, orchestrator: Orchestrator) -> None:
...
-        try:
+        with JM_LOCK:
             if self.orchestrator_active:
                 raise SmartSimError("Orchestrator exists and is active")
 
             if not osp.exists(checkpoint_file):
                 raise FileNotFoundError(
-                    f"The SmartSim database config file {checkpoint_file} cannot be found."
+                    f"The SmartSim database config file {checkpoint_file} "
+                    + "cannot be found."
```
billschereriii, line 576: delete +

smartsim/_core/control/controller.py
```diff
@@ -638,26 +636,26 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
 
         for model in manifest.models:
             if not model.colocated:
-                for db_model in model._db_models:
+                for db_model in model.db_models:
```
billschereriii, line 639: This change is exposing `db_models` to users where before it was protected. I'll defer to the rest of the team on the advisability of so doing

smartsim/_core/control/jobmanager.py
```diff
@@ -328,20 +324,19 @@ def set_db_hosts(self, orchestrator: Orchestrator) -> None:
...
         # should only be called during launch in the controller
-        self._lock.acquire()
-        try:
+        with self._lock:
             if orchestrator.batch:
                 self.db_jobs[orchestrator.name].hosts = orchestrator.hosts
             else:
                 for dbnode in orchestrator.dbnodes:
-                    if not dbnode._mpmd:
+                    if not dbnode.is_mpmd:
```
billschereriii, line 332: Another case of exposing a protected variable to users

smartsim/_core/control/manifest.py
```diff
@@ -126,58 +128,54 @@ def _check_entity_lists_nonempty(self) -> None:
...
     @property
     def has_db_objects(self) -> bool:
         """Check if any entity has DBObjects to set"""
 
-        def has_db_models(entity: t.Union[EntityList, SmartSimEntity]) -> bool:
-            if hasattr(entity, "_db_models"):
-                return len(entity._db_models) > 0
-            return False
+        def has_db_models(entity: t.Union[EntityList, Model]) -> bool:
+            return len(entity.db_models) > 0
```
billschereriii, line 175: We're guaranteeing that entity always has db_models now?

billschereriii: Is it worth adding an assert?
```python
assert hasattr(entity, "db_models")
```

smartsim/_core/control/manifest.py
```diff
@@ -187,7 +185,7 @@ def has_db_scripts(entity: t.Union[EntityList, SmartSimEntity]) -> bool:
         # we update has_db_objects so that as soon as one check
         # returns True, we can exit
         has_db_objects |= any(
-            [has_db_models(model) | has_db_scripts(model) for model in self.models]
+            has_db_models(model) | has_db_scripts(model) for model in self.models
```
billschereriii, line 188: From the code above, these should always be `True` now?

smartsim/_core/entrypoints/colocated.py
```diff
@@ -203,37 +207,51 @@ def main(
...
-                    f"\t# of Database CPUs: {db_cpus}",
+                    "\tIP Address(es): {}"
+                    "\tCommand: {}\n\n",
+                    "\t# of Database CPUs: {}",
                 )
             )
+            ),
+            ' '.join(ip_addresses + [lo_address]),
+            ' '.join(cmd),
+            db_cpus
```
billschereriii, line 221: I could be wrong, but I don't think the new code behaves the same as the old

smartsim/_core/entrypoints/colocated.py
```diff
@@ -264,52 +280,62 @@ def cleanup() -> None:
             os.remove(LOCK.lock_file)
 
 
+def register_signal_handlers() -> None:
+    for sig in SIGNALS:
+        signal.signal(sig, handle_signal)
```
billschereriii, line 285: Another signal handling change that looks like it snuck in from a different PR

.pylintrc
ankona: Great catch. Done in both places

smartsim/_core/_cli/build.py
ankona: used multiline string

smartsim/_core/_cli/build.py
ankona: 👍

smartsim/_core/_cli/build.py
ankona: replaced borked string with version that didn't go over line-length limit:
```
tf_version = self.versions.TENSORFLOW
msg = (
    f"TensorFlow {tf_version} not installed in Python environment. "
    f"Consider installing tensorflow=={tf_version} with pip"
)
```

smartsim/_core/_cli/build.py
ankona: pylint dings anything that doesn't use self. I can disable that check if you wanted to avoid the statics.

smartsim/_core/_cli/build.py
```diff
@@ -419,16 +445,29 @@ def check_backends_install(self) -> None:
         if os.environ.get("RAI_PATH"):
             if installed_redisai_backends():
                 logger.error(
-                    f"There is no need to build. Backends are already built and specified in the environment at 'RAI_PATH': {CONFIG.redisai}"
+                    (
+                        "There is no need to build. Backends are already built and "
+                        "specified in the environment at 'RAI_PATH': {}"
+                    ),
```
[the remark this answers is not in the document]
ankona: yeah, it was for the lazy eval of log format:

```
logger.error(
    (
        "There is no need to build. Backends are already built and "
        "specified in the environment at 'RAI_PATH': {}"
    ),
    CONFIG.redisai,
)
```
It looks like I can use an f-string now and not go over the line length limit. switching to:
```
logger.error(
    (
        "There is no need to build. Backends are already built and "
        f"specified in the environment at 'RAI_PATH': {CONFIG.redisai}"
    )
)
```

smartsim/_core/_cli/build.py
ankona: without them, the string breaks line length limit. a bunch had to be broken for this reason. E.g. 

```
************* Module smartsim._core._cli.build
smartsim/_core/_cli/build.py:456:0: C0301: Line too long (115/88) (line-too-long)
```

smartsim/_core/_cli/build.py
ankona: ditto on line-length

smartsim/_core/_cli/cli.py
ankona: I wasn't concerned since it wasn't used, but more importantly - I have a refactor of the CLI coming next (see: https://github.com/CrayLabs/SmartSim/pull/308) and it does away with these methods altogether.

smartsim/_core/_cli/utils.py
ankona: I did black-box testing with `smart clean` and `smart site` to ensure the same behavior. It was definitely a mitigation for the import location and I considered doing the same thing in SmartRedis. In SR, though, we alias and _use_ the import. Here, we were just letting exceptions control logic.

I'll do another sanity check here. Good eye

smartsim/_core/_cli/utils.py
ankona: <ditto above>

smartsim/_core/_install/buildenv.py
ankona: If we didn't have the comment, the pass would be required.

smartsim/_core/launcher/slurm/slurmLauncher.py
```diff
@@ -263,7 +275,10 @@ def check_for_slurm() -> None:
         :raises LauncherError: if no access to slurm
         """
         if not which("sbatch") and not which("sacct"):
-            error = "User attempted Slurm methods without access to Slurm at the call site.\n"
+            error = (
+                "User attempted Slurm methods without access to Slurm "
+                + "at the call site.\n"
```
billschereriii, line 280: delete +

smartsim/_core/launcher/slurm/slurmParser.py
```diff
@@ -80,8 +80,8 @@ def jobid_exact_match(parsed_id: str, job_id: str) -> bool:
     """
     if "." in job_id:
         return parsed_id == job_id
-    else:
-        return parsed_id.split(".")[0] == job_id
+
+    return parsed_id.split(".")[0] == job_id
```
billschereriii, line 84: If I'm not mistaken, the entire `if` clause is unnecessary here

smartsim/_core/launcher/step/lsfStep.py
```diff
@@ -221,16 +221,16 @@ def _build_exe(self) -> t.List[str]:
         args = self.run_settings.exe_args
         if self.run_settings.mpmd:
             erf_file = self.get_step_file(ending=".mpmd")
-            _ = self._make_mpmd()
+            self._make_mpmd()
```
billschereriii, line 224: I dislike this change -- explicitly showing that there is a return that we're disregarding is useful internal documentation in code

smartsim/_core/launcher/step/mpiStep.py
```diff
@@ -125,10 +125,10 @@ def _build_exe(self) -> t.List[str]:
...
         if self.run_settings.mpmd:
             return self._make_mpmd()
-        else:
-            exe = self.run_settings.exe
-            args = self.run_settings.exe_args
-            return exe + args
+
+        exe = self.run_settings.exe
+        args = self.run_settings.exe_args
+        return exe + args
```
billschereriii, line 131: Same as a previous comment, this could be merged into a single f-string

smartsim/_core/launcher/step/step.py
```diff
@@ -87,10 +93,11 @@ def get_colocated_launch_script(self) -> str:
...
 
+    # pylint: disable=no-self-use
     def add_to_batch(self, step: Step) -> None:
         """Add a job step to this batch
 
         :param step: a job step instance e.g. SrunStep
         :type step: Step
         """
-        raise NotImplementedError
+        raise SmartSimError("add_to_batch not implemented for this step type")
```
billschereriii, line 103: Why the change of exception types? This seems like a bad idea

smartsim/_core/launcher/stepInfo.py
```diff
@@ -57,22 +57,23 @@ def __str__(self) -> str:
...
         if status in SMARTSIM_STATUS:
             return SMARTSIM_STATUS[status]
-        elif status in self.mapping:
+        if status in self.mapping:
             if returncode is not None and returncode != 0:
                 return STATUS_FAILED
-            else:
-                return self.mapping[status]
+            return self.mapping[status]
         return STATUS_FAILED
```
billschereriii, line 77: Redundant code here.

```python
if status in self.mapping and (returncode is None or returncode == 0):
    return self.mapping[status]
return STATUS_FAILED
```

smartsim/_core/launcher/stepInfo.py
```diff
@@ -153,15 +153,16 @@ def __init__(
 
 
 class PBSStepInfo(StepInfo):  # cov-pbs
-
     @property
     def mapping(self) -> t.Dict[str, str]:
-        # see http://nusc.nsu.ru/wiki/lib/exe/fetch.php/doc/pbs/PBSReferenceGuide19.2.1.pdf#M11.9.90788.PBSHeading1.81.Job.States
+        # see http://nusc.nsu.ru/wiki/lib/exe/fetch.php/doc/pbs
+        # /PBSReferenceGuide19.2.1.pdf#M11.9.90788.PBSHeading1.81.Job.States
```
billschereriii, line 159: Don't break the URL here. Just flag to ignore line length

smartsim/_core/launcher/stepInfo.py
```diff
@@ -232,7 +233,8 @@ def __init__(
 class LSFBatchStepInfo(StepInfo):  # cov-lsf
     @property
     def mapping(self) -> t.Dict[str, str]:
-        # see https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=execution-about-job-states
+        # see https://www.ibm.com
+        # /docs/en/spectrum-lsf/10.1.0?topic=execution-about-job-states
```
billschereriii, line 237: Don't break the URL here. Just flag to ignore line length

smartsim/_core/launcher/stepInfo.py
```diff
@@ -265,7 +267,8 @@ def __init__(
 class LSFJsrunStepInfo(StepInfo):  # cov-lsf
     @property
     def mapping(self) -> t.Dict[str, str]:
-        # see https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=execution-about-job-states
+        # see https://www.ibm.com
+        # /docs/en/spectrum-lsf/10.1.0?topic=execution-about-job-states
```
billschereriii, line 271: Don't break the URL here. Just flag to ignore line length

smartsim/_core/_install/buildenv.py
ankona: Attempting to hash a Version_ pre-change raises a TypeError (e.g. `hash(Version_('1.2.3'))`):

```
tests/test_buildenv.py::test_version_hash failed: def test_version_hash():
        test_version = Version_("1.2.3")
>       version_hash = hash(test_version)
E       TypeError: unhashable type: 'Version_'

tests/test_buildenv.py:35: TypeError
```

The edited version fixes this behavior but may be improved. Instead of hashing the stringified version, I swapped to an implementation consistent with the rest of the Version_ dunder methods:

```
    def __hash__(self) -> int:
        return hash(Version(self))
```

NOTE: Version_ is broken in multiple ways. I added some basic tests of hash & equality to start, but I also added this test demonstrating that it does no validation and is unreliable:

```
def test_version_bad_input():
    """Test behavior when passing an invalid version string"""
    v1 = Version_("abcdefg")

    # todo: fix behavior to ensure versions are valid.
    assert v1


def test_version_bad_parse_fail():
    """Test behavior when trying to parse with an invalid input string"""
    v1 = Version_("abcdefg")

    with pytest.raises(IndexError):
        _ = v1.minor
```

smartsim/_core/launcher/util/shell.py
```diff
@@ -120,14 +122,19 @@ def execute_async_cmd(
...
             output, error = popen_obj.communicate()
-            err_msg = ""
             if output:
                 err_msg += output.decode("utf-8") + " "
             if error:
                 err_msg += error.decode("utf-8")
             raise ShellError("Command failed immediately", cmd_list, details=err_msg)
     except OSError as e:
-        raise ShellError("Failed to run command", cmd_list, details=e) from None
+        raise ShellError("Command failed", cmd_list, details=e) from None
```
billschereriii, line 138: Changing the text of an error message could impact users who have scripts that parse output looking for specific messages

smartsim/_core/utils/helpers.py
```diff
@@ -183,12 +190,12 @@ def cat_arg_and_value(arg_name: str, value: str) -> str:
 
     if arg_name.startswith("--"):
         return "=".join((arg_name, str(value)))
```
billschereriii, line 192: f"{arg_name}={value}"

billschereriii: Similar for the following clauses. Using join() seems unnecessarily heavyweight here

smartsim/_core/utils/helpers.py
```diff
@@ -198,9 +205,17 @@ def _installed(base_path: Path, backend: str) -> bool:
...
     backend_path = base_path / backend_key / f"{backend_key}.so"
     backend_so = Path(os.environ.get("RAI_PATH", backend_path)).resolve()
-    
+
     return backend_so.is_file()
 
+
+def redis_install_base(backends_path: t.Optional[str] = None) -> Path:
+    # pylint: disable-next=import-outside-toplevel
+    from ..._core.config import CONFIG
```
billschereriii, line 214: Why not just move this import to the top?

smartsim/_core/_install/buildenv.py
```diff
@@ -195,7 +205,8 @@ def __init__(self, vers: str) -> None:
                 self.version = "1.2.7"
             else:
                 raise SetupError(
-                    f"Invalid RedisAI version {vers}. Options are {self.defaults.keys()}"
+                    f"Invalid RedisAI version {vers}. Options are "
+                    + f"{self.defaults.keys()}"
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/database/orchestrator.py
```diff
@@ -379,16 +418,20 @@ def set_batch_arg(self, arg: str, value: t.Optional[str] = None) -> None:
...
+        if not hasattr(self, "batch_settings") or not self.batch_settings:
             raise SmartSimError("Not running as batch, cannot set batch_arg")
 
         if arg in self._reserved_batch_args[type(self.batch_settings)]:
             logger.warning(
-                f"Can not set batch argument {arg}: it is a reserved keyword in Orchestrator"
+                (
+                    "Can not set batch argument {}: "
+                    "it is a reserved keyword in Orchestrator"
+                ),
```
billschereriii, line 430: parens unnecessary?

smartsim/_core/_install/buildenv.py
```diff
@@ -549,8 +569,8 @@ def check_installed(package: str, version: t.Optional[Version_] = None) -> bool:
                 # detect if major or minor versions differ
                 if installed.major != version.major or installed.minor != version.minor:
                     msg = (
-                        f"Incompatible version for {package} detected.\n"
-                        + f"{package} {version} requested but {package} {installed} installed."
+                        f"Incompatible version for {package} detected.\n{package} "
+                        + f"{version} requested but {package} {installed} installed."
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/database/orchestrator.py
```diff
@@ -406,12 +449,16 @@ def set_run_arg(self, arg: str, value: t.Optional[str] = None) -> None:
         """
         if arg in self._reserved_run_args[type(self.entities[0].run_settings)]:
             logger.warning(
-                f"Can not set run argument {arg}: it is a reserved keyword in Orchestrator"
+                (
+                    "Can not set run argument {}: it is a reserved keyword "
+                    "in Orchestrator"
+                ),
```
billschereriii, line 455: parens unnecessary?

smartsim/database/orchestrator.py
```diff
@@ -514,15 +562,17 @@ def set_db_conf(self, key: str, value: t.Union[int, str]) -> None:
                 ) from None
             except TypeError:
                 raise TypeError(
-                    "Incompatible function arguments. The key and value used for setting the database configurations must be strings."
+                    "Incompatible function arguments. The key and value used for "
+                    + "setting the database configurations must be strings."
```
billschereriii, line 566: delete +

smartsim/database/orchestrator.py
```diff
@@ -514,15 +562,17 @@ def set_db_conf(self, key: str, value: t.Union[int, str]) -> None:
...
                 ) from None
         else:
             raise SmartSimError(
-                "The SmartSim Orchestrator must be active in order to set the database's configurations."
+                "The SmartSim Orchestrator must be active in order to set the "
+                + "database's configurations."
```
billschereriii, line 571: delete +

smartsim/database/orchestrator.py
```diff
@@ -760,8 +813,12 @@ def _check_network_interface(self) -> None:
             if interface not in net_if_addrs and interface != "lo":
                 available = list(net_if_addrs.keys())
                 logger.warning(
-                    f"{interface} is not a valid network interface on this node. \n"
-                    "This could be because the head node doesn't have the same networks, if so, ignore this."
+                    (
+                        "{} is not a valid network interface on this node. \n"
+                        "This could be because the head node doesn't have the same "
+                        "networks, if so, ignore this."
+                    ),
```
billschereriii, line 820: parens unnecessary?

smartsim/_core/_install/builder.py
ankona: the `from` ensures that the exception information is included with the newly raised exception. Since there is no logging here, the from should be critical to identify where the original exception occurred when handling the BuildError

smartsim/entity/dbobject.py
```diff
@@ -108,8 +111,11 @@ def _enumerate_devices(self) -> t.List[str]:
         """
         devices = []
         if ":" in self.device and self.devices_per_node > 1:
-            msg = "Cannot set devices_per_node>1 if a device numeral is specified, "
-            msg += f"the device was set to {self.device} and devices_per_node=={self.devices_per_node}"
+            msg = (
+                "Cannot set devices_per_node>1 if a device numeral is specified, "
+                + f"the device was set to {self.device} and "
```
billschereriii, line 116: delete +

smartsim/_core/config/config.py
ankona: 👍

smartsim/_core/control/controller.py
```diff
@@ -226,7 +220,8 @@ def get_entity_status(self, entity: t.Union[SmartSimEntity, EntityList]) -> str:
         """
         if not isinstance(entity, (SmartSimEntity, EntityList)):
             raise TypeError(
-                f"Argument must be of type SmartSimEntity or EntityList, not {type(entity)}"
+                "Argument must be of type SmartSimEntity or EntityList, "
+                + f"not {type(entity)}"
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/_core/control/controller.py
ankona: 👍

smartsim/entity/model.py
```diff
@@ -358,50 +363,51 @@ def _stringify_id(id: int) -> str:
...
     def params_to_args(self) -> None:
         """Convert parameters to command line arguments and update run settings."""
         if self.params_as_args is not None:
             for param in self.params_as_args:
                 if not param in self.params:
                     raise ValueError(
-                        f"Tried to convert {param} to command line argument "
-                        + f"for Model {self.name}, but its value was not found in model params"
+                        f"Tried to convert {param} to command line argument for Model "
+                        + f"{self.name}, but its value was not found in model params"
```
billschereriii, line 401: delete +

smartsim/entity/model.py
```diff
@@ -358,50 +363,51 @@ def _stringify_id(id: int) -> str:
...
                     )
                 if self.run_settings is None:
                     raise ValueError(
-                        f"Tried to configure command line parameter for Model {self.name}, "
-                        + "but no RunSettings are set."
+                        "Tried to configure command line parameter for Model "
+                        + f"{self.name}, but no RunSettings are set."
```
billschereriii, line 406: delete +

smartsim/entity/model.py
```diff
@@ -546,7 +552,10 @@ def add_function(
         db_script = DBScript(
             name=name, script=function, device=device, devices_per_node=devices_per_node
         )
-        self._append_db_script(db_script)
+        self.append_db_script(db_script)
+
+    def __hash__(self) -> int:
+        return hash(self.name)
```
billschereriii, line 558: I'm not clear on the functional impact of this change

smartsim/entity/model.py
```diff
@@ -593,17 +603,18 @@ def _check_db_objects_colo(self) -> None:
             if not db_model.is_file:
                 err_msg = (
                     "ML model can not be set from memory for colocated databases.\n"
+                    + f"Please store the ML model named {db_model.name} in binary "
+                    + "format and add it to the SmartSim Model as file."
```
billschereriii, line 607: delete +

smartsim/entity/model.py
```diff
@@ -593,17 +603,18 @@ def _check_db_objects_colo(self) -> None:
...
             if db_script.func:
                 if not isinstance(db_script.func, str):
-                    err_msg = "Functions can not be set from memory for colocated databases.\n"
-                    err_msg += f"Please convert the function named {db_script.name} to a string or store it "
-                    err_msg += "as a text file and add it to the SmartSim Model with add_script."
+                    err_msg = (
+                        "Functions can not be set from memory for colocated "
+                        + "databases.\nPlease convert the function named "
+                        + f"{db_script.name} to a string or store it as a text"
+                        + "file and add it to the SmartSim Model with add_script."
```
billschereriii, line 618: delete +

smartsim/ml/data.py
```diff
@@ -103,7 +108,10 @@ def download(self, client: Client) -> None:
         except RedisReplyError:
             # If the info was not published, proceed with default parameters
             logger.warning(
-                f"Could not retrieve data for DataInfo object, the following values will be kept."
+                (
+                    "Could not retrieve data for DataInfo object, the following "
+                    "values will be kept."
+                )
```
billschereriii, line 114: parens unnecessary?

smartsim/ml/data.py
```diff
@@ -371,18 +384,16 @@ def __len__(self) -> int:
...
-
         self.update_data()
         # Generate data
         if len(self) < 1:
-            msg = "Not enough samples in generator for one batch. "
-            msg += "Please run init_samples() or initialize generator with init_samples=True"
-            raise ValueError(msg)
+            raise ValueError(
+                "Not enough samples in generator for one batch. Please run "
+                + "init_samples() or initialize generator with init_samples=True"
```
billschereriii, line 395: delete +

smartsim/_core/control/controller.py
ankona: It seems clear we do need them to be publicly accessible given that we're already touching the properties. Making the variable names match their accessibility is purely cosmetic.

A refactor that pushes functionality like this method from the controller into the ensemble (or model) could also solve this, but feels out of scope of this task.

smartsim/_core/control/jobmanager.py
ankona: ditto on response. If we're already touching it externally, the name should reflect that until a refactor changes our access patterns.

smartsim/ml/data.py
```diff
@@ -391,24 +402,24 @@ def __iter__(self) -> t.Iterator[t.Tuple[np.ndarray, np.ndarray]]:
...
+        for ``init_trials`` times.
         :param init_trials: maximum number of attempts to fetch data
         :type init_trials: int
         """
         self._client = Client(self.address, self.cluster)
 
         num_trials = 0
         max_trials = init_trials or -1
-        while not len(self) and num_trials != max_trials:
+        while not self and num_trials != max_trials:
```
billschereriii, line 414: Seems like a functional change in the case where self is not None but has zero items in it?

smartsim/ml/data.py
```diff
@@ -391,24 +402,24 @@ def __iter__(self) -> t.Iterator[t.Tuple[np.ndarray, np.ndarray]]:
...
             self._update_samples_and_targets()
             self.log(
                 "DataLoader could not download samples, will try again in 10 seconds"
             )
             time.sleep(10)
             num_trials += 1
 
-        if not len(self):
+        if not self:
```
billschereriii, line 422: see previous comment

smartsim/ml/data.py
```diff
@@ -471,12 +482,8 @@ def _update_samples_and_targets(self) -> None:
 
             # Strictly greater, because next_index is 0-based
             if list_length > self.next_indices[uploader_idx]:
-                indices = [
-                    idx
-                    for idx in range(
-                        self.next_indices[uploader_idx], list_length, self.num_replicas
-                    )
-                ]
+                start, end = self.next_indices[uploader_idx], list_length
```
billschereriii, line 485: please don't do two assignments on the same line

smartsim/ml/tf/__init__.py
```diff
@@ -41,13 +41,17 @@
...
+        f"TensorFlow {TF_VERSION} is not installed. "
+        + "Please install it to use smartsim.tf"
     ) from None
 except AssertionError:  # pragma: no cover
-    raise SmartSimError(
-        f"TensorFlow >= {TF_VERSION} is required for smartsim.tf, you have {tf.__version__}"
-    ) from None
+    msg = (
+        f"TensorFlow >= {TF_VERSION} is required for smartsim. "
+        + f"tf, you have {tf.__version__}"
```
billschereriii, line 50: delete +

smartsim/ml/tf/data.py
```diff
@@ -34,19 +34,20 @@
 class _TFDataGenerationCommon(DataDownloader, keras.utils.Sequence):
     def __getitem__(self, index: int) -> t.Tuple[np.ndarray, np.ndarray]:
         if len(self) < 1:
-            msg = "Not enough samples in generator for one batch. "
-            msg += "Please run init_samples() or initialize generator with init_samples=True"
-            raise ValueError(msg)
+            raise ValueError(
+                "Not enough samples in generator for one batch. Please "
+                + "run init_samples() or initialize generator with init_samples=True"
```
billschereriii, line 39: delete +

smartsim/ml/tf/data.py
```diff
@@ -88,9 +89,11 @@ def __init__(self, **kwargs: t.Any) -> None:
         kwargs["dynamic"] = False
         super().__init__(**kwargs)
         if dynamic:
-            self.log(
-                "Static data generator cannot be started with dynamic=True, setting it to False"
+            msg = (
+                "Static data generator cannot be started with dynamic=True, "
+                + "setting it to False"
```
billschereriii, line 94: delete +

smartsim/ml/tf/data.py
```diff
@@ -106,9 +109,11 @@ def __init__(self, **kwargs: t.Any) -> None:
         kwargs["dynamic"] = True
         super().__init__(**kwargs)
         if not dynamic:
-            self.log(
-                "Dynamic data generator cannot be started with dynamic=False, setting it to True"
+            msg = (
+                "Dynamic data generator cannot be started with dynamic=False,"
+                + " setting it to True"
```
billschereriii, line 114: delete +

smartsim/_core/control/manifest.py
ankona: Given the typehints, we are able to start removing these checks. Yes, users can still send something else but it would be caught during static analysis instead of run time.

smartsim/ml/tf/utils.py
```diff
@@ -59,7 +61,7 @@ def freeze_model(model: keras.Model, output_dir: str, file_name: str) -> t.Tuple
     if not file_name.endswith(".pb"):
         file_name = file_name + ".pb"
 
-    full_model = tf.function(lambda x: model(x))
+    full_model = tf.function(model)
```
billschereriii, line 64: I'm not seeing the equivalence here

smartsim/ml/tf/utils.py
```diff
@@ -96,7 +98,7 @@ def serialize_model(model: keras.Model) -> t.Tuple[str, t.List[str], t.List[str]
     :rtype: str, list[str], list[str]
     """
 
-    full_model = tf.function(lambda x: model(x))
+    full_model = tf.function(model)
```
billschereriii, line 101: I'm not seeing the equivalence here

smartsim/ml/torch/data.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, **kwargs: t.Any) -> None:
         super().__init__(**kwargs)
         if init_samples:
             self.log(
-                "PyTorch Data Generator has to be created with init_samples=False. Setting it to False automatically."
+                "PyTorch Data Generator has to be created with "
+                + "init_samples=False. Setting it to False automatically."
```
billschereriii, line 43: delete +

smartsim/ml/torch/data.py
```diff
@@ -123,7 +125,8 @@ def __init__(self, **kwargs: t.Any) -> None:
         super().__init__(**kwargs)
         if not dynamic:
             self.log(
-                "Dynamic data generator cannot be started with dynamic=False, setting it to True"
+                "Dynamic data generator cannot be started with dynamic=False, "
+                + "setting it to True"
```
billschereriii, line 129: delete +

smartsim/ml/torch/data.py
```diff
@@ -139,11 +142,12 @@ def _worker_init_fn(worker_id: int) -> None:
         num_replicas=dataset.num_replicas * num_workers,
     )
     dataset.log(
-        f"Worker {worker_id+1}/{num_workers}: dataset replica {dataset.replica_rank+1}/{dataset.num_replicas}"
+        f"Worker {worker_id+1}/{num_workers}: dataset replica "
+        + str(dataset.replica_rank + 1 / dataset.num_replicas)
```
billschereriii, line 146: PEMDAS error in translation

smartsim/_core/control/manifest.py
ankona: There's a difference between having the property and having items in the collection. If the underlying collections are empty, false will still be possible.

smartsim/settings/base.py
```diff
@@ -511,24 +520,20 @@ def _build_exe_args(exe_args: t.Optional[t.Union[str, t.List[str]]]) -> t.List[s
...
                 exe_args = copy.deepcopy(exe_args)
-                plain_type = all([isinstance(arg, (str)) for arg in exe_args])
+                plain_type = all(isinstance(arg, (str)) for arg in exe_args)
                 if not plain_type:
                     nested_type = all(
-                        [
-                            all([isinstance(arg, (str)) for arg in exe_args_list])
-                            for exe_args_list in exe_args
-                        ]
+                        all(isinstance(arg, (str)) for arg in exe_args_list)
```
billschereriii, line 526: Might want to double-check that this is equivalent

smartsim/settings/containers.py
```diff
@@ -73,11 +73,15 @@ def _containerized_run_command(self, run_command: str) -> str:
         :type run_command: str
         """
         raise NotImplementedError(
-            f"Containerized run command specification not implemented for this Container type: {type(self)}"
+            "Containerized run command specification not implemented for this "
+            + f"Container type: {type(self)}"
```
billschereriii, line 77: delete +

smartsim/settings/containers.py
```diff
@@ -140,14 +146,17 @@ def _container_cmds(self, default_working_directory: str = "") -> t.List[str]:
         if self.working_directory:
             working_directory = self.working_directory
 
-        if not (working_directory in serialized_mount):
+        if not working_directory in serialized_mount:
```
billschereriii, line 149: should this be:
```python
if working_directory not in serialized_mount:
```

smartsim/settings/containers.py
```diff
@@ -140,14 +146,17 @@ def _container_cmds(self, default_working_directory: str = "") -> t.List[str]:
...
                 serialized_mount = ",".join([working_directory, serialized_mount])
             else:
                 serialized_mount = working_directory
             logger.warning(
-                f"Working directory not specified in mount: \n {working_directory}"
-                + "\nAutomatically adding it to the list of bind points"
+                (
+                    "Working directory not specified in mount: \n {}\n"
+                    "Automatically adding it to the list of bind points"
+                ),
```
billschereriii, line 158: parens unnecessary?

smartsim/settings/lsfSettings.py
```diff
@@ -347,16 +353,18 @@ def __str__(self) -> str:
...
                 cpus_per_flag_set = True
                 cpu_per_rs = int(run_arg_value)
                 if cpu_per_rs < db_cpus:
-                    msg = f"{cpu_per_rs_flag} flag was set to {cpu_per_rs}, "
-                    msg += f"but colocated DB requires {db_cpus} CPUs per RS. Automatically setting "
-                    msg += f"{cpu_per_rs_flag} flag to {db_cpus}"
+                    msg = (
+                        f"{cpu_per_rs_flag} flag was set to {cpu_per_rs}, but "
+                        + f"colocated DB requires {db_cpus} CPUs per RS. Automatically "
+                        + f"setting {cpu_per_rs_flag} flag to {db_cpus}"
```
billschereriii, line 366: delete +

smartsim/_core/entrypoints/colocated.py
```diff
@@ -54,7 +54,9 @@
 SIGNALS = [signal.SIGINT, signal.SIGTERM, signal.SIGQUIT, signal.SIGABRT]
 
 
-def handle_signal(signo: int, frame: t.Optional[FrameType]) -> None:
+def handle_signal(signo: int, _frame: t.Optional[FrameType]) -> None:
+    if not signo:
+        logger.warning("Received signal with no signo")
```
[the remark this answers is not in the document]
ankona: I think I ran out of creativity. Without that line, the linter would complain about unused parameters. It seems impossible to have the handler be called without a signal number, but it was the easiest way to avoid another suppression and does provide "real information"

I'm open to suggestions (e.g. pass to cleanup and actually inspect signal to understand whats happening)

smartsim/settings/lsfSettings.py
```diff
@@ -535,21 +545,21 @@ def set_queue(self, queue: str) -> None:
...
-    def _format_alloc_flags(self) -> None:
+    def format_alloc_flags(self) -> None:
         """Format ``alloc_flags`` checking if user already
         set it. Currently only adds SMT flag if missing
         and ``self.smts`` is set.
         """
 
         if self.smts:
-            if not "alloc_flags" in self.batch_args.keys():
+            if "alloc_flags" not in self.batch_args.keys():
```
billschereriii, line 555: This is a functional change, but I'm guessing that the new version is correct and the old version was wrong.

smartsim/settings/slurmSettings.py
```diff
@@ -301,9 +302,13 @@ def check_env_vars(self) -> None:
...
                 if preexisting_var is not None:
-                    msg = f"Variable {k} is set to {preexisting_var} in current environment. "
-                    msg += f"If the job is running in an interactive allocation, the value {v} will not be set. "
-                    msg += "Please consider removing the variable from the environment and re-run the experiment."
+                    msg = (
+                        f"Variable {k} is set to {preexisting_var} in current "
+                        + "environment. If the job is running in an interactive "
+                        + f"allocation, the value {v} will not be set. Please "
+                        + "consider removing the variable from the environment "
+                        + "and re-run the experiment."
```
billschereriii, line 310: delete +

smartsim/wlm/__init__.py
```diff
@@ -39,28 +39,30 @@ def detect_launcher() -> str:
...
         if "cobalt" in (qsub_version.stdout).lower():
             return "cobalt"
-    if (
-        which("bsub")
-        and which("jsrun")
-        and which("jslist")
-        and which("bjobs")
-        and which("bkill")
+    if all(
+        [which("bsub"), which("jsrun"), which("jslist"), which("bjobs"), which("bkill")]
```
billschereriii, line 53: Are the [] needed? You've removed them from many other places

smartsim/_core/entrypoints/colocated.py
ankona: I see a missing comma on line 213. Fantastic catch. I hate this construction with the joins, altogether. Replacing with simpler version that doesn't unnecessarily use the join:

```
logger.debug(
    (
        "\n\nColocated database information\n"
        "\n\tIP Address(es): {}"
        "\n\tCommand: {}\n\n"
        "\n\t# of Database CPUs: {}"
    ),
    ' '.join(ip_addresses + [lo_address]),
    ' '.join(cmd),
    db_cpus
)
```

smartsim/wlm/__init__.py
billschereriii: Applies here and to the next block

smartsim/wlm/slurm.py
```diff
@@ -114,7 +116,10 @@ def release_allocation(alloc_id: str) -> None:
     if returncode != 0:
         logger.error(f"Unable to revoke your allocation for jobid {str(alloc_id)}")
         logger.error(
-            "The job may have already timed out, or you may need to cancel the job manually"
+            (
+                "The job may have already timed out, or you may "
+                "need to cancel the job manually"
+            )
```
billschereriii, line 122: parens unnecessary?

billschereriii (changes requested):
Looking really good, and you've clearly put a ton of work into this!

I'm a bit squeamish about changing methods to staticmethods and about exposing private functions and member variables to users. Otherwise, changes look generally very good with exceptions as noted.

smartsim/_core/entrypoints/colocated.py
ankona: I can't recall the exact lint msg, but it was likely "too many locals" or something similar. I pulled the smallest amount out that I could that made sense to pass the test.

mellis13 (changes requested):
Overall great PR -- just a couple of questions and comments we should talk through.

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
billschereriii: We really need double parens here?

smartsim/_core/launcher/slurm/slurmLauncher.py
ankona: 👍

smartsim/_core/launcher/step/alpsStep.py
```diff
@@ -120,10 +121,10 @@ def _build_exe(self) -> t.List[str]:
...
         if self.run_settings.mpmd:
             return self._make_mpmd()
-        else:
-            exe = self.run_settings.exe
-            args = self.run_settings.exe_args
-            return exe + args
+
+        exe = self.run_settings.exe
+        args = self.run_settings.exe_args
+        return exe + args
```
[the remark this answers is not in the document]
ankona: It's actually concatenating two lists to end up with `return [exe, arg1, ... argN]`

smartsim/_core/launcher/step/lsfStep.py
ankona: `_make_mpmd()` on lsfStep (unlike alps/slurm/mpi) doesn't return anything. We're not disregarding the result here so it's safe. I would say it's a questionable LSF implementation, though out of the purview of this task to fix.

image

smartsim/_core/launcher/step/step.py
ankona: NotImplementedError makes the linter treat this class as abstract and it has some downstream effects. Since the non-typed code was loose, it results in either implementing an unnecessary method on Jsrun/Aprun/Local/etc Steps or swapping this exception type out.

I chose the exception route:


```
❯ make check-lint
************* Module smartsim._core.launcher.step.lsfStep
smartsim/_core/launcher/step/lsfStep.py:114:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'JsrunStep' (abstract-method)
************* Module smartsim._core.launcher.step.alpsStep
smartsim/_core/launcher/step/alpsStep.py:40:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'AprunStep' (abstract-method)
************* Module smartsim._core.launcher.step.localStep
smartsim/_core/launcher/step/localStep.py:35:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'LocalStep' (abstract-method)
************* Module smartsim._core.launcher.step.step
smartsim/_core/launcher/step/step.py:33:0: W0611: Unused SmartSimError imported from smartsim.error.errors (unused-import)
************* Module smartsim._core.launcher.step.slurmStep
smartsim/_core/launcher/step/slurmStep.py:109:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'SrunStep' (abstract-method)
************* Module smartsim._core.launcher.step.mpiStep
smartsim/_core/launcher/step/mpiStep.py:41:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class '_BaseMPIStep' (abstract-method)
smartsim/_core/launcher/step/mpiStep.py:149:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'MpiexecStep' (abstract-method)
smartsim/_core/launcher/step/mpiStep.py:167:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'MpirunStep' (abstract-method)
smartsim/_core/launcher/step/mpiStep.py:185:0: W0223: Method 'add_to_batch' is abstract in class 'Step' but is not overridden in child class 'OrterunStep' (abstract-method)
```

ankona: I created a follow-up ticket for this item to be reviewed further

smartsim/_core/launcher/stepInfo.py
ankona: 👍

smartsim/_core/launcher/taskManager.py
```diff
@@ -39,10 +39,9 @@
 from .util.shell import execute_async_cmd, execute_cmd
 
 logger = get_logger(__name__)
-verbose_tm = check_dev_log_level()
+VERBOSE_TM = check_dev_log_level()  # pylint: disable=invalid-name
 
-
-TM_INTERVAL = 1
+TM_INTERVAL = 1  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
ankona: Not sure of the exact path I took to get here, but the suppression was no longer necessary at all. Removed.

smartsim/_core/launcher/util/shell.py
ankona: This is a great thought. I'm not sure why I ended up looking at it, but I changed it specifically because [another error](https://github.com/CrayLabs/SmartSim/blob/6a27c451d93b5c304d8c2757bf9e4fc8fb13e384/smartsim/_core/launcher/util/shell.py#L126) has the same message. If two separate locations have the same message, I think it results in confusion as to the true source of the error.

I'm glad to change this back if you don't find that convincing.

smartsim/_core/utils/helpers.py
ankona: It results in a circular import

smartsim/database/orchestrator.py
ankona: unnecessary but gross to look at. it could appear as a "missing comma" without additional indentation. I'm sort of on a case by case basis on the parens unnecessary items, but I am not a fan.

I'm going with f-string to avoid annoying confusion about whether arg should be part of concat vs passed to logger.warning and whether or not the second line is the replacement...

smartsim/database/orchestrator.py
ankona: see last. done.

smartsim/database/orchestrator.py
ankona: 👍

smartsim/database/orchestrator.py
ankona: 👍

smartsim/entity/dbobject.py
ankona: 👍 - also f-string

smartsim/entity/model.py
ankona: 👍

smartsim/entity/model.py
ankona: 👍

smartsim/entity/model.py
ankona: linter complains if we have __eq__ without having __hash__. Since eq uses name as the unique matching part, hash is implemented to match that same uniqueness constraint.

smartsim/entity/model.py
```diff
@@ -577,14 +586,15 @@ def _append_db_model(self, db_model: DBModel) -> None:
...
 
-    def _append_db_script(self, db_script: DBScript) -> None:
+    def append_db_script(self, db_script: DBScript) -> None:
         if db_script.func and self.colocated:
             if not isinstance(db_script.func, str):
                 err_msg = (
                     "Functions can not be set from memory for colocated databases.\n"
+                    + f"Please convert the function named {db_script.name} "
+                    + "to a string or store it as a text file and add it to the "
+                    + "SmartSim Model with add_script."
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/entity/model.py
ankona: 👍

smartsim/entity/model.py
ankona: 👍

smartsim/ml/data.py
ankona: 👍

smartsim/ml/data.py
ankona: 👍

smartsim/_core/launcher/util/shell.py
billschereriii: Exact duplication in error message text is bad, so changing seems good in that case

smartsim/ml/data.py
ankona: I believe the default check here would look for a `__len__` method implementation and use that for the truthiness. `len(self)` was just repeating what the default accomplishes.

smartsim/ml/data.py
ankona: ditto on default behavior for truthiness evaluation.

smartsim/ml/data.py
ankona: zero reason to have `end` when i had `list_length`. removed!

smartsim/ml/tf/__init__.py
```diff
@@ -41,13 +41,17 @@
 
 except ImportError:  # pragma: no cover
     raise ModuleNotFoundError(
-        f"TensorFlow {TF_VERSION} is not installed. Please install it to use smartsim.tf"
+        f"TensorFlow {TF_VERSION} is not installed. "
+        + "Please install it to use smartsim.tf"
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/ml/tf/__init__.py
ankona: 👍

smartsim/ml/tf/data.py
ankona: 👍

smartsim/ml/tf/data.py
ankona: 👍

smartsim/ml/tf/data.py
ankona: 👍

smartsim/ml/tf/utils.py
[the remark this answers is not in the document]
ankona: We do have two testing paths that hit it: `def test_freeze_model(fileutils):` and `def save_tf_cnn(path, file_name):`

ankona: In the lambda, we're using whatever the `__call__` implementation on model is. There should be no difference between passing model and letting that function get executed or passing a lambda that does the same thing, but indirectly. 

Consider:

```
def do_thing(one_arg_func: Callable[[int], None]):
  one_arg_func(42)

f0 = tf.function(lambda x: model(x))
f1 = model

r0 = do_thing(f0)
r1 = do_thing(f1)
```
Both routes pass 42 to the function. f0 pushes 42 down into the lambda to execute `model(42)` but f1 just does `one_arg_fun(42)` (which is `f1(42)` and of course, that's `model(42)` also).

smartsim/ml/torch/data.py
ankona: 👍

smartsim/ml/torch/data.py
```diff
@@ -101,7 +102,8 @@ def __init__(self, **kwargs: t.Any) -> None:
         super().__init__(**kwargs)
         if dynamic:
             self.log(
-                "Static data generator cannot be started with dynamic=True, setting it to False"
+                "Static data generator cannot be started "
+                + "with dynamic=True, setting it to False"
```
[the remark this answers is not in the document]
ankona: 👍

smartsim/ml/torch/data.py
ankona: 👍

smartsim/ml/torch/data.py
ankona: fantastic catch! tys

smartsim/settings/containers.py
ankona: 👍

smartsim/settings/containers.py
```diff
@@ -73,11 +73,15 @@ def _containerized_run_command(self, run_command: str) -> str:
...
         raise NotImplementedError(
-            f"Containerized run command specification not implemented for this Container type: {type(self)}"
+            "Containerized run command specification not implemented for this "
+            + f"Container type: {type(self)}"
         )
 
 
 class Singularity(Container):
+    # pylint: disable=abstract-method
+    # todo: determine if _containerized_run_command should be abstract
```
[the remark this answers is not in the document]
ankona: It was not yet addressed. Ticket created

smartsim/settings/containers.py
ankona: yes it should

smartsim/settings/containers.py
ankona: replacing w/f-string

smartsim/settings/lsfSettings.py
ankona: 👍

smartsim/settings/lsfSettings.py
ankona: I think `not ITEM in COLLECTION` and `ITEM not in COLLECTION` are identical in their result, even if the latter is the preferred style.

smartsim/settings/mpiSettings.py
```diff
@@ -341,7 +340,7 @@ def __init__(
         super().__init__(exe, exe_args, "mpiexec", run_args, env_vars, **kwargs)
 
         completed_process = subprocess.run(
-            [self._run_command, "--help"], capture_output=True
+            [self._run_command, "--help"], capture_output=True, check=False
```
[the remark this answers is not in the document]
ankona: I used the ref: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/subprocess-run-check.html

Looks like something that's a "best practice" versus a requirement. We could disable it or leave it, whichever you prefer.

smartsim/settings/slurmSettings.py
ankona: 👍

smartsim/settings/slurmSettings.py
```diff
@@ -341,7 +346,9 @@ def format_comma_sep_env_vars(self) -> t.Tuple[str, t.List[str]]:
         fmt_exported_env = ",".join(v for v in exportable_env + key_only)
 
         for mpmd in self.mpmd:
-            compound_mpmd_env = {k: v for k, v in mpmd.env_vars.items() if "," in str(v)}
+            compound_mpmd_env = {
+                k: v for k, v in mpmd.env_vars.items() if "," in str(v)
```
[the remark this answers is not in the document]
ankona: It makes sense to keep comprehensions concise. Every rule needs an exception to prove it!

smartsim/wlm/__init__.py
ankona: all takes an iterable. without parens or brackets, i'd be giving it multiple arguments instead.

smartsim/wlm/slurm.py
ankona: 👍

smartsim/_core/launcher/stepInfo.py
ankona: love the branch reduction

smartsim/_core/_cli/utils.py
```diff
@@ -70,6 +76,7 @@ def pip_install(packages: t.List[str], end_point: t.Optional[str] = None, verbos
 
     if verbose:
         logger.info(f"Installing packages {packages}...")
+    # pylint: disable-next=consider-using-with
```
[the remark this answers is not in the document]
ankona: IMO, all suppression should be temporary. I'll create tickets

smartsim/_core/control/manifest.py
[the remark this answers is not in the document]
ankona: I think it's a combination of two factors.

First, we've added typehints which should tell users "do it this way" and gives them `make mypy` to find out if they're wrong before they run code.

Second, the main use cases now run through "strongly typed" accessors. For example:

```
@property
def models(self) -> t.List[Model]:
    """Return Model instances in Manifest

    :return: model instances
    :rtype: List[Model]
    """
    _models: t.List[Model] = [
        item for item in self._deployables if isinstance(item, Model)
    ]
    return _models
```
and
```
has_db_objects |= any(
    has_db_models(model) | has_db_scripts(model) for model in self.models
)

```

Since our code no longer passes every item in the collection and instead only sends the models, adding another check is redundant.

The old code did something like:

```
has_dbscripts = any(has_db_objects(x) for x in self._deployables)
```
which made it required that we check the type in each of the has_xxx methods.

smartsim/experiment.py
```diff
@@ -770,7 +772,7 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
             logger.error(e)
             raise
 
-    def summary(self, format: str = "github") -> str:
+    def summary(self, style: str = "github") -> str:
```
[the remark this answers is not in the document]
ankona: We're redefining a built-in which should always be frowned upon. We can avoid it but it has to be done.

smartsim/_core/launcher/step/step.py
```diff
@@ -73,7 +77,9 @@ def get_colocated_launch_script(self) -> str:
         # prep step for colocated launch if specifed in run settings
         script_path = self.get_step_file(script_name=".colocated_launcher.sh")
 
-        db_settings = self.run_settings.colocated_db_settings
+        db_settings = {}
+        if isinstance(self.step_settings, RunSettings):
```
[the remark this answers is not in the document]
ankona: 100%. Amazing catch.

reversed conditional to raise if we don't have run settings.

smartsim/settings/lsfSettings.py
billschereriii: What's the binding of `not`? Is it `not (item in collection)` or `(not item) in collection`?

smartsim/settings/base.py
billschereriii: If I remember correctly, there were two rounds of collapsing all() calls here, so I wanted to just double-check that it was equivalent

smartsim/_core/_cli/utils.py
ankona: here's a snippet showing matching outputs from the two different pieces of code.

```
import smartsim
import importlib

ss_path = smartsim.__path__[0]
il_path = importlib.util.find_spec("smartsim").submodule_search_locations[0]

print('ss_path', ss_path)
print('il_path', il_path)

# Output:
# ss_path /Users/ankona/code/ss/smartsim
# il_path /Users/ankona/code/ss/smartsim
```

Check out [this documentation](https://docs.python.org/3/library/importlib.html#:~:text=The%20purpose%20of%20the%20importlib,portable%20to%20any%20Python%20interpreter.) to assuage fears.

smartsim/_core/launcher/util/shell.py
ankona: I've reverted this. I think you're right that it's out of scope of passing the linter. If we need it, it's simple to do later, too

smartsim/ml/tf/utils.py
ankona: Sanity checking by stepping through the `freeze_model` unit test does demonstrate no issues:

image

I'm not that familiar with Keras, but [this documentation](https://keras.io/api/models/model/) does talk through creating Model subclasses by implementing the forward pass in `__call__.`

smartsim/settings/base.py
ankona: I added a couple tests where I could easily exercise the code for creating arguments.

For reference, here's the snippet of nested `all`s

```
if isinstance(exe_args, list):
  exe_args = copy.deepcopy(exe_args)
  plain_type = all(isinstance(arg, (str)) for arg in exe_args)
  if not plain_type:
      nested_type = all(
          all(isinstance(arg, (str)) for arg in exe_args_list)
          for exe_args_list in exe_args
      )
      if not nested_type:
          raise TypeError(
              "Executable arguments were not list of str or str"
          )
      return exe_args
  return exe_args
```
I added tests that exercise all the possibilities not covered in existing tests:

- exe_args is a list instead of a string.
    - `exe_args = [script, "--time", "3"]`
- exe_args is a list containing a non-string.
    - `exe_args = [script, "--time", 3]`
- exe_args is a list-of-lists containing a non-string (exercise nested alls):
    - `exe_args = [[script, "--time", 3]]`
- exe_args is a list-of-lists containing only strings (exercise nested alls):
    - `exe_args = [[script, "--time", "3"]]`

I also added a ticket to review inconsistent behaviors between adding exe_args in the run settings constructor & via `add_exe_args` method.

smartsim/settings/lsfSettings.py
ankona: Reverted change, added unit test `test_bsub_batch_alloc_flag_formatting_by_smt`, repeated change, verified same results.

smartsim/_core/_cli/utils.py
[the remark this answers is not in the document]
ankona: just above this bit, line 44/45 will prevent this:

```
if module_spec is None:
  raise SetupError("Could not import SmartSim") from None

```

`module_spec` will be None and the same result will occur as the prior implementation without catching the module-not-found or import error.

smartsim/experiment.py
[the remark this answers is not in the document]
ankona: Reverted this API break and created a new ticket

.github/workflows/run_tests.yml
ankona: Fixed!

ankona:
rebased feature branch on latest develop (post-CLI refactor)

ankona:
@mellis13 @billschereriii could you rescind any requests for changes if those things have been met?

mellis13 (approved):
Great PR -- thank you!  LGTM!

setup.py
```diff
@@ -176,7 +176,7 @@ def has_ext_modules(_placeholder):
     "dev": [
         "black>=20.8b1",
         "isort>=5.6.4",
-        "pylint>=2.6.0",
+        "pylint>=2.10.0",
```
billschereriii, line 179: Can we use a consistent version on pylint on all repos?

smartsim/_core/control/job.py
```diff
@@ -63,9 +63,12 @@ def __init__(
...
         self.entity = entity
         self.status = STATUS_NEW
-        self.raw_status: t.Optional[str] = None  # status before smartsim status mapping is applied
+        self.raw_status: t.Optional[
+            str
+        ] = None  # status before smartsim status mapping is applied
         self.returncode: t.Optional[int] = None
-        self.output: t.Optional[str] = None  # only populated if it's system related (e.g. a command failed immediately)
+        # output is only populated if it's system related (e.g. cmd failed immediately)
+        self.output: t.Optional[str] = None
```
billschereriii, line 71: minor stylistic: I like the approach you did to split this line far more than the change on the previous line

smartsim/entity/model.py
```diff
@@ -577,14 +576,15 @@ def _append_db_model(self, db_model: DBModel) -> None:
 
         self._db_models.append(db_model)
 
-    def _append_db_script(self, db_script: DBScript) -> None:
+    def append_db_script(self, db_script: DBScript) -> None:
```
billschereriii, line 579: I thought you were going to revert all cases of making private methods public?

billschereriii (approved):
LGTM! Thanks for all the hard work on this one

PR #312, opened 2023-07-03
Set Test Colo DB Pinning to the Intended Range

MattToast:
This test is running a colo database on 4 cpus but pins the process to the range of the first 5 cpus. This seems an odd choice as there is theoretically no benefit to `taskset`ing to more than 4 cpus. This was causing also test to fail as the `custom_pinning` key  in the run settings colocated db settings is was being compared to a str of `0,1,2,3` (i.e. the firsts 4 cpus).

Given this, it seems likely that the original author intended to set the `taskset` cpu range be the first 4 cpus. This PR changes the custom task set range provided in the test from `range(5) == '0,1,2,3,4'` to `range(4) == '0,1,2,3'`

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -55,12 +55,12 @@ jobs:
           cp -r doc-branch/docs/* ./docs/
           rm -rf doc-branch
 
-      - name: build documentation with docker
+      - name: Build documentation with docker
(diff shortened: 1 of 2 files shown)


mellis13 (approved):
Thanks for fixing this error in the nightly tests. LGTM

PR #313, opened 2023-07-03
Fix the dbmodel and dbscript test errors

mellis13:
In PR 304 (https://github.com/CrayLabs/SmartSim/pull/304), the DB model and script tests were updated to enable testing on workload managers.  However, these tests were not constrained to run on only one node and one process during the update.  As a result, each SmartSim ``model`` and ``Ensemble`` member is run with a generic run command (e.g. ``srun``) that results in multiple, identical clones of the test entity running on each compute node (e.g. in a generic slurm interactive allocation with three nodes, a ``srun`` command without ``-N`` and/or ``-n`` arguments will launch the test script on three compute nodes).  This oversight causes a combination of key collisions that results in sporadic test failures.  This PR addresses those test failures.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -25,6 +25,7 @@ Description
 
 A full list of changes and detailed notes can be found below:
 
+- Fix tests for db models and scripts
 - Fix add_ml_model() and add_script() documentation, tests, and code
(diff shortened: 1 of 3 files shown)


MattToast (approved):
LGTM!! Thanks for tracking down and the fast fix!

PR #318, opened 2023-07-06
Update SmartSim version numbers

ashao:
Bumps relevant version numbers in anticipation of release

### DIFF
--- a/.wci.yml
+++ b/.wci.yml
@@ -22,7 +22,7 @@
     language: Python
 
     release:
-      version: 0.4.2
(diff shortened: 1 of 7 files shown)


.wci.yml
```diff
@@ -22,7 +22,7 @@
     language: Python
 
     release:
-      version: 0.4.2
+      version: 0.5.0
```
mellis13, line 25: We should probably update the date below

mellis13 (approved):

PR #322, opened 2023-07-12
Ensemble documentation update

billschereriii:


doc/experiment.rst
```diff
@@ -259,4 +259,58 @@ the following:
...
   cd /scratch/smartsim/Training-Run ; /usr/bin/srun --output /scratch/smartsim/Training-Run/Training-Ensemble_3.out --error /scratch/smartsim/Training-Ensemble_3.err --job-name Training-Ensemble_3-CHTN0UI2TRE7 --nodes=1 --ntasks=24 /scratch/pyenvs/smartsim/bin/python ./train-model.py --LR=0.35000000000000003 &
 
-  wait
\ No newline at end of file
+  wait
+
+Prefixing Keys in the Orchestrator
+----------------------------------
+
+If each of multiple ensemble members attempt to use the same code access their respective models in
```
ankona, line 267: intro suggestion:

`Experiments often use fixed key names to access data in the orchestrator. Concurrently running models in an ensemble experience issues when key collisions exist. Ensure ensemble members use unique keys by enabling automatic key prefixing.`

doc/experiment.rst
```diff
@@ -259,4 +259,58 @@ the following:
...
+the Orchestrator, the keys by which they do this will overlap and they can end up accessing each
+others' models inadvertently. To prevent this situation, the SmartSim Entity object supports key
+prefixing, which automatically prepends the name of the model to the keys by which it is accessed.
+With this enabled, key overlapping is no longer an issue and ensemble members can use the same code.
```
mellis13, line 271: This is a good update and example.  One area where we might want to dive deeper is that users are getting tripped up on ``enable_key_prefixing`` only setting outbound key prefixing (i.e. ``SSKEYOUT``) and they would need to register incoming connections (including self) to be able to pull keys from the ensemble member namespace.  

Also, we should double check on the default client prefixing behavior with ``SSKEYOUT`` and ``SSKEYIN``.  I think there are some subtitles to prefixing behavior between tensors, models, datasets, aggregation lists, and scripts.  Maybe deeper content around this behavior belongs here and in the clients (or a link between the two)

doc/experiment.rst
```diff
@@ -259,4 +259,67 @@ the following:
...
+
+Under the hood, calling ensemble.enable_key_prefixing() causes the SSKEYOUT environment variable to
+be set, which in turn causes all keys generated by an ensemble member to be prefixed with its model
+name.  Similarly, if the model for the ensemble member has incoming entities, the SSKEYIN environment
```
mellis13, line 275: Maybe "Similarly, if the model for the ensemble member has incoming entities, the SSKEYIN environment variable will be set and the keys associated with those inputs will be automatically prefixed."  would be more instructive as "Similarly, if the model for the ensemble member has incoming entities set via ``model.register_incoming_entity`` or ``ensemble.register_incoming_entity()``, the SSKEYIN environment variable will be set and the keys associated with those inputs will be automatically prefixed.

doc/experiment.rst
```diff
@@ -259,4 +259,67 @@ the following:
...
+variable will be set and the keys associated with those inputs will be automatically prefixed.
```
mellis13, line 276: Maybe add a line at the end of this paragraph to clarify that a entity must register itself because that is not done by default.

mellis13 (changes requested):

doc/experiment.rst
```diff
@@ -259,4 +259,67 @@ the following:
...
-  wait
\ No newline at end of file
+  wait
+
+Prefixing Keys in the Orchestrator
+----------------------------------
+
+If each of multiple ensemble members attempt to use the same code access their respective models in
+the Orchestrator, the keys by which they do this will overlap and they can end up accessing each
+others' models inadvertently. To prevent this situation, the SmartSim Entity object supports key
```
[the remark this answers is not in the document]
billschereriii: yes -- good catch

doc/experiment.rst
billschereriii: Thanks, that will be much better for users.

doc/experiment.rst
billschereriii: done

ankona (approved):

mellis13 (approved):

PR #324, opened 2023-07-21
Raise error for inconsistent add_ml_model and add_script parameters

juliaputko:
Added error message for Model.add_ml_model and Model.add_script when devices_per_node parameter is >1 and device parameter is set to CPU. 

Test for add_ml_model and add_script added to ensure error is correctly thrown.


smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
...
+        if self.device in ["GPU"] and self.devices_per_node > 1:
             for device_num in range(self.devices_per_node):
                 devices.append(f"{self.device}:{str(device_num)}")
         else:
             devices = [self.device]
 
         return devices
 
+    def _check_arguments(self):
+        devices = []
```
ankona, line 121: unused var?

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
...
-        if self.device in ["CPU", "GPU"] and self.devices_per_node > 1:
+        if self.device in ["GPU"] and self.devices_per_node > 1:
             for device_num in range(self.devices_per_node):
                 devices.append(f"{self.device}:{str(device_num)}")
         else:
             devices = [self.device]
 
         return devices
 
+    def _check_arguments(self):
```
ankona, line 120: i love pulling complex validation out into separate methods but `_check_arguments` feels like it could be unclear *which* arguments are being checked (especially since it has none).

I'd rename this to `_check_init` to make it clear where it's valid to use. I'd also probably pass the inputs instead of allowing the object to maybe be partially initialized.

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
         :rtype: list[str]
         """
         devices = []
-        if ":" in self.device and self.devices_per_node > 1:
-            msg = "Cannot set devices_per_node>1 if a device numeral is specified, "
-            msg += f"the device was set to {self.device} and devices_per_node=={self.devices_per_node}"
-            raise ValueError(msg)
-        if self.device in ["CPU", "GPU"] and self.devices_per_node > 1:
+        if self.device in ["GPU"] and self.devices_per_node > 1:
             for device_num in range(self.devices_per_node):
```
ankona, line 113: I'm a big fan of declarative style. Consider a list comprehension instead of loops:

```
devices = [f"{self.device}:{index}" for index in range(self.devices_per_node)]
```

This also makes it clear we're _not_ doing multiple appends from multiple branches of the `if`


You could also refactor to avoid nesting and significantly shorten:

```
if self.device == "GPU" and self.devices_per_node > 1:
  return [f"{self.device}:{index}" for index in range(self.devices_per_node)]

return [self.device]
```

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
...
 
         return devices
 
+    def _check_arguments(self):
+        devices = []
+        if ":" in self.device and self.devices_per_node > 1:
+            msg = "Cannot set devices_per_node>1 if a device numeral is specified, "
+            msg += f"the device was set to {self.device} and devices_per_node=={self.devices_per_node}"
+            raise ValueError(msg)
+        if self.device in ["CPU"] and self.devices_per_node > 1:
```
ankona, line 126: no need for `in`?

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
...
             for device_num in range(self.devices_per_node):
                 devices.append(f"{self.device}:{str(device_num)}")
         else:
             devices = [self.device]
 
         return devices
 
+    def _check_arguments(self):
+        devices = []
+        if ":" in self.device and self.devices_per_node > 1:
```
ankona, line 122: if we do a check at the start:
```
if self.devices_per_node <= 1:
  return
```

We can avoid having compound conditionals below, making it easier to read what's required to trigger a failure.

tests/backends/test_dbscript.py
```diff
@@ -578,3 +578,60 @@ def test_db_script_errors(fileutils, wlmutils, mlutils):
     # an in-memory script
     with pytest.raises(SSUnsupportedError):
         colo_ensemble.add_model(colo_model)
+
+def test_inconsistent_params_add_script(fileutils, wlmutils, mlutils):
```
ankona, line 582: It looks like we're testing a _lot_ of extra code here. Aren't the new validators in a constructor?

Can we build tests around the smallest unit?

```
... 
# you could vary x, y, z and exercise the validation code 
# more directly instead of allowing other code to potentially break
with pytest.raises(SSUnsupportedError):
  dbscript = DBScript(x,y,z)

```

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
...
+            msg = "Cannot set devices_per_node>1 if a device numeral is specified, "
+            msg += f"the device was set to {self.device} and devices_per_node=={self.devices_per_node}"
+            raise ValueError(msg)
+        if self.device in ["CPU"] and self.devices_per_node > 1:
+            raise SSUnsupportedError(
+                "Cannot set devices_per_node>1 if CPU is specified under devices"
+            )
```
ankona, line 129: Do we need to check device?

```
if self.device not in ["CPU","GPU"]:
  raise ValueError("invalid device...")

File: smartsim/entity/dbobject.py [hunk 8]
```diff
@@ -53,6 +54,7 @@ def __init__(
             self.file = self._check_filepath(file_path)
         self.device = self._check_device(device)
         self.devices_per_node = devices_per_node
+        self._check_arguments()
```
ankona, line 57: Consider using `device: Literal["CPU", "GPU"]` type hint to tighten the arguments for device (instead of `str`) in `_check_device`/`__init__`

smartsim/entity/dbobject.py
ashao: I could see this as being a general sanity checker of the initialization. There's a bit of a tradeoff between cluttering up the constructor and breaking it out into a validator for just devices and number of devices.

As currently coded up (where it only relies on the set properties after initialization is completed), it's unambiguous that the settings are correct. Thoughts @juliaputko @ankona?

smartsim/entity/dbobject.py
```diff
@@ -107,18 +109,24 @@ def _enumerate_devices(self) -> t.List[str]:
         :rtype: list[str]
         """
         devices = []
-        if ":" in self.device and self.devices_per_node > 1:
-            msg = "Cannot set devices_per_node>1 if a device numeral is specified, "
-            msg += f"the device was set to {self.device} and devices_per_node=={self.devices_per_node}"
-            raise ValueError(msg)
-        if self.device in ["CPU", "GPU"] and self.devices_per_node > 1:
+        if self.device in ["GPU"] and self.devices_per_node > 1:
```
ashao, line 112: Just check against GPU (no `in` needed). I suppose too we could always make CPU and GPU something like an `enum` and attach them to `DBObject` so that we're comparing an object of some variety instead of a magic string

tests/backends/test_dbmodel.py
```diff
@@ -793,3 +793,37 @@ def test_colocated_db_model_errors(fileutils, wlmutils, mlutils):
 
     with pytest.raises(SSUnsupportedError):
         colo_ensemble.add_model(colo_model)
+
+def test_inconsistent_params_add_ml_model(fileutils, wlmutils, mlutils):
+    """Test error when devices_per_node parameter>1 when devices is set to CPU in addd_ml_model function"""
```
ashao, line 798: Typo `addd_ml_model`

tests/backends/test_dbscript.py
ashao: Good point especially because the `add_model`, `add_script`, `add_function` are essentially just factory methods for their respective `DBEntity`

ashao (changes requested):
Thanks for fixing this up! 

A few typos and some excellent suggestions from @ankona to address

smartsim/entity/dbobject.py
ankona: It looks like `_check_xxx` methods in the constructor follow the pattern of offloading per-field validation to another method. I wouldn't care if they're combined but `_check_parameters` is definitely improper naming.

ashao (approved):

PR #325, opened 2023-07-21
Raising error for reserved keywords under function parameter options in get_allocation

juliaputko:
Added error if a user calls get_allocation with reserved keywords (nodes, account, time) as options eg. get_allocation(options={"account"="ACCOUNT"}). 

Added test to ensure error is correctly thrown.


smartsim/wlm/slurm.py
```diff
@@ -253,6 +253,8 @@ def _get_alloc_cmd(nodes: int,
                     salloc_args += [prefix + opt, str(val)]
                 else:
                     salloc_args += ["=".join((prefix + opt, str(val)))]
+        else: 
+            raise SSConfigError("Expecting time, nodes, account as an argument")
```
MattToast, line 257: Incredibly minor point: The `SSConfigError` tends to be used for when smartsim cannot find `redis-server`, `redisai.so`, one of the ml backends, or some other binary file, and is usually thrown when a user is trying to run a an experiment, but forgot to run `smart build` or set their `SMARTSIM_DEP_INSTALL_PATH` or `RAI_PATH` env var.

I might suggest changing to a plain old `ValueError` (or maybe a `KeyError`, but that is kind of a stretch on the expected use of `KeyError`)

tests/full_wlm/test_slurm_allocation.py
```diff
@@ -52,6 +52,12 @@ def test_get_release_allocation_w_options(wlmutils):
     time.sleep(5)  # give slurm a rest
     slurm.release_allocation(alloc)
 
+@pytest.mark.parametrize("func_parameters,test_parameters", [("time","TIME"),("nodes","NODES"),("account","ACCOUNT")])
+def test_get_allocation_bad_params(func_parameters,test_parameters):
+    """test get_allocation with reserved keywords as option """
```
MattToast, line 57:
```suggestion
    """test get_allocation with reserved keywords as option"""
```
Doesn't even matter from a technical standpoint; the white space gets stripped off automatically at compile time. But it will drive me insane if we leave it, hahahaha

tests/full_wlm/test_slurm_allocation.py
```diff
@@ -52,6 +52,12 @@ def test_get_release_allocation_w_options(wlmutils):
...
+    options = {"ntasks-per-node": 1}
```
MattToast, line 58: Looks like `options` is unused, should be free to rm!

MattToast (approved):
LGTM pending a couple of knit picks and stripping the whitespace!!

smartsim/wlm/slurm.py
```diff
@@ -253,6 +253,8 @@ def _get_alloc_cmd(nodes: int,
                     salloc_args += [prefix + opt, str(val)]
                 else:
                     salloc_args += ["=".join((prefix + opt, str(val)))]
+        else:
+            raise ValueError("Expecting time, nodes, account as an argument")
```
ankona, line 257: Should we tighten the looping by shifting validation outside of it and only leaving "list building logic" inside?

e.g. 

```
arguments = {key for key in options.keys()}
valid = {"t", "time", "N", "nodes", "A", "account"}

if invalids := arguments - valid:
   raise ValueError("Expecting time, nodes, account as an argument. Also received: {invalids}")

```

tests/full_wlm/test_slurm_allocation.py
```diff
@@ -52,6 +52,11 @@ def test_get_release_allocation_w_options(wlmutils):
     time.sleep(5)  # give slurm a rest
     slurm.release_allocation(alloc)
 
+@pytest.mark.parametrize("func_parameters,test_parameters", [("time","TIME"),("nodes","NODES"),("account","ACCOUNT")])
+def test_get_allocation_bad_params(func_parameters,test_parameters):
+    """test get_allocation with reserved keywords as option"""
+    with pytest.raises(ValueError):
+       alloc = slurm.get_allocation(options={func_parameters:test_parameters})
```
ankona, line 59: We might want to make sure it's really the error we expected to be thrown:

```
with pytest.raises(ValueError) as ex:
    alloc = slurm.get_xxx(...)

assert 'xyz' in ex.details.msg[0]
```

smartsim/wlm/slurm.py
```diff
@@ -253,6 +253,8 @@ def _get_alloc_cmd(nodes: int,
                     salloc_args += [prefix + opt, str(val)]
                 else:
                     salloc_args += ["=".join((prefix + opt, str(val)))]
+        else:
```
mellis13, line 256: Line 239 of this file has a ``TODO``.  Can we capture that in a ticket and remove that ``TODO`` comment?

ashao (changes requested):
Very small change requested. Otherwise thanks for fixing this silent, but very weird piece of behavior!

tests/full_wlm/test_slurm_allocation.py
```diff
@@ -28,7 +28,7 @@
 
 import pytest
 
-from smartsim.error import AllocationError
+from smartsim.error import AllocationError, SSConfigError
```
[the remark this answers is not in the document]
MattToast: @ashao, my argument was that in the past the `SSConfigError` has been used for when SmartSim has not been correctly built (e.g. the user has not run `smart build`). For example:
```py
raise SSConfigError(
    "RedisAI dependency not found. Build with `smart` cli "
    "or specify RAI_PATH"
)
```

I'd be okay as creating a new, more appropriate error, but I'm going to strongly push against using `SSConfigError` as it is a subclass of an `SSInternalError` and we raise here due to bad user input, not due to something wrong within SmartSim:
```py
class SSInternalError(Exception):
    """
    SSInternalError is raised when an internal error is encountered.
    """

class SSConfigError(SSInternalError):
    """Raised when there is an error in the configuration of SmartSim"""
```

ankona (approved):
I see Matt's req for a ticket to handle the TODO, but otherwise looks great to me.

ashao (approved):
Looks great! Thank you for the change

PR #327, opened 2023-07-24
Avoid using `shell=True` 

ankona:
Remove explicit use of shell=True to avoid potential security issues


ashao (approved):
Looking at this in context, I don't see a good reason why this should have a `shell=True`.

PR #329, opened 2023-07-26
Alter launchers to pass env when starting a local step

ankona:
This ticket ensures that an experiment with a workload manager will pass environment variables to unmanaged steps. This edge case occurs if a developer manually creates and attach a `RunSettings` instance instead of using the `experiment.create_run_settings` factory method.

Changes include: 
- minor formatting (multi-line statements converted to single line)
- `LocalStep` environment is passed to task manager when task is unmanaged by launcher
- slurm-specific tests added to verify task run in popen inherits calling env

### DIFF
--- a/smartsim/_core/launcher/slurm/slurmLauncher.py
+++ b/smartsim/_core/launcher/slurm/slurmLauncher.py
@@ -149,17 +149,22 @@ class SlurmLauncher(WLMLauncher):
                 logger.debug(f"Gleaned batch job id: {step_id} for {step.name}")
 
         # Launch a in-allocation or on-allocation (if srun) command
+        elif isinstance(step, SrunStep):
+            task_id = self.task_manager.start_task(cmd_list, step.cwd)
         else:
-            if isinstance(step, SrunStep):
-                task_id = self.task_manager.start_task(cmd_list, step.cwd)
-            else:
-                # Mpirun doesn't direct output for us like srun does
-                out, err = step.get_output_files()
-                output = open(out, "w+", encoding="utf-8")  # pylint: disable=consider-using-with
-                error = open(err, "w+", encoding="utf-8")  # pylint: disable=consider-using-with
-                task_id = self.task_manager.start_task(
-                    cmd_list, step.cwd, out=output.fileno(), err=error.fileno()
-                )
+            # MPI/local steps don't direct output like slurm steps
+            out, err = step.get_output_files()
+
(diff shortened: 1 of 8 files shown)


smartsim/_core/launcher/slurm/slurmLauncher.py
```diff
@@ -149,17 +149,22 @@ def run(self, step: Step) -> t.Optional[str]:
...
-                out, err = step.get_output_files()
-                output = open(out, "w+", encoding="utf-8")  # pylint: disable=consider-using-with
-                error = open(err, "w+", encoding="utf-8")  # pylint: disable=consider-using-with
-                task_id = self.task_manager.start_task(
-                    cmd_list, step.cwd, out=output.fileno(), err=error.fileno()
-                )
+            # MPI/local steps don't direct output like slurm steps
+            out, err = step.get_output_files()
+
+            # LocalStep.run_command omits env, include it here
```
ashao, line 158: Suggest leaving a comment here that `None` means that the child process inherits the caller's environment by default

tests/on_wlm/test_local_step.py
```diff
@@ -0,0 +1,111 @@
...
+import uuid
+
+from smartsim import Experiment, status
+from smartsim.settings import RunSettings
+
+
+"""
+Test execution of local steps within the WLM
+"""
+
```
ashao, line 37: These tests should only run if the WLM is slurm only. Otherwise this will fail because `srun` will not be available

ashao (approved):
Small changes to ensure that the test doesn't fail on non-slurm machines.  Other question, do you want to reserve adding tests for the other WLMs for later? It's a little challenging because of the combinations of WLMs and run commands

Thanks for digging into this one!

tests/on_wlm/test_local_step.py
ankona: Great catch. Added file-level `pytest.mark.skip`

smartsim/_core/launcher/slurm/slurmLauncher.py
ankona: Love the idea, but I instead updated the doc string for `TaskManager.start_task` so the env parameter states this behavior. I think this could be more helpful than a comment:

image

ashao (approved):
Thanks for the changes! Let's hold off on adding PBS+aprun tests since it might be a little harder to detect the presence of aprun.

PR #331, opened 2023-08-01
Log sacct failures

ankona:
After experiencing a crash in the slurm database, tests began failing unexpectedly. This fix ensures that a warning will be logged if the database goes down again.


ashao (approved):
Thanks @ankona! This will help differentiate what might be a SmartSim problem versus a problem that might be a result of system defect.

mellis13 (approved):

PR #332, opened 2023-08-01
Fix bug in logging msg format string

ankona:
Fix another buggy format string in logging introduced during lint mitigation.


amandarichardsonn (approved):
LGTM!

PR #334, opened 2023-08-02
Apply typehints to `smartsim._core.launcher.step.*`

ankona:


smartsim/_core/launcher/step/alpsStep.py
```diff
@@ -120,24 +142,24 @@ def _build_exe(self) -> t.List[str]:
...
-        args = self.run_settings.exe_args
+        args = self.run_settings._exe_args  # pylint: disable=protected-access
         return exe + args
 
     def _make_mpmd(self) -> t.List[str]:
         """Build Aprun (MPMD) executable"""
 
         exe = self.run_settings.exe
-        exe_args = self.run_settings.exe_args
+        exe_args = self.run_settings._exe_args  # pylint: disable=protected-access
```
ankona, line 156: note to self: go back & look @ when exe_args was added (was it during mitigation of mypy?). if not in open use, consider changing the signature to avoid the t.Union[str, t.List[str]] and define `def set_exe_args`

smartsim/_core/launcher/step/lsfStep.py
```diff
@@ -163,6 +177,9 @@ def get_launch_cmd(self) -> t.List[str]:
         :rtype: list[str]
         """
         jsrun = self.run_settings.run_command
+        if not jsrun:
+            logger.warning("jsrun not found in PATH")
+            raise RuntimeError("Could not find jsrun in PATH")
```
ankona, line 182: Can we do this with a RunSettings (e.g. local on slurm)?

smartsim/_core/launcher/step/lsfStep.py
```diff
@@ -228,25 +255,33 @@ def _build_exe(self) -> t.List[str]:
...
 
+        distr_line: str = ""
+        preamble_lines: t.List[str] = []
+        all_preamble_lines: t.List[str] = []
+
         # Find launch_distribution command
-        preamble_lines = self.run_settings.mpmd_preamble_lines.copy()
-        distr_line = None
-        for line in self.run_settings.mpmd_preamble_lines:
+        if hasattr(self.run_settings, "mpmd_preamble_lines"):
```
ankona, line 268: I think the new _lsf_run_settings property pattern could replace this hasattr

smartsim/_core/launcher/step/slurmStep.py
```diff
@@ -192,38 +206,47 @@ def _set_alloc(self) -> None:
...
         """Build the executable for this step
 
         :return: executable list
         :rtype: list[str]
         """
-        if self.run_settings.mpmd:
-            return self._make_mpmd()
+        if srs := self._srun_settings(ignore=True):
+            if srs.mpmd:
+                return self._make_mpmd()
```
ankona, line 224: Consider
```
srs = self._srun_settings(ignore=True)
if srs and srs.mpmd:
   ...
```

ankona: used convenience wrapper instead to avoid repeating `if srs: return srs.mpmd`

smartsim/_core/launcher/step/alpsStep.py
```diff
@@ -49,13 +49,29 @@ def __init__(self, name: str, cwd: str, run_settings: AprunSettings) -> None:
...
         super().__init__(name, cwd, run_settings)
-        self.alloc = None
+        self.alloc: t.Optional[str] = None
         if not self.run_settings.in_batch:
             self._set_alloc()
 
     @property
-    def run_settings(self) -> AprunSettings:
-        return self.step_settings
+    def run_settings(self) -> RunSettings:
```
al-rigazzi, line 57: Do you think we should have a concise docstring for this kind of `@property` (mostly for the rendered docs)?

smartsim/_core/launcher/step/alpsStep.py
```diff
@@ -49,13 +49,29 @@ def __init__(self, name: str, cwd: str, run_settings: AprunSettings) -> None:
...
+        if isinstance(self.step_settings, RunSettings):
+            return self.step_settings
+        raise TypeError("Run settings must be of type RunSettings")
+
+    def _aprun_settings(self, ignore: bool = False) -> t.Optional[AprunSettings]:
```
al-rigazzi, line 62: Even though the method is private, I think the `ignore` argument is a bit cryptic. It could be either renamed to something more explicit (and yes, verbose) or documented. Or both.

Same comment applies to similar methods added in this PR

smartsim/_core/launcher/step/alpsStep.py
```diff
@@ -49,13 +49,29 @@ def __init__(self, name: str, cwd: str, run_settings: AprunSettings) -> None:
...
+        if isinstance(self.step_settings, AprunSettings):
+            return self.step_settings
+        if not ignore:
+            raise TypeError("Run settings must be of type AprunSettings")
+        return None
+
+    def _get_mpmd(self) -> t.List[RunSettings]:
+        """temporary convenience function to return a typed list
```
al-rigazzi, line 70: Can you capitalize the `t` in `temporary`? I think it is our convention for docstrings. This applies to similar functions added in this PR.

smartsim/_core/launcher/step/localStep.py
```diff
@@ -39,7 +40,9 @@ def __init__(self, name: str, cwd: str, run_settings: RunSettings):
 
     @property
     def run_settings(self) -> RunSettings:
-        return self.step_settings
+        if isinstance(self.step_settings, RunSettings):
+            return self.step_settings
+        raise TypeError("Run settings must be of type RunSettings")
```
al-rigazzi, line 45: Codecov is complaining that the added line is not covered by tests. Do you think it is worth to add tests for this and all other new lines which are not covered? Or should we disable coverage for them?

smartsim/_core/launcher/step/lsfStep.py
al-rigazzi: Do _this_ what?

smartsim/_core/launcher/step/pbsStep.py
```diff
@@ -49,8 +50,10 @@ def __init__(self, name: str, cwd: str, batch_settings: QsubBatchSettings) -> No
         self.managed = True
 
     @property
-    def batch_settings(self) -> QsubBatchSettings:
-        return self.step_settings
+    def batch_settings(self) -> BatchSettings:
+        if isinstance(self.step_settings, BatchSettings):
+            return self.step_settings
+        raise TypeError("Batch settings must be subtype of BatchSettings")
```
al-rigazzi, line 56: Should we add `f"but it is of type {type(self.step_settings}"` to the error message to help the user understand the error? This comment applies to similar `TypeError`s added in this PR.

al-rigazzi (changes requested):
Good stuff! I like the convenience functions and the improved readability of the code. I have left a couple of requests and some comments that I'd like to be addressed before merging.

smartsim/_core/launcher/step/lsfStep.py
ankona: @al-rigazzi Can we assume that a failure to find & expand jsrun is worth raising an exception that will stop the experiment?

smartsim/_core/launcher/step/alpsStep.py
ankona: Added `"""Get the run settings attached to this step"""`

smartsim/_core/launcher/step/alpsStep.py
ankona: Renamed to `ignore_type_mismatch`

smartsim/_core/launcher/step/pbsStep.py
ankona: Done

smartsim/_core/launcher/step/alpsStep.py
ankona: reviewed..

al-rigazzi (approved):
LGTM! Thanks for getting this in and addressing my concerns!

PR #336, opened 2023-08-04
Fetch torch at pip install

MattToast:
This PR is a rivial purposal to unify SmartSim optional dependency resolution to #309. This Pr divides the the optional ML dependencies into 2 categories (`[ml-cpu]` and `[ml-cuida]`) to add the additional context needed to know which version of torch to fetch at `pip install` time.

Similar to #309, `smart build` will continue to warn about ML package conflicts if RAI is built in a different context to the installed  ML packages. A new target to the CLI `smart test` has also been added to verify that a user's install of SmartSim is able to run a simple ML experiment.

Since this PR removes the ability to fetch python packages through pip, the `--only_python_packages` flag of `smart build` has been removed.

### DIFF
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -30,11 +30,9 @@ import sys
 import typing as t
 from pathlib import Path
 
-import pkg_resources
 from tabulate import tabulate
 
-from smartsim._core._cli.utils import (color_bool, pip_install,
-                                       SMART_LOGGER_FORMAT)
+from smartsim._core._cli.utils import color_bool, SMART_LOGGER_FORMAT
 from smartsim._core._install import builder
 from smartsim._core._install.buildenv import (
     BuildEnv,
@@ -42,6 +40,7 @@ from smartsim._core._install.buildenv import (
     Version_,
     Versioner,
     DbEngine,
+    VersionConflictError,
 )
 from smartsim._core._install.builder import BuildError
 from smartsim._core.config import CONFIG
@@ -49,43 +48,17 @@ from smartsim._core.utils.helpers import installed_redisai_backends
 from smartsim.error import SSConfigError
 from smartsim.log import get_logger
 
-SMART_LOGGER_FORMAT = "[%(name)s] %(levelname)s %(message)s"
 logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 # NOTE: all smartsim modules need full paths as the smart cli
 #       may be installed into a different directory.
 
 
-def _install_torch_from_pip(
-    versions: Versioner, device: str = "cpu", verbose: bool = False
-) -> None:
-    packages = []
-    end_point = None
-    device_suffix = ""
-
-    if sys.platform == "darwin":
-        if device == "gpu":
-            logger.warning("GPU support is not available on Mac OS X")
-        # The following is deliberately left blank as there is no
-        # alternative package available on Mac OS X
-        device_suffix = ""
-        end_point = None
-
-    # if we are on linux cpu, either CUDA or CPU must be installed
-    elif sys.platform == "linux":
-        end_point = "https://download.pytorch.org/whl/torch_stable.html"
-        if device in ["gpu", "cuda"]:
-            device_suffix = versions.TORCH_CUDA_SUFFIX
-        elif device == "cpu":
-            device_suffix = versions.TORCH_CPU_SUFFIX
-
-    packages.append(f"torch=={versions.TORCH}{device_suffix}")
-    packages.append(f"torchvision=={versions.TORCHVISION}{device_suffix}")
-
-    pip_install(packages, end_point=end_point, verbose=verbose)
+_TDeviceStr = t.Literal["cpu", "gpu"]
+_TPinningStr = t.Literal["==", "!=", ">=", ">", "<=", "<", "~="]
 
 
-def check_onnx_install(build_env: BuildEnv, versions: Versioner) -> None:
+def check_py_onnx_version(versions: Versioner) -> None:
     """Check Python environment for ONNX installation"""
     if not versions.ONNX:
         py_version = sys.version_info
@@ -98,36 +71,19 @@ def check_onnx_install(build_env: BuildEnv, versions: Versioner) -> None:
             msg += "1.2.5 or "
         msg += "1.2.7."
         raise SetupError(msg)
-    try:
-        if not build_env.check_installed("onnx", versions.ONNX):
-            msg = (
-                f"ONNX {versions.ONNX} not installed in python environment. "
-                f"Consider installing onnx=={versions.ONNX} with pip"
-            )
-            logger.warning(msg)
-        else:
-            logger.info(f"ONNX {versions.ONNX} installed in Python environment")
-    except SetupError as e:
-        logger.warning(str(e))
+    _check_packages_in_python_env(
+        {
+            "onnx": Version_(versions.ONNX),
+            "skl2onnx": Version_(versions.REDISAI.skl2onnx),
+            "onnxmltools": Version_(versions.REDISAI.onnxmltools),
+            "scikit-learn": Version_(getattr(versions.REDISAI, "scikit-learn")),
+        },
+    )
 
 
-def check_tf_install(build_env: BuildEnv, versions: Versioner) -> None:
+def check_py_tf_version(versions: Versioner) -> None:
     """Check Python environment for TensorFlow installation"""
-
-    try:
-        if not build_env.check_installed("tensorflow", versions.TENSORFLOW):
-            msg = (
-                f"TensorFlow {versions.TENSORFLOW} not installed in Python "
-                "environment. Consider installing tensorflow=="
-                f"{versions.TENSORFLOW} with pip"
-            )
-            logger.warning(msg)
-        else:
-            logger.info(
-                f"TensorFlow {versions.TENSORFLOW} installed in Python environment"
-            )
-    except SetupError as e:
-        logger.warning(str(e))
+    _check_packages_in_python_env({"tensorflow": Version_(versions.TENSORFLOW)})
 
 
 def check_backends_install() -> bool:
@@ -179,10 +135,11 @@ def build_database(
         database_builder.cleanup()
     logger.info(f"{database_name} build complete!")
 
+
 def build_redis_ai(
     build_env: BuildEnv,
     versions: Versioner,
-    device: str,
+    device: _TDeviceStr,
     use_torch: bool = True,
     use_tf: bool = True,
     use_onnx: bool = False,
@@ -190,7 +147,6 @@ def build_redis_ai(
     libtf_dir: t.Union[str, Path, None] = None,
     verbose: bool = False,
 ) -> None:
-
     # make sure user isn't trying to do something silly on MacOS
     if build_env.PLATFORM == "darwin" and device == "gpu":
         raise BuildError("SmartSim does not support GPU on MacOS")
@@ -208,33 +164,17 @@ def build_redis_ai(
     if not check_backends_install():
         sys.exit(1)
 
-    # Check for onnx and tf in user python environemnt and prompt user
-    # to download them if they are not installed. this should not break
-    # the build however, as we use onnx and tf directly from RAI instead
-    # of pip like we do PyTorch.
-    if use_onnx:
-        check_onnx_install(build_env, versions)
-    if use_tf:
-        check_tf_install(build_env, versions)
-
     # TORCH
-    if use_torch:
-        if torch_dir:
-            torch_dir = Path(torch_dir).resolve()
-            if not torch_dir.is_dir():
-                # we will always be able to find a torch version downloaded by
-                # pip so if we can't find it we know the user suggested a torch
-                # installation path that doesn't exist
-                logger.error("Could not find requested user Torch installation")
-                sys.exit(1)
-        else:
-            # install pytorch wheel, and get the path to the cmake dir
-            # we will use in the RAI build
-            install_torch(build_env, versions, device=device)
-            torch_dir = build_env.torch_cmake_path
+    if use_torch and torch_dir:
+        torch_dir = Path(torch_dir).resolve()
+        if not torch_dir.is_dir():
+            raise SetupError("Could not find requested user Torch installation")
 
+    # TF
     if use_tf and libtf_dir:
         libtf_dir = Path(libtf_dir).resolve()
+        if not libtf_dir.is_dir():
+            raise SetupError("Could not find requested user TF installation")
 
     build_env_dict = build_env()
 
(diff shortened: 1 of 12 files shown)


smartsim/_core/_cli/test.py
```diff
@@ -0,0 +1,207 @@
...
+            self._finalizer.detach()  # type: ignore[attr-defined]
+
+
+def execute(_args: argparse.Namespace, /) -> int:
+    backends = installed_redisai_backends()
+    try:
+        with _VerificationTempDir() as temp_dir:
+            test_install(
+                location=temp_dir,
+                port=None,  # TODO: allow users to pass as arg `--port`?
```
ankona, line 76: I'll steal from Matt and ask if you want to create a task for this and remove the TODO from the code

smartsim/_core/_install/builder.py
```diff
@@ -455,10 +470,10 @@ def build_from_git(self, git_url: str, branch: str, device: str = "cpu") -> None
         # get RedisAI dependencies
         dep_cmd = [
```
ankona, line 471: could we pull the right hand side of this out into a function and avoid some minor duplication?

```
dep_cmd = create_git_build_cmd(self.fetch_pt, self.fetch_tf, self.fetch_onnx, verbose=True)
build_cmd = create_git_build_cmd(self.torch, self.tf, self.onnx)

File: smartsim/_core/_install/builder.py [hunk 3]
```diff
@@ -309,9 +311,9 @@ def __init__(
         self.rai_install_path: t.Optional[Path] = None
 
         # convert to int for RAI build script
-        self.torch = 1 if build_torch else 0
-        self.tf = 1 if build_tf else 0   # pylint: disable=invalid-name
-        self.onnx = 1 if build_onnx else 0
+        self.torch: _TBoolInt = 1 if build_torch else 0
+        self.tf: _TBoolInt = 1 if build_tf else 0  # pylint: disable=invalid-name
+        self.onnx: _TBoolInt = 1 if build_onnx else 0
```
ankona, line 316: should we mark these attributes as protected now that we have the public `fetch_xxx` properties to avoid unintended access?

smartsim/_core/_install/builder.py
```diff
@@ -504,7 +519,9 @@ def build_from_git(self, git_url: str, branch: str, device: str = "cpu") -> None
         self.run_command(build_cmd, cwd=self.rai_build_path)
 
         self._install_backends(device)
-        if self.torch:
+        if self.torch and self.torch_dir:
```
ankona, line 522: would:
```
def user_supplied_backend(backend: str):
  if backend == "torch":
    return self.torch and self.torch_dir
  if backend == "...":
    ...
```
avoid needing comments like the one below?

MattToast: YES!! I should have known when I was adding that comment that this is probably a good block of code to isolate into its own fn 🤦

smartsim/_core/_install/buildenv.py
```diff
@@ -312,18 +345,39 @@ def as_dict(self, db_name: DbEngine = "REDIS") -> t.Dict[str, t.Any]:
...
+        def _format_custom_linux_torch_deps(
+            torchv: str, torchvisionv: str, arc: str
+        ) -> t.Tuple[str, ...]:
+            # The correct versions and suffixes were scraped from
+            # https://pytorch.org/get-started/previous-versions/
+            supported_py_versions = ("3.8", "3.9", "3.10")
+            return tuple(
+                itertools.chain.from_iterable(
+                    (
+                        "torch"
```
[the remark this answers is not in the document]
MattToast: I think it does! Plus it might be nicer to keep this isolated on the change that the links change, as opposed to needing to dive into some obscure `Versioner` method

smartsim/_core/_cli/test.py
MattToast: damn it, the consequences of my actions are catching up with me :(. Consider it made if/when this goes in!

smartsim/_core/_install/builder.py
MattToast: I like it! It also removes the need for the `_TBoolInt` type alias. Will add!

smartsim/_core/_install/builder.py
MattToast: Made them private! I don't think anyone is using these attrs outside of this class, but just in case, I added some read-only `build_*` properties in case someone ever needs to introspect whether or not a backend is going to be built, but removes the ability for them to modify it making the class just a bit more static.

smartsim/_core/_cli/cli.py
```diff
@@ -95,16 +96,22 @@ def default_cli() -> SmartCli:
...
+            site_execute,
+        ),
         MenuItemConfig(
             "clobber",
             "Remove all previous dependency installations",
-            clobber_execute
+            clobber_execute,
+        ),
+        MenuItemConfig(
+            "test",
```
al-rigazzi, line 112: This is a great idea! Nevertheless, I wouldn't want users to believe this is our official way of running tests. Can we change it's name to something more restrictive, such as `smoke-test`, `dry-run`, `mini-test`, or similar?

smartsim/_core/_cli/test.py
```diff
@@ -0,0 +1,215 @@
...
+    def __exit__(
+        self,
+        exc: t.Optional[t.Type[BaseException]],
+        value: t.Optional[BaseException],
+        tb: t.Optional[TracebackType],
+    ) -> None:
+        if not value:  # Yay, no error! Clean up as normal
+            super().__exit__(exc, value, tb)
+        else:  # Uh-oh! Better make sure this is not implicitly cleaned up
+            self._finalizer.detach()  # type: ignore[attr-defined]
```
al-rigazzi, line 75: Out of curiosity, what does the scope of `_finalizer` become once you detach it? Does it live until the main thread exits?

smartsim/_core/_cli/test.py
```diff
@@ -0,0 +1,215 @@
...
+    exp.generate(orc)
+    exp.start(orc)
+    try:
+        (client_addr,) = orc.get_address()
+        yield Client(address=client_addr, cluster=False)
+    finally:
+        exp.stop(orc)
+
+
+def _find_free_port() -> int:
```
al-rigazzi, line 138: Interesting approach. Maybe in the future we could think of implementing this for the test ports (in the test suite) too?

tests/test_cli.py
```diff
@@ -668,3 +670,73 @@ def mock_operation(*args, **kwargs) -> int:
...
+def _good_build(*args, **kwargs):
+    print("LGTM")
+
+
+def _bad_build(*args, **kwargs):
+    raise Exception
+
+
+@contextmanager
+def _totally_real_temp_dir(*a, **kw):
```
al-rigazzi, line 684: OK, so, I don't want to be the killjoy here, but as this is code that _might_ be read by users, I'd use more explanatory names for functions and variables, and I'd apply the same standard to mock values. I know, I know, I'm old.

al-rigazzi (changes requested):
Looks great! Only a couple of minor comments/requests, but I like this new approach (pending future discussions with the rest of the team).

smartsim/_core/_cli/test.py
```diff
@@ -0,0 +1,215 @@
...
+
+import numpy as np
+from smartredis import Client
+
+from smartsim import Experiment
+from smartsim._core.utils.helpers import installed_redisai_backends
+from smartsim._core._cli.utils import SMART_LOGGER_FORMAT
+from smartsim.log import get_logger
+
+SMART_LOGGER_FORMAT = "[%(name)s] %(levelname)s %(message)s"
```
MattToast, line 43: This assignment is shadowing the previous import

smartsim/_core/_cli/build.py
```diff
@@ -30,62 +30,35 @@
...
-                                       SMART_LOGGER_FORMAT)
+from smartsim._core._cli.utils import color_bool, SMART_LOGGER_FORMAT
 from smartsim._core._install import builder
 from smartsim._core._install.buildenv import (
     BuildEnv,
     SetupError,
     Version_,
     Versioner,
     DbEngine,
+    VersionConflictError,
```
ashao, line 43: Just double checking, `VersionConflictError` and `SetupError` are both here because they get used inside of `buildenv` too (which we're trying to keep 100% separated from all SmartSim code)?

smartsim/_core/_cli/build.py
```diff
@@ -208,33 +164,17 @@ def build_redis_ai(
...
-                sys.exit(1)
-        else:
-            # install pytorch wheel, and get the path to the cmake dir
-            # we will use in the RAI build
-            install_torch(build_env, versions, device=device)
-            torch_dir = build_env.torch_cmake_path
+    if use_torch and torch_dir:
+        torch_dir = Path(torch_dir).resolve()
+        if not torch_dir.is_dir():
+            raise SetupError("Could not find requested user Torch installation")
```
ashao, line 171: Print the `torch_dir` string too

smartsim/_core/_cli/build.py
```diff
@@ -285,124 +225,181 @@ def build_redis_ai(
...
-    backend_torch_path = f"{CONFIG.lib_path}/backends/redisai_torch"
-    device = "cpu"
-    if Path(f"{backend_torch_path}/lib/libtorch_cuda.so").is_file():
-        device = "gpu"
-    return device
+def check_py_torch_version(versions: Versioner, device: _TDeviceStr = "cpu") -> None:
+    """Check Python environment for TensorFlow installation"""
+
+    if BuildEnv.is_macos():
+        device_suffix = ""
```
ashao, line 232: For paranoia, throw a warning (at least) here if someone requests gpu on a Mac. Or maybe do it in the extras vehcking stage.

smartsim/_core/_cli/cli.py
```diff
@@ -95,16 +96,22 @@ def default_cli() -> SmartCli:
...
+            "Run a simple SmartSim experiment to cofirm that it is built correctly",
```
ashao, line 113: 'cofirm' -> 'confirm'

smartsim/_core/_cli/test.py
```diff
@@ -0,0 +1,215 @@
...
+        ],
+        name="FullyConnectedNetwork",
+    )
+    fcn.compile(
+        optimizer="adam", loss="sparse_categorical_crossentropy", metrics=["accuracy"]
+    )
+    model, inputs, outputs = serialize_model(fcn)
+
+    client.set_model(
+        "keras-fcn", model, "TF", device="CPU", inputs=inputs, outputs=outputs
```
ashao, line 166: For all these backend tests, let's add an optional flag to `validate`, `--device` to specify whether we're testing CPU or GPU. GPU is almost always where we run into trouble

ashao (changes requested):
Some small changes, with maybe the request to add `--device` to a `smart validate` being the one to require a small review

smartsim/_core/_cli/cli.py
MattToast: I thought there was a proposal to change it to `test` to mimic the gmake pattern of `make install; make test`. More than willing to revert to `smart validate` though!

tests/test_cli.py
MattToast: Hahaha, tbf this is absolutely a name I will regent the literally the next time I have to edit this test. Consider it changed to something reasonable!

smartsim/_core/_cli/test.py
MattToast: I'd be more than willing to add something like this to the test suite! It's defiantly not the most robust solution, but certainly should be "good enough" for our test suite!

I'll write up a ticket!!

smartsim/_core/_cli/build.py
MattToast: Exactly! The `buildenv` module needs to be imported into the `setup.py` which means it any imports in that module would need to be done through the `importlib` module which can be kind of a pain. As much as I don't love that `VersionConflictError` does not live in the `errors` module, I do think this is the lesser of two evils.

smartsim/_core/_cli/build.py
MattToast: Smart!!

smartsim/_core/_cli/build.py
MattToast: good call!

smartsim/_core/_cli/cli.py
MattToast: Spelling is hard...

smartsim/_core/_cli/test.py
MattToast: I'm a little worried about feature creep as this isn't directly related to moving the torch deps. Should we put this into this PR? Would you mind if I kicked this suggestion out into the [`--port` flag](https://github.com/CrayLabs/SmartSim/pull/336/files/ac4734676050a5866d9b1a7013e66a1993d8277f#diff-0f39699c06b8555c21a533e5b264ca4ffc290d50284859a3e1710d1a89f1304c) ticket previously surfaced?

smartsim/_core/_install/buildenv.py
```diff
@@ -333,7 +405,26 @@ def ml_extras_required(self) -> t.List[str]:
...
+                *common,
+                # osx
+                f'torch=={self.TORCH} ; sys_platform == "darwin"',
+                f'torchvision=={self.TORCHVISION} ; sys_platform == "darwin"',
+                # linux
+                *_format_custom_linux_torch_deps(
+                    self.TORCH, self.TORCHVISION, self.TORCH_CPU_SUFFIX.lstrip("+")
+                ),
+            ],
+            "ml-cuda": [
```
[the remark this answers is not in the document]
MattToast: I'm a little nervous about introducing a `pip install smartsim[ml-rocm]` optional deps without also introducing a `smart build --device rocm` target, even if it is possible to do via the `smart build --torch_dir=/path/to/rocm/torch/...`.

That said I don't have a super strong opinion here and could be convinced to add it. Thoughts from other reviewers (@ashao)?

smartsim/_core/_install/builder.py
```diff
@@ -453,17 +480,20 @@ def build_from_git(self, git_url: str, branch: str, device: str = "cpu") -> None
...
-            f"WITH_ORT={self.onnx}",
-            "VERBOSE=1",
-            self.binary_path("bash"),
-            str(self.rai_build_path / "get_deps.sh"),
-            str(device),
-        ]
+        dep_cmd = self._rai_build_env_prefix(
+            with_pt=self.build_torch,
+            with_tf=self.build_tf,
+            with_ort=self.build_onnx,
```
ankona, line 486: very nice pivot from the hardcoded strings!

smartsim/_core/_cli/test.py
MattToast: Other additional arguments were requested for the `validate` target. A `--port` was additionally added in d9316f595f099a8ee79c0dd042bdca8acaa7532d

smartsim/_core/_cli/test.py
MattToast: NVM, I ended up throwing something together in d9316f595f099a8ee79c0dd042bdca8acaa7532d and e1720e1d5ceb876ba2ccc2a602e8d20220122d39. You should now be able to validate against a GPU

smartsim/_core/_install/builder.py
MattToast: I'd love to say I came up with the idea, but honestly I was just following the advice of a very wise reviewer, hahaha!!

smartsim/_core/_cli/test.py
al-rigazzi: Lovely, thanks for the explanation.

smartsim/_core/_install/buildenv.py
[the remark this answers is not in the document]
al-rigazzi: OK, so, just triple-checking here: a user who wanted to use AMD GPUs would still be capable of doing so by passing `--torch_dir=/path/to/amd_version_of_torch`, correct?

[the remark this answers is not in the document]
MattToast: That is correct!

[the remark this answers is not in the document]
MattToast: Similarly users could also still independently install AMD GPU compatible versions of the torch python package deps via `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocmX.Y.Z`.

I just don't want to give ROCm the illusion of first class support if it is not yet supported to the extent that CPU/CUDA is currently

al-rigazzi (approved):
LGTM, thanks for taking the time of adding all those options!

smartsim/_core/_install/buildenv.py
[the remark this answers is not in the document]
al-rigazzi: Grazie

PR #338, opened 2023-08-07
Enable mypy generic-related checks

ankona:
Enable `disallow_any_generics` and `warn_return_any` in pyproject.toml and mitigate issues that were identified.

### DIFF
--- a/smartsim/_core/_install/buildenv.py
+++ b/smartsim/_core/_install/buildenv.py
@@ -105,41 +105,31 @@ class Version_(str):
 
     def __gt__(self, cmp: t.Any) -> bool:
         try:
-            if Version(self).__gt__(self._convert_to_version(cmp)):
-                return True
-            return False
+            return bool(Version(self).__gt__(self._convert_to_version(cmp)))
         except InvalidVersion:
             return super().__gt__(cmp)
 
     def __lt__(self, cmp: t.Any) -> bool:
         try:
-            if Version(self).__lt__(self._convert_to_version(cmp)):
-                return True
-            return False
+            return bool(Version(self).__lt__(self._convert_to_version(cmp)))
         except InvalidVersion:
             return super().__lt__(cmp)
 
     def __eq__(self, cmp: t.Any) -> bool:
         try:
-            if Version(self).__eq__(self._convert_to_version(cmp)):
-                return True
-            return False
+            return bool(Version(self).__eq__(self._convert_to_version(cmp)))
         except InvalidVersion:
             return super().__eq__(cmp)
 
     def __ge__(self, cmp: t.Any) -> bool:
         try:
-            if Version(self).__ge__(self._convert_to_version(cmp)):
-                return True
-            return False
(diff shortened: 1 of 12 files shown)


smartsim/ml/data.py
```diff
@@ -491,7 +496,9 @@ def update_data(self) -> None:
         if self.shuffle:
             np.random.shuffle(self.indices)
 
-    def _data_generation(self, indices: np.ndarray) -> t.Tuple[np.ndarray, np.ndarray]:
+    def _data_generation(
+        self, indices: np.ndarray  # type: ignore[type-arg]# type: ignore
```
MattToast, line 500: redundant `# type: ignore` here?

smartsim/_core/_install/buildenv.py
```diff
@@ -105,31 +105,41 @@ def patch(self) -> str:
 
     def __gt__(self, cmp: t.Any) -> bool:
         try:
-            return Version(self).__gt__(self._convert_to_version(cmp))
+            if Version(self).__gt__(self._convert_to_version(cmp)):
+                return True
```
MattToast, line 109: Just making sure I understand the reason for this change:

Is this bc mypy is unable to understand the types of the vendored version of `Version` and is thus resolving to `Any`? An then this is just an explicit way to cast to `bool`? If so I'm happy to let this through!

smartsim/_core/control/controller.py
```diff
@@ -618,7 +618,7 @@ def reload_saved_db(self, checkpoint_file: str) -> Orchestrator:
             if not self._jobs.actively_monitoring:
                 self._jobs.start()
 
-            return orc
+            return orc # type: ignore # noqa: no-any-return
```
MattToast, line 621: instead of the `# type: ignore`/`# noqa: ` comments here, can we instead to an explicit cast at ~line 603?

MattToast (approved):
A couple of small questions, but otherwise LGTM!!

smartsim/_core/_install/buildenv.py
ankona: To mypy, Version.__gt__ returned a `t.Any` so this was done to ensure that I return an actual boolean for `def __gt__(self, cmp: t.Any) -> bool:`

I now realize that maybe I should have looked for stubs for packaging?

smartsim/_core/control/controller.py
ankona: 🤦 I don't know what I was thinking. Done!

smartsim/ml/data.py
ankona: copypasta fail. tys

smartsim/_core/_install/buildenv.py
ankona: I realized I could just cast and avoid the extra silliness. Fixed~!

MattToast: ooh! Like that even more! Consider this resolved!!

PR #339, opened 2023-08-10
Fix colocated db preparation bug when using `JsrunSettings`

ankona:
A model fails to properly configure colocated db settings when the run settings are of type `JsrunSettings.` This bug was introduced when protected members were made public (due to linter noting public use) but the change was partially rolled back.

The model method `_set_colocated_db_settings` attempts the following:

```
        if hasattr(self.run_settings, "prep_colocated_db"):
            self.run_settings.prep_colocated_db(common_options["cpus"])
```

There are no run settings classes with an attribute of that name. `JsrunSettings` does have `_prep_colocated_db` (note protected/underscore). 

```
 def _prep_colocated_db(self, db_cpus: int) -> None:
        cpus_per_flag_set = False
        for cpu_per_rs_flag in ["cpu_per_rs", "c"]:
            if run_arg_value := self.run_args.get(cpu_per_rs_flag, 0):
                cpus_per_flag_set = True
                ...
```

This change fixes the incorrect call to a non-existent method by invoking `self.run_settings._prep_colocated_db(...)` instead.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -19,6 +19,7 @@ To be released at some future point in time
 
 Description
 
+- Fix colocated db preparation bug when using `JsrunSettings`
 - Log ignored error messages from `sacct`
 - Fix malformed logging format strings
 - Fix bug when user specify CPU and devices greater than 1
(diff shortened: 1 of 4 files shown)


MattToast (approved):
Good catch! LGTM, pending one change to the changelog

doc/changelog.rst
```diff
@@ -27,12 +28,14 @@ Description
...
 Detailed Notes
 
+- Fix a failure to execute `_prep_colocated_db` due to incorrect named attr check (PR507_)
 - Log errors reported from slurm WLM when attempts to retrieve status fail (PR331_, PR332_)
 - Fix incorrectly formatted positional arguments in log format strings (PR330_)
 - Add error if user calls get_allocation with reserved keywords in slurm get_allocation (PR325_)
 - Add error when user requests CPU with devices greater than 1 within add_ml_model and add_script (PR324_)
 - Update pylint dependency, update .pylintrc, mitigate non-breaking issues, suppress api breaks (PR311_)
 
+.. _PR507: https://github.com/CrayLabs/SmartSim/pull/507
```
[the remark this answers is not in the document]
ankona: d'oh! on it!

tests/test_colo_model_lsf.py
```diff
@@ -0,0 +1,232 @@
...
+
+        coloutils.setup_test_colo(
+            fileutils,
+            db_type,
+            exp,
+            db_args,
+            colo_settings = run_settings,
+        )
+
+    assert ex
```
MattToast, line 71: Will this `assert` ever be false? Surely the `pytest.raises` will throw if no exception is raised

MattToast (approved):
One small question, but otherwise my approval stands!

tests/test_colo_model_lsf.py
ankona: 100. removing!

PR #348, opened 2023-08-25
Fix test output paths

al-rigazzi:
This PR is a hotfix. In the DB Object tests, we were calling `exp.generate` for some ensembles, but that is not needed, as they are automatically generated in the exp path. While this PR fixes the test output path issue, there remains an issue with the generation of ensembles: the path should be honored. I will put up a different ticket for that issue.


al-rigazzi:
Branch is polluted, opening a new PR from fresh branch

PR #349, opened 2023-08-25
Remove ensemble generation from DB Object tests

al-rigazzi:
This PR is a quick bugfix. In the DB Object tests (bot DB scripts and DB models), we are currently calling `exp.generate` for some ensembles, but that is not needed, as they are automatically generated in the exp path.

While this PR fixes the test output path issue, there remains an issue with the generation of ensembles: the path should be honored. I will put up a different ticket for that issue.


MattToast (approved):
LGTM!! Thanks for chasing this down, this alone is going to make testing locally MUCH easier!!

PR #351, opened 2023-08-28
`smart validate` should not hang when error in TF process

MattToast:
Closes an unneeded `mp.Connection` causing `smart` to hang.


al-rigazzi (approved):
Good catch, LGTM

PR #352, opened 2023-08-29
Print attached files

al-rigazzi:
This PR adds the possibility of printing files attached to models and to ensembles.

### DIFF
--- a/smartsim/entity/ensemble.py
+++ b/smartsim/entity/ensemble.py
@@ -29,6 +29,8 @@ import typing as t
 from copy import deepcopy
 from os import getcwd
 
+from tabulate import tabulate
+
 from .._core.utils.helpers import init_default
 from ..error import (
     EntityExistsError,
@@ -275,6 +277,29 @@ class Ensemble(EntityList):
                 to_copy=to_copy, to_symlink=to_symlink, to_configure=to_configure
             )
 
+    def print_attached_files(self):
+        """Print a table with information about attached files
+
+        :returns: Table representation of attached files
+        :rtype: str
+        """
+        if not self.models:
+            print("The ensemble is empty, no files to show.")
+            return
+
+        values = []
+        for model in self.models:
+            if not model.files:
+                files = "No file attached to this model."
+            else:
+                files = str(model.files)
+
+            values.append([model.name, files])
+
+        table = tabulate(values, headers=["Model name", "Files"], tablefmt="grid")
+
+        print(table)
+
     @staticmethod
     def _set_strategy(strategy: str) -> StrategyFunction:
         """Set the permutation strategy for generating models within
@@ -335,7 +360,7 @@ class Ensemble(EntityList):
         backend: str,
         model: t.Optional[str] = None,
         model_path: t.Optional[str] = None,
-        device: t.Literal["CPU","GPU"] = "CPU",
+        device: t.Literal["CPU", "GPU"] = "CPU",
         devices_per_node: int = 1,
         batch_size: int = 0,
         min_batch_size: int = 0,
@@ -395,7 +420,7 @@ class Ensemble(EntityList):
         name: str,
         script: t.Optional[str] = None,
         script_path: t.Optional[str] = None,
-        device: t.Literal["CPU","GPU"] = "CPU",
+        device: t.Literal["CPU", "GPU"] = "CPU",
         devices_per_node: int = 1,
     ) -> None:
         """TorchScript to launch with every entity belonging to this ensemble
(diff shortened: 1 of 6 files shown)


smartsim/entity/ensemble.py
```diff
@@ -275,6 +277,29 @@ def attach_generator_files(
                 to_copy=to_copy, to_symlink=to_symlink, to_configure=to_configure
             )
 
+    def print_attached_files(self) -> None:
```
MattToast, line 280: Thoughts on adding another method that simply formats the str and then have this method be solely responsible for printing it? Something like:
```py
def attached_files_table(self) -> str:
    # essentially what you have for this method
    # but just return the `tabulate` str
    ...

def print_attached_files(self) -> None:
    print(self.attached_files_table())
```

The only reason I ask is that the former might more closely mimic something like `pandas.DataFrame.head()` in an IPython/Jupyter environment.

smartsim/entity/ensemble.py
```diff
@@ -275,6 +277,29 @@ def attach_generator_files(
...
+        """Print a table with information about attached files
+
+        :returns: Table representation of attached files
+        :rtype: str
```
MattToast, line 284: Looks like the typehint (`None`) and return type docstr (`str`) do not match

MattToast (changes requested):
Looks great!! Couple of minor knit-picks, but otherwise LGTM!!

smartsim/entity/ensemble.py
ankona: 100% agree w/Matt on splitting finding the attached fileset & printing it but not on example impl.

I'd consider a method that returns the values-of-interest only:
```

def attached files(self) -> t.List[str]:
  per_model_files = {}

  for model in self.models:
    per_model_files[model.name] = []
    if model.files:
        per_model_files[model.name].extend(model.files)

  return per_model_files
```

Now we've got a complete representation of the fileset but it's not commingled with any specific presentation choices. Otherwise, "the ensemble is empty..." makes me have to parse strings to determine if there were files instead of a truthiness check.

ankona: On second thought, adding the empty message to tabulate may obviate a need for the attached_files method completely since the dict-comprehension is only a single line....

I'd kind of like to build a contract that's an implementation of the visitor pattern and be consistent & clear in where we're offering a tabulating/display API
```
class Tabular(t.Protocol):
  def to_tabular(self) -> Tabular:
    ...
  def on_empty(self) -> str:
    ...

class EnsembleTable(Tabular):
  def to_tabular(self) -> Tabular:
    return {model.name: model.files or [] for model in self.models}
  @property
  def on_empty(self) -> str:
    return "No file attached to this model"

```
Now tabulate would be able to just consume a consistent display API:

```
def tabulate(tabular: Tabular) -> None:
  output = ""
  for thing in tabular:
    output += thing or tabular.empty_msg
```

tests/test_generator.py
```diff
@@ -198,6 +281,7 @@ def test_config_dir(fileutils):
     gen.generate_experiment(ensemble)
 
     assert osp.isdir(osp.join(test_dir, "test"))
+
     # assert False
```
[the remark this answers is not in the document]
al-rigazzi: Interesting catch! Done in next commit.

smartsim/entity/files.py
```diff
@@ -139,6 +140,22 @@ def _check_path(file_path: str) -> str:
...
+            values.append(["Copy", "\n".join(self.copy)])
+        if self.link:
+            values.append(["Symlink", "\n".join(self.link)])
+        if self.tagged:
+            values.append(["Configure", "\n".join(self.tagged)])
+
+        if not values:
+            return "No file attached to this entity."
+
+        return tabulate(values, headers=["Strategy", "File"], tablefmt="grid")
```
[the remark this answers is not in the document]
al-rigazzi: Yeah, it's also more consistent with the other tables we create. Done in next commit.

MattToast (approved):
LGTM!!

smartsim/entity/ensemble.py
```diff
@@ -275,6 +277,30 @@ def attach_generator_files(
...
+        """
+        if not self.models:
+            return "The ensemble is empty, no files to show."
+
+        values = []
+        for model in self.models:
+            values.append([model.name, model.attached_files_table])
+
+        table = tabulate(values, headers=["Model name", "Files"], tablefmt="grid")
```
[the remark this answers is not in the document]
al-rigazzi: Good catch!

PR #353, opened 2023-08-31
Add log to file of parameters

al-rigazzi:
This PR adds logging features for parameter values when generating entities or ensembles.

The log on the standard out is now optional and gives information about values set for each parameter in each model. The same information, combined with the affected files is now written to a file, `param_settings.txt` which is located in the experiment path.

The content of the file is of this type:
```
Model name: dir_test_0
File name                   Parameters
--------------------------  ---------------
dir_test/dir_test_0/in.atm  Name      Value
                            ------  -------
                            THERMO       10
                            STEPS        10

Model name: dir_test_1
File name                   Parameters
--------------------------  ---------------
dir_test/dir_test_1/in.atm  Name      Value
                            ------  -------
                            THERMO       10
                            STEPS        20

Model name: dir_test_2
File name                   Parameters
--------------------------  ---------------
dir_test/dir_test_2/in.atm  Name      Value
                            ------  -------
                            THERMO       20
                            STEPS        10

Model name: dir_test_3
File name                   Parameters
--------------------------  ---------------
dir_test/dir_test_3/in.atm  Name      Value
                            ------  -------
                            THERMO       20
                            STEPS        20


```


MattToast (commented):
Haven't had a chance to fully review yet, but I did have one thought I wanted to make sure I voiced

smartsim/_core/generation/generator.py
```diff
@@ -249,10 +262,50 @@ def _build_tagged_files(tagged: TaggedFilesHierarchy) -> None:
...
+
+        if used_params:
+            used_params_str = ", ".join(
+                [f"{name}={value}" for name, value in used_params.items()]
+            )
+            logger.log(
+                level=self.log_level,
+                msg=f"Configured model {entity.name} with params {used_params_str}",
+            )
+            with open(self.log_file, mode="a", encoding="utf-8") as logfile:
```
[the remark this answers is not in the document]
al-rigazzi: I am in favor of the additional ones, but I'd like to keep the experiment-wide one in the root, for a quicker inspection.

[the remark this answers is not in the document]
al-rigazzi: Added them.

smartsim/_core/generation/generator.py
```diff
@@ -249,10 +271,56 @@ def _build_tagged_files(tagged: TaggedFilesHierarchy) -> None:
...
+        :param entity: the model being generated
+        :type entity: Model
+        :param files_to_params: a dict connecting each file to its parameter settings
+        :type files_to_params: t.Dict[str, t.Dict[str, str]]
+        """
+        used_params: t.Dict[str, str] = {}
+        file_to_tables: t.Dict[str, str] = {}
+        for file, params in files_to_params.items():
+            used_params.update(params)
+            table = tabulate([[param, value] for param, value in params.items()],
```
MattToast, line 295: Seems like this is an unneeded cast to turn `dict_items` into a `list` when `tabulate` only requires an `Iterable`. This should be equivalent to simply
```py
tabulate(params.items(), headers=["Name", "Value"])
```
if you want to save the extra iteration

smartsim/_core/generation/generator.py
```diff
@@ -249,10 +271,56 @@ def _build_tagged_files(tagged: TaggedFilesHierarchy) -> None:
...
+        if used_params:
+            used_params_str = ", ".join(
+                [f"{name}={value}" for name, value in used_params.items()]
+            )
+            logger.log(
+                level=self.log_level,
+                msg=f"Configured model {entity.name} with params {used_params_str}",
+            )
+            file_table = tabulate(
+                    [[file, table] for file, table in file_to_tables.items()],
```
MattToast, line 308: Same cast from `dict_items` to `list` here. If you decide to change above, it might be worth changing here as well.

If you leave the above "as is", feel free to resolve this comment

smartsim/_core/generation/generator.py
MattToast: I could be convinced that the extra cast to `list[list[str]]` makes this a hair more readable, and `dict_items` should be evaluated lazily, so its not like there is a performance hit here.

Feel free to leave "as is" if you see fit

smartsim/_core/generation/generator.py
```diff
@@ -249,10 +271,56 @@ def _build_tagged_files(tagged: TaggedFilesHierarchy) -> None:
...
+                    headers=["File name", "Parameters"],
+                )
+            log_entry = f"Model name: {entity.name}" + "\n" + file_table + "\n\n"
+            with open(self.log_file, mode="a", encoding="utf-8") as logfile:
+                logfile.write(log_entry)
+            with open(join(entity.path, "param_settings.txt"),
```
MattToast, line 314: Incredibly wild corner case to consider here:
```py
exp = Experiment(...)
rs = exp.create_run_settings(...)
ens = experiment.create_ensemble(params=..., run_settings=rs)
ens.attach_generator_files(to_configure=['./param_settings.txt'])
```
will actually overwrite the generated file. I know I was the one who originally asked for this file, but don't like the idea of "banning" users from attaching files of a certain name. Given this corner case, on second thought, I think we should probably only generate the top level `param_settings.txt`

If we do decide to keep the per-entity `param_settings.txt`, we should instead explicitly raise an error if a user attempts to copy/symlink/generate a `param_settings.txt`

tests/test_generator.py
```diff
@@ -265,6 +267,43 @@ def test_multiple_tags(fileutils):
...
+    params = {"THERMO": [10, 20], "STEPS": [10, 20]}
+    ensemble = exp.create_ensemble("dir_test", params=params, run_settings=rs)
+    conf_file = get_gen_file(fileutils, "in.atm")
+    ensemble.attach_generator_files(to_configure=conf_file)
+
+    def not_header(line):
+        """you can add other general checks in here"""
+        if line.startswith("Generation start date and time:"):
+            return False  # ignore it
+        return True
```
MattToast, line 285: If you want to save a couple of lines
```py
return not line.startswith(...)
```

tests/test_modelwriter.py
```diff
@@ -144,3 +144,24 @@ def test_mw_error_2():
...
+
+    conf_path = get_gen_file(fileutils, "easy/marked/")
+
+    # copy confs to gen directory
+    dir_util.copy_tree(conf_path, test_dir)
+    assert path.isdir(test_dir)
+
+    # init modelwriter
+    writer = ModelWriter()
+    with pytest.raises(SmartSimError):
```
MattToast, line 164: I know this wasn't a decision made as part of this PR, but is there a reason the `make_missing_tags_fatal=True` raises a generic `SmartSimError` over a `ParameterWriterError`?

smartsim/_core/generation/generator.py
```diff
@@ -59,12 +65,18 @@ def __init__(self, gen_path: str, overwrite: bool = False) -> None:
...
+        :param gen_path: Path in which files need to be generated
+        :type gen_path: str
         :param overwrite: toggle entity replacement, defaults to False
         :type overwrite: bool, optional
+        :param verbose: Whether generation information should be logged to std out
+        :type verbose: bool, optional
         """
         self._writer = ModelWriter()
         self.gen_path = gen_path
+        self.log_file = join(gen_path, "param_settings.txt")
```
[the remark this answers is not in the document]
al-rigazzi: It makes sense, especially in reference to what I'm going to write as reply to another comment.

smartsim/_core/generation/generator.py
al-rigazzi: OK, this needs an in-depth discussion.
The top-level `param_settings.txt` file gets overwritten every time a generation is called. But as we allow partial generation of experiments, this scenario leads to an unexpected result:
```python
exp = Experiment(...)
rs1 = exp.create_run_settings(...)
rs2 = exp.create_run_settings(...)
ens1 = experiment.create_ensemble(params=..., run_settings=rs1)
ens1.attach_generator_files(...)
ens2 = experiment.create_ensemble(params=..., run_settings=rs2)
ens2.attach_generator_files(...)
exp.generate(ens1)  # the top-level file now contains info about ens1 parameters
exp.generate(ens2)  # the top-level file now contains info about ens2 parameters, but ens1 info was overwritten
```
This happens because we are overwriting the file instead of appending.
But if we appended, then there would be no way to overwrite the file if the experiment is re-generated, meaning we would keep appending for every run (even if the user chose `overwrite=True`, because we cannot interpret that correctly again due to partial generation being possible).

I suggest we:
- select a more unique name, such as `smartsim_params.txt` which should be unlikely to be chosen by users
- we make it a reserved file name and `raise ValueError` if the user really wants that file name.


Thoughts?

tests/test_modelwriter.py
al-rigazzi: Nope, I think this was an old decision.

smartsim/_core/generation/generator.py
MattToast: I would be okay with the purposed suggestion above!

Given the reasoning for the replace-not-append decision, I'm inclined to agree with the need for, and am once again in favor of, the per-entity parameter reporting file.

The chances of a name conflict on a `smartsim_params.txt` file seems incredibly unlikely, and if we can raise an exception about a potential file conflict prior to `exp.start(ens)`, that should save users a lot of debugging headache.

tests/test_modelwriter.py
MattToast: I thought so. Don't change it with this PR, but I may check with the team to see if there is interest to using the more specific error in this/a future release.

smartsim/_core/generation/generator.py
al-rigazzi: Done!

smartsim/_core/generation/generator.py
al-rigazzi: Done

smartsim/_core/generation/generator.py
```diff
@@ -249,10 +271,56 @@ def _build_tagged_files(tagged: TaggedFilesHierarchy) -> None:
...
+            )
+            logger.log(
+                level=self.log_level,
+                msg=f"Configured model {entity.name} with params {used_params_str}",
+            )
+            file_table = tabulate(
+                    [[file, table] for file, table in file_to_tables.items()],
+                    headers=["File name", "Parameters"],
+                )
+            log_entry = f"Model name: {entity.name}" + "\n" + file_table + "\n\n"
```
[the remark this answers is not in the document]
al-rigazzi: Done

tests/test_generator.py
al-rigazzi: Let's save lines, in case we ever print everything to paper.

smartsim/_core/generation/generator.py
```diff
@@ -129,7 +141,17 @@ def _gen_exp_dir(self) -> None:
...
+                level=self.log_level, msg="Working in previously created experiment"
+            )
+
+        # The log_file only keeps track of the last generation
+        # this is to avoid gigantic files in case the user repeats
+        # generation several times. The information is anyhow
+        # redundant, as it is also written in each entity's dir
+        with open(self.log_file, mode= 'w', encoding='utf-8') as log_file:
+            dt_string = datetime.now().strftime("%d/%m/%Y %H:%M:%S")
+            log_file.write("Generation start date and time: " + dt_string + "\n")
```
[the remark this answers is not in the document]
al-rigazzi: Done

MattToast (approved):
LGTM!!

smartsim/_core/generation/generator.py
al-rigazzi: Added exception at generation level!

PR #356, opened 2023-09-07
Oss517

ankona:


smartsim/_core/control/controller.py
```diff
@@ -512,19 +598,26 @@ def _create_batch_job_step(
                 "EntityList must have batch settings to be launched as batch"
             )
 
+        telemetry_dir = telemetry_dir / entity_list.name
         batch_step = self._launcher.create_step(
             entity_list.name, entity_list.path, entity_list.batch_settings
         )
+        batch_step.meta["entity_type"] = str(type(entity_list).__name__).lower()
+        batch_step.meta["exp_path"] = self.exp_path
```
MattToast, line 606: I'm not sure that I like idea of `exp_path` being an attr on the `Controller` that is only ever set to a "proper" value in `_launch`. Not that would ever expect someone to do something like
```py
Controller()._create_job_step(model, "path/to/a/dir")
```
but it _could_ be done.

I also can't find a good way to remove the attr either tho that doesn't just devolve into just param diving. If you can find a good way to remove it that would be much appreciated, but otherwise don't sweat it and we can add this onto the list of things to address post this release.

MattToast (changes requested):
Throwing up some small initial feedback for now while I continue to look over the TM and the indirect in more detail. So far everything seems to be working great!!

smartsim/_core/control/controller.py
```diff
@@ -250,6 +309,19 @@ def get_jobs(self) -> t.Dict[str, Job]:
         with JM_LOCK:
             return self._jobs.completed
 
+    def get_db_jobs(self) -> t.Dict[str, t.Tuple[Job, t.Union[DBNode, Orchestrator]]]:
```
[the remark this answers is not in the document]
ankona: It was used by the buggy auto-shutdown. It may/may not be removable post-fix.

smartsim/_core/control/job.py
```diff
@@ -27,10 +27,46 @@
...
 import typing as t
 
-from ...entity import SmartSimEntity, EntitySequence
+from collections import namedtuple
+
+from ...entity import SmartSimEntity, EntityList, EntitySequence
 from ...status import STATUS_NEW
 
 
+_JobKey = namedtuple("_JobKey", ["step_id", "task_id"])
```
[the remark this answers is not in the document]
ankona: NP. Done!

smartsim/_core/launcher/local/local.py
```diff
@@ -34,6 +36,8 @@
 from ..stepInfo import UnmanagedStepInfo, StepInfo
 from ..stepMapping import StepMapping
 from ..taskManager import TaskManager
+from ...._core.utils import serialize
+from ...._core.utils.helpers import encode_cmd
```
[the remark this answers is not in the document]
ankona: done!

tests/test_indirect.py
```diff
@@ -0,0 +1,201 @@
...
+            pytest.param("indirect.py -t orchestrator", {"+c", "+t", "+n", "+d"}, id="etype typo"),
+            pytest.param("indirect.py -n expname", {"+c", "+t", "+n", "+d"}, id="name typo"),
+            pytest.param("indirect.py -d /foo/bar", {"+c", "+t", "+n", "+d"}, id="dir typo"),
+            pytest.param("indirect.py        +t ttt +d ddd +n nnn", {"+c", "+t", "+n", "+d"}, id="no cmd"),
+            pytest.param("indirect.py +c ccc        +d ddd +n nnn", {"+c", "+t", "+n", "+d"}, id="no etype"),
+            pytest.param("indirect.py +c ccc +t ttt +d ddd       ", {"+c", "+t", "+n", "+d"}, id="no name"),
+            pytest.param("indirect.py +c ccc +t ttt        +n nnn", {"+c", "+t", "+n", "+d"}, id="no dir"),
+        ]
+)
+def test_parser(capsys, cmd, missing):
```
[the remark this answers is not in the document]
ankona: absolutely not. fixed that test case and got rid of the unused +n argument

smartsim/_core/control/job.py
```diff
@@ -27,10 +27,46 @@
...
+        self.status_dir: str = ""
+
+    @property
+    def is_db(self) -> bool:
+        return self.type in ["orchestrator", "dbnode"]
+
+    @property
+    def is_managed(self) -> bool:
+        if self.step_id == "None":
+            return False
```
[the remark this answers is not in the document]
ankona: I added the check to _hydrate_persistable previously and forgot to remove this one. Removed. Good catch!

smartsim/_core/control/jobmanager.py
```diff
@@ -185,10 +188,26 @@ def add_job(
...
+            self.db_jobs[entity.name] = job
         else:
             self.jobs[entity.name] = job
 
+        for hook in self.on_start_hook:
+            hook(job, logger)
+
+    def add_job_onstart_callback(self, hook: t.Callable[[Job, Logger], None]) -> None:
+        if not hook in self.on_start_hook:
+            self.on_start_hook.append(hook)
```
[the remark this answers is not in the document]
ankona: done!

smartsim/_core/launcher/local/local.py
```diff
@@ -43,10 +47,10 @@ class LocalLauncher(Launcher):
 
     @property
     def supported_rs(self) -> t.Dict[t.Type[SettingsBase], t.Type[Step]]:
-       return {
+        return {
```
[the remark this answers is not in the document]
ankona: looks like it slipped through during the typehinting PR. removed it and shifted the abstract method declaration to `WLMLauncher`

smartsim/_core/control/job.py
```diff
@@ -27,10 +27,46 @@
...
+        self,
+        entity: t.Optional[t.Union[SmartSimEntity, EntityList[SmartSimEntity]]] = None,
+    ) -> None:
+        self.name: str = entity.name if entity else ""
+        self.path: str = entity.path if entity else ""
+        self.step_id: str = ""
+        self.task_id: str = ""
+        self.type: str = type(entity).__name__.lower() if entity else ""
+        self.timestamp: int = 0
+        self.status_dir: str = ""
```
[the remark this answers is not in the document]
ankona: It was a leftover from the intitial idea of using the job manager to create the JobEntities using the "original entities" but is totally unnecessary now. Great :eye:!

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+            time.sleep(0.5 * try_count)
+        finally:
+            try_count += 1
+
+    if not manifest_dict:
+        return None
+
+    exp = manifest_dict.get("experiment", None)
+    if not exp:
+        raise ValueError("Manifest missing required experiment")
```
MattToast, line 228: Not wrong just curious: any reason the first invalid state returns a null-ish value while the second one raises an exception?

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+    ) -> None:
+        super().__init__(
+            [pattern], ignore_patterns, ignore_directories, case_sensitive
+        )  # type: ignore
+        self._logger = logger
+        self._tracked_runs: t.Dict[int, Run] = {}
+        self._tracked_jobs: t.Dict[_JobKey, JobEntity] = {}
+        self._completed_jobs: t.Dict[_JobKey, JobEntity] = {}
+        self._launcher_type: str = ""
+        self._launcher: t.Optional[Launcher] = None
```
MattToast, line 337: Not wrong, but do we strictly need both a launcher and a launcher type? Do we need to know the launcher type before setting the launcher itself?

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+            self._logger.error(f"Malformed manifest encountered: {manifest_path}")
+            return
+        except ValueError:
+            self._logger.error("Manifest content error", exc_info=True)
+            return
+
+        self.set_launcher(manifest.launcher)
+
+        if not self._launcher:
+            raise SmartSimError(f"Unable to set launcher from {manifest_path}")
```
MattToast, line 404: Not wrong, but theoretically un-reachable code block here. `set_launcher` -> `self._launcher = self.init_launcher(...)` and `self.init_launcher(...)` will always return a non-nullish value.

Willing to keep this block around for safety though

smartsim/_core/control/controller.py
```diff
@@ -70,12 +80,55 @@
     from ..launcher.stepMapping import StepMap
     from ..utils.serialize import TStepLaunchMetaData
 
+    _TELEMETRY_MONITOR: t.Optional[subprocess.Popen[t.Any]] = None
```
MattToast, line 83: I'm like 95% confident that all of our `Popen`s are `str` based, if you want to tighten up this typehint

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,687 @@
...
+
+    if detail is not None:
+        entity_dict["detail"] = detail
+
+    if return_code is not None:
+        entity_dict["return_code"] = return_code
+
+    try:
+        if not tgt_path.exists():
+            """Don't overwrite existing tracking files"""
```
MattToast, line 284: One of my more pedantic requests:
```suggestion
            # Don't overwrite existing tracking files
```
I'm pretty sure that string gets optimized away at compile time, so it doesn't even save an alloc. Just felt better to have it as a comment hahahaha

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,687 @@
...
+            self.job_manager.set_launcher(self._launcher)
+            self.job_manager.add_job_onstep_callback(track_timestep)
+
+            self.job_manager.start()
+
+    @property
+    def launcher(self) -> Launcher:
+        """Return a launcher appropriate for the experiment"""
+        if not self._launcher:
+            self._launcher = LocalLauncher()
```
MattToast, line 381: Should this also be calling `self.set_launcher("local")` just so that the JM launcher and the `self._launcher` attr do not go "out-of-sync"?

MattToast: @MattToast probably worth writing up a ticket to expose the JM's launcher as a property so that things that share a reference to a launcher (Controller/TM/etc.) can also have their own launcher property just expose the underlying launcher used by the JM

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,687 @@
...
+                    )
+            self._tracked_runs[run.timestamp] = run
+
+    def on_modified(self, event: FileModifiedEvent) -> None:
+        """Event handler for when a file or directory is modified.
+
+        :param event: Event representing file/directory modification.
+        :type event: FileModifiedEvent"""
+        super().on_modified(event)  # type: ignore
+        print(f"processing modified manifest @ {event.src_path}")
```
MattToast, line 445: super pedantic "can we replace `print` with `logger`" request here

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,687 @@
...
+        if step_info:
+            detail = f"status: {step_info.status}, error: {step_info.error}"
+        else:
+            detail = "unknown status. step_info not retrieved"
+
+        if hasattr(job.entity, "status_dir"):
+            write_path = pathlib.Path(job.entity.status_dir)
+        else:
+            # todo: raise exception to figure out if this gets hit
+            write_path = pathlib.Path(job.entity.path)
```
MattToast, line 490: `TODO` still needed?

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,687 @@
...
+    if observer is None:
+        observer = Observer()
+
+    try:
+        if manifest_path.exists():
+            # a manifest may not exist depending on startup timing
+            action_handler.process_manifest(str(manifest_path))
+
+        observer.schedule(log_handler, experiment_dir, recursive=True)  # type:ignore
+        observer.schedule(action_handler, experiment_dir, recursive=True)  # type:ignore
```
MattToast, line 615: I'm guessing these suppressions are all coming from "untyped fn calls in typed context" errors. If you want feel free to just disable at a global level!

MattToast (changes requested):
Looks great!! I threw some more pedantic notes for the TM and the indirect, but so far everything seems to be being handled correctly on my end! Feel free to let me know what you think!!

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+        entity.task_id = ""
+
+    return entity
+
+
+def hydrate_persistable(
+    entity_type: str,
+    persistable_entity: t.Dict[str, t.Any],
+    exp_dir: pathlib.Path,
+# ) -> t.Dict[str, t.List[JobEntity]]:
```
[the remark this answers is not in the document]
MattToast: Outdated! You beat me to the punch before I could post this review!

smartsim/_core/entrypoints/indirect.py
```diff
@@ -0,0 +1,221 @@
...
+            process.terminate()
+
+    except psutil.NoSuchProcess:
+        # swallow exception to avoid overwriting outputs from cmd
+        ...
+
+    except OSError as ex:
+        logger.warning(f"Failed to clean up step executor gracefully: {ex}")
+    finally:
+        STEP_PID = 0
```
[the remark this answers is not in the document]
ankona: Swapped to None

smartsim/_core/launcher/local/local.py
```diff
@@ -85,6 +89,11 @@ def get_step_update(self, step_names: t.List[str]) -> t.List[t.Tuple[str, t.Opti
     def get_step_nodes(self, step_names: t.List[str]) -> t.List[t.List[str]]:
         """Return the address of nodes assigned to the step
 
+        :param step_names: list of step_names
+        :type step_names: list[str]
+        :return: list of node addresses
+        :rtype: list[str]
+
         TODO: Use socket to find the actual Lo address?
         :return: a list containing the local host address
```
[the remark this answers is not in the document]
ankona: fixed!

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: Agree 100. Maybe copilot created this monstrosity for me!? Yeah, blame the tool! Fixed

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: debug code has feelings too!

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: I'm divided.... I think I'd rather leave it on globally and catch missing stubs.

smartsim/_core/entrypoints/indirect.py
```diff
@@ -0,0 +1,221 @@
...
+        "+n", type=str, help="The step name being executed", required=True
+    )
+    parser.add_argument(
+        "+w", type=str, help="The working directory of the executable", required=True
+    )
+    parser.add_argument("+o", type=str, help="Output file", required=True)
+    parser.add_argument("+e", type=str, help="Erorr output file", required=True)
+    parser.add_argument(
+        "+d", type=str, help="Directory for telemetry output", required=True
+    )
```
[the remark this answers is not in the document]
ankona: added long args

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: You can try to load an empty file or maybe a file that doesn't exist. That's expected behavior (return None)

I figured that if I load a file and it has content that is not in the right format, I cannot recover so I throw.

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: removed

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+    entity.step_id = str(metadata.get("step_id", ""))
+    entity.task_id = str(metadata.get("task_id", ""))
+    entity.timestamp = int(persistable_entity.get("timestamp", "0"))
+    entity.path = str(exp_dir)
+    entity.status_dir = str(status_dir)
+
+    if entity.step_id == "None":
+        entity.step_id = ""
+    if entity.task_id == "None":
+        entity.task_id = ""
```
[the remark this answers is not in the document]
ankona: done

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+        "model": [],
+        "orchestrator": [],
+        "ensemble": [],
+    }
+    for item in run[entity_type]:
+        entities = hydrate_persistable(entity_type, item, exp_dir)
+        for new_entity in entities:
+            persisted[new_entity.type].append(new_entity)
+
+    return persisted
```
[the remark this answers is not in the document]
ankona: It depends on how we do child entities... a colo model might create a model and a orchestrator, or an ensemble might create an ensemble and a set of models.

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+    """Map run data persisted in a manifest file to an object"""
+    the_runs: t.List[Run] = []
+    for run_instance in persisted_runs:
+        run_entities: t.Dict[str, t.List[JobEntity]] = {
+            "model": [],
+            "orchestrator": [],
+            "ensemble": [],
+        }
+
+        for key in ["model", "orchestrator", "ensemble"]:
```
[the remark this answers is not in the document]
ankona: 100

smartsim/_core/launcher/local/local.py
```diff
@@ -104,16 +113,26 @@ def run(self, step: Step) -> str:
             self.task_manager.start()
 
         out, err = step.get_output_files()
-        output = open(out, "w+")
-        error = open(err, "w+")
         cmd = step.get_launch_cmd()
 
+        is_enabled = int(os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1))
```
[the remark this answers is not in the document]
ankona: done

smartsim/_core/control/controller.py
ankona: Fixed!

MattToast:
One last thing to consider, and for sake of time I'm going to request this be kicked into a new ticket/PR!

It looks like unmanaged tasks need to launched through the `indirect.py` entry-point. We are successfully launching everything through the local launcher and that seems to be working perfectly! Unfortunately, it looks like we do not have a similar method to tack unamanged steps that may have been launched via a WLM:

For testing consider the following driver script: 
```py

def main() -> int:
    slurm = "slurm"      # <-- I believe this problem exists for any WLM
    open_mpi = "mpirun"  #     and run command combo that produces an
                         #     unmanaged step
    exp = Experiment("my-exp", launcher=slurm)

    rs = exp.create_run_settings("echo", ["hello", "world"],
                                 run_command=open_mpi)
    rs.set_tasks(1)
    model = exp.create_model("my-model", run_settings=rs)

    exp.generate(model, overwrite=True)
    exp.start(model, block=True)  # <-- Does not look like any start or
                                  #     stop jsons are produced bc the
                                  #     unmanaged indirect is never invoked
    print("Done")
    return 0

if __name__ == "__main__":
    raise SystemExit(main())
```
While a `slurm/mpirun` pairing tends to be wayyyy less popular than the usual `slurm/srun` configuration, it is not uncommon for users use a `PBS/mpi{exec,run}` pairing, so it may be worth testing that explicitly as well!

smartsim/_core/entrypoints/telemetrymonitor.py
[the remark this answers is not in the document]
ankona: removed

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,688 @@
...
+def hydrate_persistables(
+    entity_type: str,
+    run: t.Dict[str, t.Any],
+    exp_dir: pathlib.Path,
+) -> t.Dict[str, t.List[JobEntity]]:
+    """Map a collection of entity data persisted in a manifest file to an object"""
+    persisted: t.Dict[str, t.List[JobEntity]] = {
+        "model": [],
+        "orchestrator": [],
+        "ensemble": [],
```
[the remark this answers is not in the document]
ankona: removed

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: removed

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: telmon -> `set_launcher` does pass the same launcher into the job manager. once the first note above is added, we already have a shared launcher (though in the opposite direction you mentioned)

ankona:
> One last thing to consider, and for sake of time I'm going to request this be kicked into a new ticket/PR!
> 
> It looks like unmanaged tasks need to launched through the `indirect.py` entry-point. We are successfully launching everything through the local launcher and that seems to be working perfectly! Unfortunately, it looks like we do not have a similar method to tack unamanged steps that may have been launched via a WLM:
> 
> For testing consider the following driver script:
> 
> ```python
> from smartsim import Experiment
> 
> def main() -> int:
>     slurm = "slurm"      # <-- I believe this problem exists for any WLM
>     open_mpi = "mpirun"  #     and run command combo that produces an
>                          #     unmanaged step
>     exp = Experiment("my-exp", launcher=slurm)
> 
>     rs = exp.create_run_settings("echo", ["hello", "world"],
>                                  run_command=open_mpi)
>     rs.set_tasks(1)
>     model = exp.create_model("my-model", run_settings=rs)
> 
>     exp.generate(model, overwrite=True)
>     exp.start(model, block=True)  # <-- Does not look like any start or
>                                   #     stop jsons are produced bc the
>                                   #     unmanaged indirect is never invoked
>     print("Done")
>     return 0
> 
> if __name__ == "__main__":
>     raise SystemExit(main())
> ```
> 
> While a `slurm/mpirun` pairing tends to be wayyyy less popular than the usual `slurm/srun` configuration, it is not uncommon for users use a `PBS/mpi{exec,run}` pairing, so it may be worth testing that explicitly as well!



> One last thing to consider, and for sake of time I'm going to request this be kicked into a new ticket/PR!
> 
> It looks like unmanaged tasks need to launched through the `indirect.py` entry-point. We are successfully launching everything through the local launcher and that seems to be working perfectly! Unfortunately, it looks like we do not have a similar method to tack unamanged steps that may have been launched via a WLM:
> 
> For testing consider the following driver script:
> 
> ```python
> from smartsim import Experiment
> 
> def main() -> int:
>     slurm = "slurm"      # <-- I believe this problem exists for any WLM
>     open_mpi = "mpirun"  #     and run command combo that produces an
>                          #     unmanaged step
>     exp = Experiment("my-exp", launcher=slurm)
> 
>     rs = exp.create_run_settings("echo", ["hello", "world"],
>                                  run_command=open_mpi)
>     rs.set_tasks(1)
>     model = exp.create_model("my-model", run_settings=rs)
> 
>     exp.generate(model, overwrite=True)
>     exp.start(model, block=True)  # <-- Does not look like any start or
>                                   #     stop jsons are produced bc the
>                                   #     unmanaged indirect is never invoked
>     print("Done")
>     return 0
> 
> if __name__ == "__main__":
>     raise SystemExit(main())
> ```
> 
> While a `slurm/mpirun` pairing tends to be wayyyy less popular than the usual `slurm/srun` configuration, it is not uncommon for users use a `PBS/mpi{exec,run}` pairing, so it may be worth testing that explicitly as well!

Added a [follow-up ticket](https://github.com/CrayLabs/Poseidon/issues/575) to give this the love it deserves.

smartsim/_core/control/controller.py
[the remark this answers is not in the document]
ankona: i removed it for now just to avoid hassle. if i need it later, i'll re-create it.

smartsim/_core/control/controller.py
ankona: Removed the unused `step.meta['exp_path']` and removed the associated member variable on the controller. Per discussion, a refactor of the manifest builder may be necessary to avoid passing `exp_path` into the launch/start methods of controller

smartsim/_core/entrypoints/indirect.py
```diff
@@ -0,0 +1,228 @@
...
+
+def register_signal_handlers() -> None:
+    """Register a signal handling function for all termination events"""
+    for sig in SIGNALS:
+        signal.signal(sig, handle_signal)
+
+
+def get_parser() -> argparse.ArgumentParser:
+    # NOTE: plus prefix avoids passing param incorrectly to python interpreter,
+    # e.g. `python -m smartsim._core.entrypoints.indirect -c ... -t ...`
```
MattToast, line 167: Even more pedantic than usual!: `-c` and `-t` are no longer args for the indirect. Still got the point of the message, but my brain is angry they do not match :(

smartsim/_core/launcher/local/local.py
```diff
@@ -34,19 +36,15 @@
 from ..stepInfo import UnmanagedStepInfo, StepInfo
 from ..stepMapping import StepMapping
 from ..taskManager import TaskManager
+from ...utils.helpers import encode_cmd
+from ...config.config import get_config
 
 logger = get_logger(__name__)
-
+cfg = get_config()
```
MattToast, line 43: Not wrong, but there is a bit of pattern (anti-pattern?) that SS uses that I'd like to stick to for now:

SS will actually make a global `CONFIG` config that can be imported here
```py
from smartsim._core.config import CONFIG
```
while the `get_config()` method actually makes a new instance. This shouldn't  make a difference as of today bc as far as I can recall the `Config` class is stateless, but even still I'd like to keep that pattern in case we do add some state in future

smartsim/_core/launcher/local/local.py
```diff
@@ -127,11 +139,51 @@ def stop(self, step_name: str) -> UnmanagedStepInfo:
...
+            "+entity_type",
+            etype,
+            "+telemetry_dir",
+            status_dir,
+            "+working_dir",
+            step.cwd,
+            "+output_file",
+            out,
+            "+error_file",
+            err,
```
MattToast, line 187: TY for the long args!! Much more readable here IMO!!

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,675 @@
...
+        help="Experiment root directory",
+        required=True,
+    )
+    arg_parser.add_argument(
+        "-n",
+        type=int,
+        help="Automatically shutdown after a specific number of polling iterations",
+        default=0,
+        required=False,
+    )
```
MattToast, line 652: Sorry to be pedantic again, but can I request long args here as well 😅

smartsim/_core/control/controller.py
```diff
@@ -110,6 +162,11 @@ def start(
         self._jobs.kill_on_interrupt = kill_on_interrupt
         # register custom signal handler for ^C (SIGINT)
         signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
+
+        # have job manager launch a telemetry monitor if any jobs are run
+        if not self._jobs._on_start_hook:  # pylint: disable=protected-access
+            self._jobs.add_job_onstart_callback(create_telmon_callback(exp_path))
```
MattToast, line 168: Still not in love with this solution, but willing to call it good enough for now!! @MattToast and/or @ankona remember to write up a ticket to address in future

smartsim/_core/control/controller.py
```diff
@@ -497,7 +565,9 @@ def _launch_step(
             self._jobs.add_job(job_step.name, job_id, entity, is_task)
 
     def _create_batch_job_step(
-        self, entity_list: t.Union[Orchestrator, Ensemble, _AnonymousBatchJob]
+        self,
+        entity_list: t.Union[Orchestrator, Ensemble, _AnonymousBatchJob],
+        telemetry_dir: pathlib.Path,
```
MattToast, line 570: Same "don't love, but good enough for now" concern with param diving the telemetry_dir: @MattToast and/or @ankona remember to ticket this as well!!

MattToast (approved):
Looks amazing and seems to be working great! I left a couple _suuuupppeer_ pedantic change requests and some notes for either you/I to write up some tickets for some known defects that we want to address, but nothing worth holding up approval over!

Otherwise LGTM!! Thanks for all of the hard work on this!!

smartsim/_core/launcher/local/local.py
ankona: Swapped to shared instance of config!

smartsim/_core/launcher/local/local.py
ankona: w00t!

smartsim/_core/entrypoints/indirect.py
ankona: removing the comment altogether. i didn't like it's face!

smartsim/_core/control/controller.py
ankona: I think i'm abusing the callbacks here, anyway. Why don't I just call the start method instead of add a callback?

ankona: I've removed the callbacks altogether.

smartsim/_core/entrypoints/telemetrymonitor.py
ankona: done!

smartsim/_core/control/controller.py
ankona: +1

PR #357, opened 2023-09-07
519: Update Streamlit code to parse SmartSim manifest.json

AlyssaCote:
I added a FileReader class that parses a manifest.json file that the dashboard can consume. This PR also has most of the dashboard code added to it as well. This is super long because the test manifest.json files!


AlyssaCote:
This pr was moved to the SmartDashboard repo.

PR #358, opened 2023-09-08
Remove references in docs to nonexistent CLI flag

MattToast:
Remove references in the docs to the (now invalid) `--only_python_pacakges` flag of `smart build`


ashao (approved):
Thanks for catching that! Especially like the fact that you actually call out to `smart build --help` now so that we don't have to remember to manually touch this in the future.

PR #359, opened 2023-09-11
Pretty print error message when onnx wheel not available

MattToast:
In the case where a user is using python 3.10 and has requested to build with
```sh
smart build --onnx
```
there is no compliant onnx wheel for python 3.10 for the ORT backend installed by default for RAI. This was leading to a `SetupError` being raised, and eventually exposing itself to the user.

This PR will handle that raised exception and pretty print the error message within, before exiting with a non-zero exit code.


al-rigazzi (approved):
LGTM!

PR #365, opened 2023-09-14
Torch install back to build

MattToast:
Move torch python package back into the `smart build` target for PyPI


al-rigazzi (approved):
LGTM, thanks for this quick fix

ashao (approved):
Thanks!

PR #366, opened 2023-09-14
Improvements to colocated DB launch

al-rigazzi:
This PR adds two improvements to our current colocated launch mechanism.

In detail:
- the script used to run a colocated application has now a unique name, to avoid race conditions with other entities being launched in the same directory
- a warning was issued about incorrect interface binding if no interface was supplied. This warning is now removed, as colocated DBs are either bound to `lo` or use Unix Domain Sockets to communicate.


MattToast (approved):
LGTM!! Thanks for handling this; this should make debugging launching databases much less noisy!

smartsim/_core/launcher/step/step.py
```diff
@@ -75,7 +75,9 @@ def get_step_file(
 
     def get_colocated_launch_script(self) -> str:
         # prep step for colocated launch if specifed in run settings
-        script_path = self.get_step_file(script_name=".colocated_launcher.sh")
+        script_path = self.get_step_file(
+            script_name=f".colocated_launcher_{self.entity_name}.sh"
```
mellis13, line 79: I'm not sure that we enforce unique names for constructors ``experiment.create_model()``.  Is this potentially an edge case that we will hit?  Since this is internal only, would we want to add a very short uuid to the end that guarantees no overwrites or races?

tests/test_colo_model_local.py
```diff
@@ -49,7 +49,7 @@ def test_macosx_warning(fileutils, coloutils):
         RuntimeWarning,
         match="CPU pinning is not supported on MacOSX. Ignoring pinning specification."
     ):
-        colo_model = coloutils.setup_test_colo(
```
mellis13, line 52: General questions on tests:  do we have a test that checks for non-generated experiments with more than one colocated model?   That was how the bug was first discovered.

mellis13 (changes requested):
Couple of questions but otherwise great PR.  Thanks for the fixes!

smartsim/_core/launcher/step/step.py
al-rigazzi: The issue with UUID is that if the user runs the experiment N times, we end up creating N scripts. And a similar issue arises if we try to issue an error on duplicated entity names: re-running the same entity, we might misinterpret its own launcher script for that of another entity. I think we should open a ticket about making sure two entities with the same name can not be run at the same time - from what I see in the code base, if that happens, there are other parts of the `JobManager` which break. Thoughts?

tests/test_colo_model_local.py
al-rigazzi: I think we do, but not for local launcher. I'll add one.

al-rigazzi: Done

smartsim/_core/launcher/step/step.py
al-rigazzi: I opened a dedicated ticket

ankona: `get_step_file` automatically prepends the entity name to the file name. If we had to do some sort of special character removal (or any processing of the entity name), using this formatting string would make us inconsistent or have to repeat that code.

Is there any reason we don't do:

```
script_path = self.get_step_file("colocated_launcher")
```
and let `get_step_file` just spit out the standard `{entity_name}.{file_name}.{file_ext}`?

tests/backends/test_dbmodel.py
```diff
@@ -807,7 +807,7 @@ def test_inconsistent_params_db_model():
     # Create and save ML model to filesystem
     model, inputs, outputs = create_tf_cnn()
     with pytest.raises(SSUnsupportedError) as ex:
-        db_model = DBModel(
+        _ = DBModel(
```
ankona, line 810: I think Toast dislikes bothering with an unused "anonymous" variable. Why bother with `_ = ` at all if you don't need it?

tests/test_colo_model_local.py
```diff
@@ -37,70 +37,81 @@
...
 def test_unsupported_limit_app(fileutils, coloutils):
-    db_args = {"limit_app_cpus":True}
-    db_type = 'uds' # Test is insensitive to choice of db
+    db_args = {"limit_app_cpus": True}
+    db_type = "uds"  # Test is insensitive to choice of db
 
     exp = Experiment("colocated_model_defaults", launcher="local")
     with pytest.raises(SSUnsupportedError):
-        colo_model = coloutils.setup_test_colo(
+        _ = coloutils.setup_test_colo(
```
ankona, line 67: ditto on unnecessary var

tests/test_colo_model_local.py
```diff
@@ -37,70 +37,81 @@
...
+@pytest.mark.parametrize("custom_pinning", [1, "10", "#", 1.0, ["a"], [1.0]])
 def test_unsupported_custom_pinning(fileutils, coloutils, custom_pinning):
-    db_type = "uds" # Test is insensitive to choice of db
+    db_type = "uds"  # Test is insensitive to choice of db
     db_args = {"custom_pinning": custom_pinning}
 
     exp = Experiment("colocated_model_defaults", launcher="local")
     with pytest.raises(TypeError):
-        colo_model = coloutils.setup_test_colo(
+        _ = coloutils.setup_test_colo(
```
ankona, line 83: ditto on unnecessary var

tests/backends/test_dbmodel.py
al-rigazzi: Right. Did it on all occurrences you pointed out.

mellis13 (approved):
LGTM!

ashao (approved):
All looks good to me!

ankona (approved):

PR #372, opened 2023-09-19
Publish artifacts from test_output on CI fail

ashao:
When trying to debug tests on the Github CI, the output from pytest is often insufficient because some errors are written to the redirected stderr and stdout of launched applications. This PR uploads the test_output directory as an artifact of the workflow on failure.


MattToast (approved):
Great idea!! This seems like something that's been long overdue!

al-rigazzi (approved):
Very cool! LGTM

PR #374, opened 2023-09-20
Fix broken 'Install' link in README

MattToast:
The 'Install' link in README was pointing to a since moved and renamed page on the Cray Labs docs. It has been updated to now point to the basic installation of SmartSim.


al-rigazzi (approved):
LGTM, thanks for this fix!

PR #383, opened 2023-10-03
Drop support for RedisAI 1.2.5

al-rigazzi:
This PR removes support for RedisAI 1.2.5. This also implies that the entries in the GitHub actions matrix is reduced.


smartsim/_core/_cli/build.py
```diff
@@ -67,8 +67,6 @@ def check_py_onnx_version(versions: Versioner) -> None:
             f"Python {py_version.major}.{py_version.minor}. "
             "Instead consider using Python 3.8 or 3.9 with Onnx "
         )
-        if sys.platform == "linux":
-            msg += "1.2.5 or "
         msg += "1.2.7."
```
MattToast, line 70: Since we no longer have branching logic based on OS, would you mind folding this into a single str to avoid the extra bin op and alloc?
```py
"Instead consider using Python 3.8 or 3.9 with RedisAI 1.2.7"
#                                              ^^^^^^^^^^^^^
# Also could fix this `Onnx` vs `RAI` typo, while your here?
```

smartsim/_core/_install/buildenv.py
```diff
@@ -189,17 +189,6 @@ class RedisAIVersion(Version_):
...
-            "tensorflow": "2.6.2",
-            "onnx": "1.9.0",
-            "skl2onnx": "1.10.3",
-            "onnxmltools": "1.10.0",
-            "scikit-learn": "1.0.2",
-            "torch": "1.9.1",
-            "torch_cpu_suffix": "+cpu",
-            "torch_cuda_suffix": "+cu111",
-            "torchvision": "0.10.1",
-        },
```
MattToast, line 202: You have no idea how happy this makes me! Goodbye TF 2.6.x!! 🎉

smartsim/_core/_install/builder.py
```diff
@@ -351,21 +351,6 @@ def build_onnx(self) -> bool:
     def fetch_onnx(self) -> bool:
         return self.build_onnx
 
-    def copy_tf_cmake(self) -> None:
-        """Copy the FindTensorFlow.cmake file to the build directory
-        as the version included in RedisAI is out of date for us.
-        Note: opt/cmake/modules removed in RedisAI v1.2.5
```
MattToast, line 357: If we are no longer using it (and presumably haven't since RAI 1.2.3) can we check `smartsim/_core/bin/modules/FindTensorFlow.cmake` out of our source control?

doc/installation_instructions/platform/olcf-summit.rst
```diff
@@ -50,7 +50,7 @@ directory from which these instructions are executed.
   # install PyTorch and TensorFlow backend for the Orchestrator database.
   export Torch_DIR=/ccs/home/$USER/.conda/envs/$ENV_NAME/lib/python3.9/site-packages/torch/share/cmake/Torch/
   export CFLAGS="$CFLAGS -I/ccs/home/$USER/.conda/envs/$ENV_NAME/lib/python3.9/site-packages/tensorflow/include"
-  export SMARTSIM_REDISAI=1.2.5
+  export SMARTSIM_REDISAI=1.2.7
```
MattToast, line 53: Do we need to update the site install on summit if we are going to make this change to the docs? Should be put in a ticket to do that sometime before we release?

.github/workflows/run_tests.yml
```diff
@@ -100,7 +93,7 @@ jobs:
           sudo make -C builddir install
 
       - name: singularity pull test container # This lets us time how long the pull takes
-        if: contains( matrix.os, 'ubuntu' ) && matrix.py_v == 3.9 && matrix.rai == '1.2.5'
```
mellis13, line 103: Does this docker image use an old version of pytorch we need to upgrade (1.9.1 instead of 1.11):

https://github.com/CrayLabs/SmartSim/blob/develop/docker/testing/Dockerfile

MattToast (changes requested):
Looks great! I have some small questions and couple of things I'm gonna ask to tack onto this PR:

1) Can you update the change log for this PR now? I'd like have this documented in case anyone is currently working off of develop and I ESPECIALLY want to make sure that we do not forget to call this out when release time rolls around.
2) There are a 2 pylint suppressions that needed to be made on account of the pylint version that came from TF 2.6.x. Can you remove them while you are here? They are located at `smartsim/entity/entityList.py:36` and `smartsim/_core/_cli/validate.py:55`.

Other than that and the comments below, this looks ready to go on my end!

smartsim/_core/_cli/build.py
```diff
@@ -67,8 +67,6 @@ def check_py_onnx_version(versions: Versioner) -> None:
             f"Python {py_version.major}.{py_version.minor}. "
             "Instead consider using Python 3.8 or 3.9 with Onnx "
         )
-        if sys.platform == "linux":
-            msg += "1.2.5 or "
```
mellis13, line 71: Your removal makes sense but I'm confused by the underlying error message.  Shouldn't we be saying that 1.2.7 applies to the redisAI version not the ONNX version?

mellis13 (changes requested):
Thanks for this PR -- just a couple of minor questions

mellis13:
One more comment -- don't forget to add tags to the PR for release documentation

mellis13:
@MattToast @al-rigazzi Should we remove the branching logic in ``RedisAIBuilder.build_from_git()`` since we only have 1.2.7 now?  It seems like it would make sense for all of 1.2.7 to build from the same commit.  What do you all think?
https://github.com/CrayLabs/SmartSim/blob/e930cbb14dfda515ff08fdd34b23982dd63eadd5/smartsim/_core/_install/builder.py#L441 .

smartsim/_core/_cli/build.py
MattToast: yeah, that is a typo that I just noticed and asked Al to fix up while he is here

doc/installation_instructions/platform/olcf-summit.rst
al-rigazzi: I updated this script and I think it will work, even though there is no matching IBM version of PyTorch available (but I remember 1.10 being largely compatible). We will open a ticket for the site install and make sure it works.

smartsim/_core/_cli/build.py
al-rigazzi: OK, wait, here there is something I don't understand (and I think @mellis13 also mentioned this): if a user is here, RedisAI 1.2.7 is being used (we do not support other versions), but the problem is that ONNX is not compatible with Python. Should we put the ONNX version (1.11) instead of the RedisAI version?

.github/workflows/run_tests.yml
al-rigazzi: Yes, I think you are right. We need to update the container too before we release.

smartsim/_core/_cli/build.py
al-rigazzi: Yep, see other comment on modified file - I am a bit confused.

smartsim/_core/_install/buildenv.py
al-rigazzi: Progress is unstoppable!

smartsim/_core/_cli/build.py
al-rigazzi: @mellis13 @MattToast unifying warning message thread

MattToast: ^^ Agreed! lets use that

MattToast (approved):
LGTM pending one last error message tweak!! So happy to see those very old ml deps go away! The future is NOW!!

PR #386, opened 2023-10-05
Serialize launched entities manifest

MattToast:
Bare bones, rough draft implementation of how SmartSim may be able to serialize information about launched entities and/or collections of entities and dump that information to a file so that external or third party tools may be able to display and monitor jobs outside of the process running the python interpreter that ran a driver script.

So far this has been tested by hand and seems to be working for local and slurm launchers for models/ensembles/databases on both interactive and batch jobs.

TODO:
- [x] TESTS!! (unit and integration)
- [ ] Run against other WLMs (mainly PBS)
- [x] Address and remove numerous ``TODO`` and ``FIXME`` comments

### DIFF
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -27,34 +27,31 @@
 from __future__ import annotations
 
 import os.path as osp
-from os import environ
 import pickle
 import signal
 import threading
 import time
 import typing as t
+from os import environ
 
 from smartredis import Client, ConfigOptions
 
 from ..._core.launcher.step import Step
+from ..._core.utils.helpers import unpack_colo_db_identifier, unpack_db_identifier
 from ..._core.utils.redis import db_is_active, set_ml_model, set_script, shutdown_db
-from ..._core.utils.helpers import (
-    unpack_db_identifier,
-    unpack_colo_db_identifier,
-)
 from ...database import Orchestrator
 from ...entity import Ensemble, EntityList, EntitySequence, Model, SmartSimEntity
 from ...error import (
     LauncherError,
     SmartSimError,
+    SSDBIDConflictError,
     SSInternalError,
     SSUnsupportedError,
-    SSDBIDConflictError,
 )
 from ...log import get_logger
+from ...servertype import CLUSTERED, STANDALONE
 from ...settings.base import BatchSettings
 from ...status import STATUS_CANCELLED, STATUS_RUNNING, TERMINAL_STATUSES
-from ...servertype import STANDALONE, CLUSTERED
 from ..config import CONFIG
 from ..launcher import (
     CobaltLauncher,
@@ -64,13 +61,14 @@ from ..launcher import (
     SlurmLauncher,
 )
 from ..launcher.launcher import Launcher
-from ..utils import check_cluster_status, create_cluster
+from ..utils import check_cluster_status, create_cluster, serialize
 from .job import Job
 from .jobmanager import JobManager
-from .manifest import Manifest, LaunchedManifest
+from .manifest import LaunchedManifest, LaunchedManifestBuilder, Manifest
 
 if t.TYPE_CHECKING:
     from ..launcher.stepMapping import StepMap
+    from ..utils.serialize import TStepLaunchMetaData
 
 logger = get_logger(__name__)
 
@@ -94,10 +92,13 @@ class Controller:
         self.init_launcher(launcher)
 
     def start(
-        self, manifest: Manifest, block: bool = True, kill_on_interrupt: bool = True
-    ) -> LaunchedManifest[
-        t.Tuple[t.Optional[str], t.Optional[str], t.Optional[bool], str, str]
-    ]:
+        self,
+        exp_name: str,
+        exp_path: str,
+        manifest: Manifest,
+        block: bool = True,
+        kill_on_interrupt: bool = True,
+    ) -> None:
         """Start the passed SmartSim entities
 
         This function should not be called directly, but rather
@@ -109,54 +110,21 @@ class Controller:
         self._jobs.kill_on_interrupt = kill_on_interrupt
         # register custom signal handler for ^C (SIGINT)
         signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
-        launched = self._launch(manifest)
+        launched = self._launch(exp_name, exp_path, manifest)
 
         # start the job manager thread if not already started
         if not self._jobs.actively_monitoring:
             self._jobs.start()
+
+        serialize.save_launch_manifest(
+            launched.map(_look_up_launched_data(self._launcher))
+        )
 
         # block until all non-database jobs are complete
         if block:
             # poll handles its own keyboard interrupt as
             # it may be called seperately
             self.poll(5, True, kill_on_interrupt=kill_on_interrupt)
-
-        def _prep_launched_metadata(
-            launcher: Launcher, launched_step_name: str, step: Step
-        ) -> t.Tuple[t.Optional[str], t.Optional[str], t.Optional[bool], str, str]:
-            # NOTE: we cannot assume that the name of the launched step
-            # ``launched_step_name`` is equal to the name of the step referring
-            # to the entity ``step.name`` as is the case when an entity list is
-            # launched as a batch job
-            launched_step_map = launcher.step_mapping[launched_step_name]
-            out_file, err_file = step.get_output_files()
-            return (
-                launched_step_map.step_id,
-                launched_step_map.task_id,
-                launched_step_map.managed,
-                out_file,
-                err_file,
-            )
-
-        # FIXME: These list comprehensions are unreadable
-        return LaunchedManifest(
-            models=[
-                (model, _prep_launched_metadata(self._launcher, step_name, step))
-                for model, (step_name, step) in launched.models
-            ],
-            ensembles=[
-                (ens, [(model,
-                        _prep_launched_metadata(self._launcher, step_name, step))
-                       for model, (step_name, step) in model_list])
-                for ens, model_list in launched.ensembles
-            ],
-            database=[
-                (db_, [(node,
-                        _prep_launched_metadata(self._launcher, step_name, step))
-                       for node, (step_name, step) in node_list])
-                for db_, node_list in launched.database
-            ],
-        )
 
     @property
     def orchestrator_active(self) -> bool:
@@ -354,7 +322,9 @@ class Controller:
         else:
             raise TypeError("Must provide a 'launcher' argument")
 
-    def _launch(self, manifest: Manifest) -> LaunchedManifest[t.Tuple[str, Step]]:
+    def _launch(
+        self, exp_name: str, exp_path: str, manifest: Manifest
+    ) -> LaunchedManifest[t.Tuple[str, Step]]:
         """Main launching function of the controller
 
         Orchestrators are always launched first so that the
@@ -364,7 +334,7 @@ class Controller:
         :type manifest: Manifest
         """
 
-        launched_manifest = LaunchedManifest[t.Tuple[str, Step]]()
+        manifest_builder = LaunchedManifestBuilder[t.Tuple[str, Step]]()
         # Loop over deployables to launch and launch multiple orchestrators
         for orchestrator in manifest.dbs:
             for key in self._jobs.get_db_host_addresses():
@@ -382,7 +352,7 @@ class Controller:
                 raise SmartSimError(
                     "Local launcher does not support multi-host orchestrators"
                 )
-            self._launch_orchestrator(orchestrator, launched_manifest)
+            self._launch_orchestrator(orchestrator, manifest_builder)
 
         if self.orchestrator_active:
             self._set_dbobjects(manifest)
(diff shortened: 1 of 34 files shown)


docker/dev/Dockerfile
```diff
@@ -52,7 +52,7 @@ USER craylabs
 
 RUN cd SmartSim && SMARTSIM_SUFFIX=dev python -m pip install .
 
-RUN python -m pip install smartsim[ml,dev]==0.5.0 jupyter jupyterlab matplotlib && \
+RUN python -m pip install smartsim[ml,dev]==0.5.1 jupyter jupyterlab matplotlib && \
```
ankona, line 55: Could we pass the version using:

```
ARG SSVERSION 
... 
RUN python -m pip install smartsim[ml,dev]==${SSVERSION}
```

and avoid having to make this change in multiple places each time we release?

Could we also use this same key to generate ./doc/version.txt and avoid updating `doc/conf.py`?

docker/prod/Dockerfile
```diff
@@ -44,7 +44,7 @@ WORKDIR /home/craylabs
 COPY --chown=craylabs:root ./tutorials/ /home/craylabs/tutorials/
 
 USER craylabs
-RUN python -m pip install smartsim[ml]==0.5.0 jupyter jupyterlab matplotlib && \
+RUN python -m pip install smartsim[ml]==0.5.1 jupyter jupyterlab matplotlib && \
```
ankona, line 47: ditto on parameterization

docker/dev/Dockerfile
ankona: maybe better to do `ENV SMARTSIM_VERSION` instead to use the same source being used in [buildenv.py](https://github.com/CrayLabs/SmartSim/pull/386/files#diff-7945f9e10b608a1cc73eec4773459e082e858f3f7f37fed67c8a4fe6e0a7db5fR276)

smartsim/_core/_install/builder.py
```diff
@@ -462,7 +447,7 @@ def build_from_git(self, git_url: str, branch: str, device: str = "cpu") -> None
                 "634916c722e718cc6ea3fad46e63f7d798f9adc2",
             ]
         else:
-            # Clone RAI release commit
+            # Clone RAI release commit for versions > 1.2.7
```
ankona, line 450: nitpick: I think the comment doesn't accurately reflect that a version both LT and GT 1.2.7 will be cloned here

smartsim/_core/control/controller.py
```diff
@@ -105,6 +109,43 @@ def start(
...
+            out_file, err_file = step.get_output_files()
+            return (
+                launched_step_map.step_id,
+                launched_step_map.task_id,
+                launched_step_map.managed,
+                out_file,
+                err_file,
+            )
+
+        # FIXME: These list comprehensions are unreadable
```
ankona, line 129: I can _read_ them!

smartsim/_core/control/controller.py
```diff
@@ -189,11 +234,26 @@ def stop_entity(self, entity: t.Union[SmartSimEntity, EntityList]) -> None:
                 )
                 self._jobs.move_to_completed(job)
 
-    def stop_entity_list(self, entity_list: EntityList) -> None:
+    def stop_db(self, db: Orchestrator) -> None:
```
ankona, line 237: Is there any easy way we can add something like `def stop(self)` into the API for EntityList and SmartSimEntity so the controller can avoid having custom methods for each type. 

The result would be that stop_entity and shutdown_db use <entity>.stop and the controller no longer needs to know exact details of db implementation like `cmd = [redis_cli, "-h", host_ip, "-p", str(port), "shutdown"]`

ankona: Probably better for follow-up task if deemed useful

smartsim/_core/control/controller.py
```diff
@@ -209,34 +269,40 @@ def get_jobs(self) -> t.Dict[str, Job]:
         with JM_LOCK:
             return self._jobs.completed
 
-    def get_entity_status(self, entity: t.Union[SmartSimEntity, EntityList]) -> str:
+    def get_entity_status(
+        self, entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]]
```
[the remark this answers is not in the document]
ankona: or, since we dropped python 3.7, can we swap to A | B syntax?

smartsim/_core/control/controller.py
```diff
@@ -296,21 +364,38 @@ def _launch(self, manifest: Manifest) -> None:
                 msg = "Attempted to launch a second Orchestrator instance. "
                 msg += "Only 1 Orchestrator can be active at a time"
                 raise SmartSimError(msg)
-            self._launch_orchestrator(orchestrator)
+            self._launch_orchestrator(orchestrator, launched_manifest)
```
ankona, line 367: What do you think of having `._launch_xxx` methods return something like a ``LaunchManifest`` (basically, a little event type that contains the details) instead of modifying the launched_manifest in-place? Maybe that gives us an easier way to test that the right things are changed and minimizes the ability to change the launch manifest in undesired ways?

smartsim/_core/control/controller.py
```diff
@@ -275,7 +341,7 @@ def init_launcher(self, launcher: str) -> None:
         else:
             raise TypeError("Must provide a 'launcher' argument")
 
-    def _launch(self, manifest: Manifest) -> None:
+    def _launch(self, manifest: Manifest) -> LaunchedManifest[t.Tuple[str, Step]]:
```
ankona, line 344: out-of-scope enhancement request: I think you're potentially implementing exactly what we might need for an upcoming launch preview ticket. 

Could we save some time later by separating the launch manifest building from the "true launch?"

@amandarichardsonn might be interested in this work

smartsim/_core/entrypoints/redis.py
```diff
@@ -54,31 +58,64 @@ def handle_signal(signo: int, _frame: t.Optional[FrameType]) -> None:
...
-    bind_src_arg = f"--bind-source-addr {ip_addresses[0]}"
-    return [bind_arg, bind_src_arg]
+def build_bind_args(source_addr: str, *addrs: str) -> t.Tuple[str, ...]:
+    return (
+        "--bind",
+        source_addr,
+        *addrs,
+        # pin source address to avoid random selection by Redis
+        "--bind-source-addr",
+        source_addr,
```
ankona, line 68: in the prior version, correctly or not, it built --bind-source-addr with the first address only and --bind with all addresses.

i like the idea of specifying the source_addr instead of assuming 0, but wonder if we need to ensure that source_addr is also in addrs (e.g. call was `x.build_bind_args(addrs[0], addrs[1:])`

smartsim/_core/generation/generator.py
```diff
@@ -59,12 +65,28 @@ def __init__(self, gen_path: str, overwrite: bool = False) -> None:
...
+    @property
+    def log_file(self) -> str:
+        """Returns the location of the file
+        summarizing the parameters used for the last generation
+        of all generated entities.
+
+        :returns: path to file with parameter settings
+        :rtype: str
+        """
+        return join(self.gen_path, "smartsim_params.txt")
```
ankona, line 89: should we shift default `smartsim_params.txt` into `smartsim._core.config.config` and retrieve it from there instead?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,216 @@
...
+            "Symlink": [],
+            "Configure": [],
+            "Copy": [],
+        },
+        "colocated_db": {
+            "settings": model.run_settings.colocated_db_settings,
+            "scripts": [
+                {
+                    script.name: {
+                        "backend": "TORCH",
```
ankona, line 114: should this always be torch?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,216 @@
...
+) -> None:
+    manifest_dir = Path(experiment.exp_path) / ".smartsim/manifest"
+    manifest_dir.mkdir(parents=True, exist_ok=True)
+    manifest_file = manifest_dir / "manifest.json"
+
+    # FIXME: Did we decide on if this should be uuid/literal count of the
+    #        runs/time since epoch/etc.? Whatever it is, it probably should not
+    #        be secs since epoch as this could lead to conflicts if many
+    #        non-blocking runs are made in quick succession.
+    run_id = int(time.time())
```
AlyssaCote, line 33: I am the cause of this confusion! I'm pretty sure that originally we wanted a time. Correct me if I'm wrong @ankona but I'm pretty sure you needed it for something? And then in my fake manifests I wrote integers, which made them seem more like literal counts.

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,216 @@
...
+def _dictify_db(
+    db: Orchestrator,
+    nodes: t.List[t.Tuple[DBNode, _TStepLaunchMetaData]],
+    telemetry_data_path: Path,
+) -> t.Dict[str, t.Any]:
+    db_path = _utils.get_db_path()
+    if db_path:
+        db_type, _ = db_path.name.split("-", 1)
+    else:
+        db_type = "Unkown"
```
AlyssaCote, line 192: spelling

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,216 @@
...
+        },
+        "out_file": out_file,
+        "error_file": err_file,
+    }
+
+
+def _dictify_ensemble(
+    ens: Ensemble,
+    members: t.List[t.Tuple[Model, _TStepLaunchMetaData]],
+    telemetry_data_path: Path,
```
AlyssaCote, line 146: Ensembles should have their permutation strategy included

AlyssaCote (changes requested):
Wanted to get a couple comments in before you finish working on the TODOs in your PR! I think it's looking really good. We need to figure out what Chris needs for run_ids and talk about experiment logs still, but that's pretty much it!

smartsim/_core/utils/serialize.py
MattToast: 🤦

smartsim/_core/utils/serialize.py
AlyssaCote: now that we are selected entities on name and run_id and displaying both in the selection dropdowns, i'd prefer these be a literal count or uuid simply because i think that would look better when displayed.

ankona: Since time doesn't usually go backwards, I chose to use the timestamp as the unique identifier when logging details for a run. The integer conversion was just because a per-experiment folder-naming collision at the second level is extremely unlikely and I prefer to avoid special characters in folder names if a human will peruse them. Also, each run would appear _in-order_ in the output directory.

I've [added](https://github.com/CrayLabs/SmartSim/pull/356/files#diff-59a63a8a45350c2c4a46ef5841c206c22b72d74a7230792e37c8a1499c6109f4R252-R254) a `get_ts` function into the helpers that converts a timestamp float to an integer:

image

A count, IMO, is more likely to have collision issues w/outputs. A UUID doesn't have collision issues but is longer than necessary, has special characters, and forces a user browsing the folders to hop around to find things in order.

MattToast: Discussed offline and decided to use a uuid for `"run_id"`, but added a `"timestep"` key to the manifest json.

smartsim/_core/utils/serialize.py
MattToast: Right now, the only way a user can execute arbitrary python code on the database is if they compile it to torch script. So (for now) the `DBScript` does not have a `backend` attr and there is the (potentially naive) assumption throughout SmartSim that all `DBScript`s are using the Torch backend

smartsim/_core/control/controller.py
MattToast: LOL!! Moved mapping of `LaunchedManifest` data to a separate (hopefully more readable) helper method

smartsim/_core/control/controller.py
MattToast: Interesting! I tried decoupling the construction of the `LaunchedManifest` by moving it into a `LuanchedManifestBuilder` class which at least makes the intention of the code clear, but I'm still not happy with the out parameter.

I might need to pick you brain on this as I would much rather keep the use of out parameters to minimum and constructing piece wise via "actions" seems like it could be a better solution!!

smartsim/_core/control/controller.py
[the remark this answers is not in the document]
MattToast: > Should we add a _SmartSimEntity = t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]] alias for brevity due to how often this union is repeated?

Absolutely, but I would argue that it is outside the scope of this ticket. I will put one up to do this

> or, since we dropped python 3.7, can we swap to A | B syntax?

Unfortunately, we cannot "out of the box" as the `A | B` syntax was introduced natively in python 3.10. That said, if we are willing to adopt the use of `from __future__ import annotations` and perhaps a adjust couple of mypy settings, it should be backward compatible! What would you think about adding a ticket for this work as well?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,221 @@
...
+    err_file: str,
+    telemetry_data_path: Path,
+) -> t.Dict[str, t.Any]:
+    colo_settings = (model.run_settings.colocated_db_settings or {}).copy()
+    db_scripts = t.cast("t.List[DBScript]", colo_settings.pop("db_scripts", []))
+    db_models = t.cast("t.List[DBModel]", colo_settings.pop("db_models", []))
+    return {
+        "name": model.name,
+        "path": model.path,
+        "run_settings": _dictify_run_settings(model.run_settings),
```
AlyssaCote, line 99: Currently we have `exe_args` in the `applications` section of the dashboard, rather than have it inside `run_settings`.  I think that's something we should get aligned on.

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,221 @@
...
+        if colo_settings
+        else None,
+        "telemetry_metadata": {
+            "status_dir": str(telemetry_data_path / model.name),
+            "step_id": step_id,
+            "task_id": task_id,
+            "managed": managed,
+        },
+        "out_file": out_file,
+        "error_file": err_file,
```
AlyssaCote, line 145: Can we rename this to `err_file` for consistency?

MattToast: Whoops! Consider it renamed!

tests/test_serialize.py
```diff
@@ -0,0 +1,80 @@
+from pathlib import Path
+import json
+
+from smartsim import Experiment
+from smartsim._core.utils import serialize
+from smartsim._core.control.manifest import LaunchedManifestBuilder
+
+
+def test_serialize_creates_a_maifest_json_file_if_dne(fileutils):
```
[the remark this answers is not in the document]
MattToast: spelling is hard  :(

smartsim/_core/utils/serialize.py
MattToast: I moved them back to the `"models"`/application section, but I do think they should be moved back to run settings as that is where they exist in memory and (at least imo) the manifest should should follow that structure

Adding a TODO comment and created a ticket to track the work!

PR #391, opened 2023-10-11
Switch to `style`

al-rigazzi:
This PR addresses the shadowing of the builitin symbol `format` in the signature of the `Experiment.summary()` member function.

A lint pragma instruction could be removed, following renaming of `format` to `style`. As the argument name changes, this has to be considered and API break.


mellis13 (approved):
Thanks, LGTM!

PR #392, opened 2023-10-11
Make some Slurm errors fatal for execution

al-rigazzi:
When running on Slurm-based systems, it can happen that some commands are available, but return an error when the SlurmDB does not work properly. This error could silently lead to SmartSim's execution looping forever. It was decided that any error on `sacct` will be fatal for execution, as SmartSim makes use of `sacct` to acquire information about running , paused,and finished jobs.


smartsim/_core/launcher/slurm/slurmCommands.py
```diff
@@ -120,3 +110,18 @@ def _find_slurm_command(cmd: str) -> str:
         raise LauncherError(
             f"Slurm Launcher could not find path of {cmd} command"
         ) from e
+
+
+def _execute_slurm_cmd(
```
ankona, line 115: It seems like all the WLMs would benefit from this type of check and two other options might be preferable:

- add `def ensure_preconditions()` to the launcher contract and execute that when launching an experiment instead of getting down to an individual cmd. This lets every launcher add their own check for preconditions and gives a quicker, clearer indicator of what failed (bonus: it can be executed in the preview to show a user that their experiment won't complete)
- generalize `_find_slurm_command` into `_find_required_command(..., owner: str)` in shell.py. Pass in "Slurm"/"PBS" and avoid every WLM copying the same 30 lines

al-rigazzi: As per our private conversation, we will create a ticket for this and won't implement it as part of this ticket. 🚀

ankona (approved):
per offline convo, I will create a ticket to review adding checks to a higher level abstraction. 

otherwise, looks good!

smartsim/_core/launcher/slurm/slurmCommands.py
```diff
@@ -38,10 +41,8 @@ def sstat(args: t.List[str]) -> t.Tuple[str, str]:
...
     """
-    _sstat = _find_slurm_command("sstat")
-    cmd = [_sstat] + args
-    _, out, error = execute_cmd(cmd)
-    return out, error
+    _, out, err = _execute_slurm_cmd("sstat", args, True)
+    return out, err
 
 
 def sacct(args: t.List[str]) -> t.Tuple[str, str]:
```
MattToast, line 48: I'm not sure that love that some of the slurm commands raise and some of them silently fail with no real way to tell which does which without checking the implementation.

Potentially more work than it's worth, but I figured I would throw the idea out there. What would your thoughts be on allowing the caller of `sstat`/`sacct`/etc. decide where or not an error should be exposed by allowing them to set a kw-only `raise_on_err` flag that is passed directly to `_execute_slurm_cmd` and for ALL slurm commands is default `False`?
```py
def sacct(args: t.List[str], *, raise_on_err: bool = False) -> t.Tuple[str, str]:
```
That way at the call site users would need to specify
```py
sacct(["--noheader", ...], raise_on_err=True)
```
just to make it blisteringly clear to other future readers where we want the commands to raise and where we want the commands to silently fail.

MattToast (changes requested):
Two potential change requests that I wanted to get your opinion on, but otherwise this LGTM!

smartsim/_core/launcher/slurm/slurmCommands.py
```diff
@@ -120,3 +110,18 @@ def _find_slurm_command(cmd: str) -> str:
...
+
+
+def _execute_slurm_cmd(
+    command: str, args: t.List[str], raise_on_err: bool = False
+) -> t.Tuple[int, str, str]:
+    cmd_exe = _find_slurm_command(command)
+    cmd = [cmd_exe] + args
+    returncode, out, error = execute_cmd(cmd)
+    # Only raise if there was no output
+    if error:
```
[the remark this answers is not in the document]
al-rigazzi: Yeah, I think your way is safer.

smartsim/_core/launcher/slurm/slurmCommands.py
al-rigazzi: OK, after a first moment of confusion, I think I followed your suggestion now.

MattToast: Looks fantastic! Thanks for humoring this change even for now even if we decide to rm JIT dependency checking in the near future!

MattToast (approved):
LGTM pending a re-run of tests after the merge of the https://github.com/CrayLabs/SmartRedis/pull/421 hotfix!

al-rigazzi:
> LGTM pending a re-run of tests after the merge of the [CrayLabs/SmartRedis#421](https://github.com/CrayLabs/SmartRedis/pull/421) hotfix!

OK thanks, I'll let other - more urgent - PRs get in before this.

PR #394, opened 2023-10-12
Expose first_device parameter for setting models, scripts, functions

billschereriii:


tests/backends/test_dbmodel.py
```diff
@@ -487,7 +495,8 @@ def test_colocated_db_model_pytorch(fileutils, wlmutils, mlutils):
                             "TORCH",
                             model_path=model_file,
                             device=test_device,
-                            devices_per_node=test_num_gpus)
+                            devices_per_node=test_num_gpus
```
al-rigazzi, line 498: Missing comma at the end of this line

smartsim/_core/entrypoints/colocated.py
```diff
@@ -100,8 +101,9 @@ def launch_db_model(client: Client, db_model: t.List[str]) -> str:
             name=name,
             model_file=args.file,
             backend=args.backend,
-            fist_gpu=0,
+            first_gpu=args.first_device,
             num_gpus=args.devices_per_node,
+            first_gpu=args.first_device,
```
al-rigazzi, line 106: Repeated `first_gpu` argument at lines 104 and 106

doc/changelog.rst
```diff
@@ -19,11 +19,15 @@ To be released at some future point in time
...
 
+- Expose first_device parameter for scripts, functions, models
 - Added support for MINBATCHTIMEOUT in model execution
 - Remove support for RedisAI 1.2.5, use RedisAI 1.2.7 commit
 
 Detailed Notes
 
+- Added support for first_device parameter for scripts, functions,
+  and models. This causes them to be loaded to the first num_devices
+  beginning with first_device (PR389_)
```
al-rigazzi, line 30: We may want to merge `develop` and check that your entries are correctly added above Julia's.

al-rigazzi (commented):
Everything looks good, just a couple minor fixes to make sure all checks pass.

smartsim/_core/entrypoints/colocated.py
billschereriii: Good catch!

al-rigazzi (approved):
LGTM! This addition will empower users on new-gen systems, super-exciting! Thanks for doing this!

smartsim/entity/dbobject.py
```diff
@@ -118,17 +120,20 @@ def _enumerate_devices(self) -> t.List[str]:
 
         if self.device == "GPU" and self.devices_per_node > 1:
             return [
-                f"{self.device}:{str(device_num)}"
+                f"{self.device}:{str(device_num + self.first_device)}"
                 for device_num in range(self.devices_per_node)
             ]
```
MattToast, line 125: Pedantic style thing: You can save the redundant binary ops by supplying a second arg to range. Also, not from this PR, but there is no need for the explicit cast to `str`
```py
[
    f"{self.device}:{device_num}"
    for device_num in range(self.first_device, self.first_device + self.devices_per_node)
]
```

MattToast (changes requested):
Looks great! A couple of pedantic style things, and called out some points where I wanted to check my understanding or bring something to attention, but otherwise looks good to go!

tests/backends/test_dbscript.py
```diff
@@ -591,8 +608,9 @@ def test_inconsistent_params_db_script(fileutils):
             script_path=torch_script,
             device="CPU",
             devices_per_node=2,
+            first_device=0,
```
MattToast, line 611: Could we add one more check to this test? Basically just check to make sure that an `SSUnusportedError` is raised when `device="CPU"` & `first_device` is anything other than 0 like like we do when `deivce="CPU"` & `devices_per_node` is anything other than 1.
```py
with pytest.raises(SSUnsupportedError) as ex:
    db_script = DBScript(
         name="test_script_db",
        script_path=torch_script,
        device="CPU",
        devices_per_node=1,
        first_device=123,
    )
assert ex.value.args[0] == "Optional: Check for some msg here"

MattToast (approved):
Looks great! One last request for a new test, but otherwise LGTM!

PR #397, opened 2023-10-17
Add ability to passthrough to CLI plugin

ankona:


AlyssaCote:
Are there any dashboard tests in here?

smartsim/_core/_cli/plugin.py
```diff
@@ -0,0 +1,37 @@
...
+                stdout, _ = process.communicate()
+                print(stdout.decode("utf-8"))
+
+            return process.returncode
+
+    return process_execute
+
+
+def dashboard() -> MenuItemConfig:
+    return MenuItemConfig(
```
MattToast, line 29: I'm not sure that I like that first class commands are the same "type" as plugins... What do you think about this potential refactor:

```py
_TMenuItemHandler = Callable[[Namespace], int]
_TPluginItemHandler = Callable[[Namespace, t.List[str]], int]

@dataclass(frozen=True)
class MenuItemConfig:
    command: str
    description: str
    handler: _TMenuItemHandler
    configurator: t.Optional[_CliParserConfigurator] = None

    def execute(self, parser: argparse.ArgumentParser, args: Sequence[str]) -> None:
        ns = parser.parse_args(args)
        return self.handler(ns)


# TODO: I've not fully decided if these two classes should have
#     some common base class or protocol or something for
#     typing/static analysis purposes. I'll leave that to PR
#     author's discretion


@dataclass(frozen=True)
class MenuPluginConfig:
    command: str
    description: str
    handler: _TPluginItemHandler
    configurator: t.Optional[_CliParserConfigurator] = None  # <-- TODO: not strictly sure
                                                             #           if this is needed

    def execute(self, parser: argparse.ArgumentParser, args: Sequence[str]) -> None:
        ns, unparsed = parser.parse_known_args(args)
        # TODO: `ns` will always be empty if we do not allow for a
        #             `configurator`. Feel free to omit this line and just
        #             pass `args` directly if we do not need it!
        return self.handler(ns, unparsed)
```
And then in `SmartCli` we can remove the `SmartCli().args` attr (from what I can tell its not used outside of the `Smart().execute()` method anyway) and simply remove the `if menu_item.is_plugin` check and just directly call
```py
menu_item.execute(self.parser, app_args)
```
?

Pros: I _really_ like that first class items and plugins can be differentiated by type
Cons: Adds a new class/makes the whole CLI construction a hair more difficult :(

smartsim/_core/_cli/cli.py
```diff
@@ -73,14 +74,34 @@ def execute(self, cli_args: t.List[str]) -> int:
...
+        return menu_item.handler(self.args, unparsed_args)
+
+    def _register_plugin(self, item: MenuItemConfig) -> None:
+        parser = self.subparsers.add_parser(
+            item.command, description=item.description, help=item.description
+        )
+        if item.configurator:
+            item.configurator(parser)
+
+        self.menu[item.command] = item
```
MattToast, line 100: Can add a check to the nature of:
```py
if item.commnad in self.menu:
    raise ValueError("Cannot overwrite existing CLI command")
```
here to make sure a plugin does not accidentally overwrite another plugin (or worse a first class command)?

smartsim/_core/_cli/cli.py
```diff
@@ -73,14 +74,34 @@ def execute(self, cli_args: t.List[str]) -> int:
...
+
+    def register_plugins(self) -> None:
+        for plugin in plugins:
+            self._register_plugin(plugin())
```
MattToast, line 104: These two methods look like they work pretty similar to what the `SmartCli.__init__` method does. Could we change their names to `_register_cli_item` and `register_cli_items` (or something similar) respectively and utilize them in the `__init__` to remove the "duplication"?

smartsim/_core/_cli/cli.py
```diff
@@ -73,14 +74,34 @@ def execute(self, cli_args: t.List[str]) -> int:
...
-        app_args = cli_args[1:]
-        self.args = self.parser.parse_args(app_args)
+        app_args = cli_args[1:]  # exclude the path to executable
+        subcommand = cli_args[1]  # first positional arg is the subcommand
 
-        if not (menu_item := self.menu.get(app_args[0], None)):
+        menu_item = self.menu.get(subcommand, None)
+        if not menu_item:
             self.parser.print_help()
             return 0
```
MattToast, line 83: Not part of this PR, but noticed it and wanted your opinion while we are here: I'm not sure that I like that if a user requests a sub-command that does not exist, a successful return code is given back. Thoughts on making this a non-zero return code (I think `argparse` uses 2 by default, but I would need to check)

smartsim/_core/_cli/cli.py
```diff
@@ -73,14 +74,34 @@ def execute(self, cli_args: t.List[str]) -> int:
...
 
-        app_args = cli_args[1:]
-        self.args = self.parser.parse_args(app_args)
+        app_args = cli_args[1:]  # exclude the path to executable
+        subcommand = cli_args[1]  # first positional arg is the subcommand
```
MattToast, line 78: Thanks for these comments! Made it much easier to figure out what the slice operation was meant for!!

MattToast (changes requested):
Looks amazing! I threw one potential refactor your way that I wanted your two-cents on, but otherwise this looks good to go!

smartsim/_core/_cli/cli.py
ankona: If we're automating calls to the CLI, I suppose we can't tell that it failed to do something because it gave back a zero when given a bad arg.... That seals it for me, tbh.

I agree that it should be non-zero. I'll give back 2 to be consistent w/argparse. Great catch.

smartsim/_core/_cli/cli.py
ankona: Fantastic idea. Done!

smartsim/_core/_cli/plugin.py
```diff
@@ -0,0 +1,37 @@
+import argparse
+import shutil
+import subprocess as sp
+import typing as t
+
+from smartsim._core._cli.utils import MenuItemConfig
+
+
+plugins: t.List[t.Callable[[], MenuItemConfig]] = []
```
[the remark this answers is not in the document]
ankona: +1 on split declaration. removed separate instantiation line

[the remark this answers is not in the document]
ankona: swapped to tuple

smartsim/_core/_cli/cli.py
ankona: I honestly said the same thing myself but err'd on the side of doing fewer changes to existing code. Done!

smartsim/_core/_cli/plugin.py
MattToast: Per offline conversation, we have decided to kick this into a new ticket

MattToast: @MattToast ^^ Make sure to do this since you made the suggestion!

MattToast (approved):
LGTM pending test!! Super exciting to see a SmartSim plugin ecosystem come together!

smartsim/_core/_cli/plugin.py
ankona: Created ticket 568

PR #400, opened 2023-10-19
Serialize launched entities manifest

ankona:
Bare bones, rough draft implementation of how SmartSim may be able to serialize information about launched entities and/or collections of entities and dump that information to a file so that external or third party tools may be able to display and monitor jobs outside of the process running the python interpreter that ran a driver script.

So far this has been tested by hand and seems to be working for local and slurm launchers for models/ensembles/databases on both interactive and batch jobs.

TODO:
- [x] TESTS!! (unit and integration)
- [x] Run against other WLMs (mainly PBS)
- [x] Address and remove numerous ``TODO`` and ``FIXME`` comments

### DIFF
--- a/smartsim/_core/utils/serialize.py
+++ b/smartsim/_core/utils/serialize.py
@@ -31,7 +31,7 @@ TStepLaunchMetaData = t.Tuple[
 
 
 def save_launch_manifest(manifest: _Manifest[TStepLaunchMetaData]) -> None:
-    manifest_dir = Path(manifest.metadata.exp_path) / ".smartsim/manifest"
+    manifest_dir = Path(manifest.metadata.exp_path) / ".smartsim/telemetry"
     manifest_dir.mkdir(parents=True, exist_ok=True)
     manifest_file = manifest_dir / "manifest.json"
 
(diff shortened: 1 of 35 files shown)


smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+            "Symlink": [],
+            "Configure": [],
+            "Copy": [],
+        },
+        "colocated_db": {
+            "settings": colo_settings,
+            "scripts": [
+                {
+                    script.name: {
+                        "backend": "TORCH",
```
ankona, line 139: is this always torch?

ankona (commented):
LGTM! As discussed, version verification will happen later.

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+            _dictify_ensemble(ens, member_info, telemetry_data_root / "ensemble")
+            for ens, member_info in manifest.ensembles
+        ],
+    }
+    try:
+        with open(manifest_file, "r", encoding="utf-8") as file:
+            manifest_dict = json.load(file)
+    except (FileNotFoundError, json.JSONDecodeError):
+        manifest_dict = {
+            "schema info": {
```
AlyssaCote, line 84: Thank you for remembering to do this!!

AlyssaCote: Because I totally forgot.

AlyssaCote (approved):
Great job, Matt! It looks great!

smartsim/_core/utils/serialize.py
MattToast: We resolved this offline, but wanted to make sure is is documented here for any future readers!

At time of writing `DBScript` at the moment only refers to TorchScripts, and as such the `DBScript` has no `backend` attr because we assume  that it must be `TORCH`. This may change in the future, but for now it is the current state of the world.

PR #406, opened 2023-10-30
Fixes to multi-db introspection

al-rigazzi:
This PR fixes a bug encountered when launching more than one DB from a driver script. Addresses of running DBs could get mixed and some internal checks would incorrectly fail.


MattToast (approved):
LGTM! Thanks for catching this!!

PR #415, opened 2023-11-10
Pin version of types-tensorflow

al-rigazzi:
With the last update of `types-tensorflow` (2.12.0.10), some types are not defined anymore, and this causes mypy to fail.

This PR pins the version number to `2.12.0.9`, which is the last compatible with TF `2.8.x`.


ankona (approved):
LGTM

MattToast (approved):
LGTM! Thanks for the catch!!

PR #417, opened 2023-11-13
Split tests into groups for parallel execution in CI/CD runners

billschereriii:
Split tests into groups for parallel execution in CI/CD runners
Remove some unused import statements

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -153,5 +153,5 @@ jobs:
       - name: Upload Pytest coverage to Codecov
         uses: codecov/codecov-action@v2
         with:
-          fail_ci_if_error: true
+          fail_ci_if_error: false
           files: ./coverage.xml
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -44,11 +44,12 @@ env:
 
 jobs:
   run_tests:
-    name: Run tests with ${{ matrix.os }}, Python ${{ matrix.py_v}}, RedisAI ${{ matrix.rai }}
+    name: Run tests ${{ matrix.subset }} with ${{ matrix.os }}, Python ${{ matrix.py_v}}, RedisAI ${{ matrix.rai }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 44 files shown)


al-rigazzi (commented):
I like the divide-and-conquer approach. Nevertheless, the probability of hitting the pre-existing issue with Codecov is multiplied by a factor of ~3, thus I'd suggest we decide whether we are going to mark as passed the action even if codecov failed and, if that's the case, we do it as part of this PR.

billschereriii:
Looks like all the tests ran in between 9 to 18 minutes, btw

billschereriii:
Note: subdirectories `full_wlm` and `on_wlm` are intentionally skipped for CI/CD as we aren't running on HPC systems for these tests

ankona:
Per @al-rigazzi comment about codecov... Do we want to do a minor rearrange of the build to avoid 3x the uploads? Perhaps we could:

1. run independent test step
2. store each codecov report as an artifact in a test step
3. split code coverage upload to run into step successor of (1)
4. [retrieve](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#downloading-or-deleting-artifacts) all artifacts and upload only one time

.github/workflows/run_tests.yml
```diff
@@ -44,11 +44,12 @@ env:
 
 jobs:
   run_tests:
-    name: Run tests with ${{ matrix.os }}, Python ${{ matrix.py_v}}, RedisAI ${{ matrix.rai }}
+    name: Run tests ${{ matrix.subset }} with ${{ matrix.os }}, Python ${{ matrix.py_v}}, RedisAI ${{ matrix.rai }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        subset: [backends, slow_tests, group_a, group_b]
```
ankona, line 52: Other than the opaque group names, this is simple & effective. LGTM as the first step toward test domination!

ankona (approved):
LGTM. Approving & assuming you'll want to split out anything about rearranging to upload codecov results fewer times into a new ticket.

PR #418, opened 2023-11-13
Allow TM to track Unmanaged Steps through WLM

MattToast:
Moves logic for launching a process through a the indirect module into a dedicated proxy step ~class~ decorator, so that unmanaged steps can be tracked by the TM without updating the logic to existing steps or launchers.


tests/test_pals_settings.py
```diff
@@ -39,6 +39,15 @@
 default_exe = sys.executable
 default_kwargs = {"fail_if_missing_exec": False}
 
+@pytest.fixture(autouse=True)
+def turnoff_telemetry_inderict(monkeypatch):
```
ankona, line 43: interdict?

MattToast: believe it or not it was supposed to be _indirect_ 🤦 

Spelling is hard :(

smartsim/_core/launcher/launcher.py
```diff
@@ -88,6 +88,7 @@ def supported_rs(self) -> t.Dict[t.Type[SettingsBase], t.Type[Step]]:
 
     # every launcher utilizing this interface must have a map
     # of supported RunSettings types (see slurmLauncher.py for ex)
+    @t.final
```
MattToast, line 91: This is no longer needed as there is now no special logic when creating a step to prepare for it to be sent through the proxy. Remove this so that subclasses can extend this method (should they choose)

tests/test_pals_settings.py
```diff
@@ -39,6 +39,16 @@
 default_exe = sys.executable
 default_kwargs = {"fail_if_missing_exec": False}
 
+
+@pytest.fixture(autouse=True)
+def turn_off_telemetry_indirect(monkeypatch):
+    import smartsim._core.launcher.launcher
```
MattToast, line 45: Why did I import this??

smartsim/_core/launcher/step/localStep.py
```diff
@@ -37,8 +37,13 @@ class LocalStep(Step):
     def __init__(self, name: str, cwd: str, run_settings: RunSettings):
         super().__init__(name, cwd, run_settings)
         self.run_settings = run_settings
-        self.env = self._set_env()
+        self._env = self._set_env()
 
+    @property
+    def env(self) -> t.Dict[str, str]:
```
ankona, line 43: I notice that this override returns a `dict` but the base class has `optional[dict]`. This seems like it might break the LSP and result in potential null object errors.

Should we make them consistent?

MattToast: It should be okay for a method on a subclass to return a subclass of the expected type. All instances of type `T` are implicitly `T | None` as well (they just happen to never be `None`).

I chose to narrow the type for mainly as a convenience for developers who happen to know for fact they are working with a `LocalStep`. For e.g.
```py
step: Step = ...
step.env.items()  # <-- mypy should error here saying "type `None` has no method `items`"

known_local_step: LocalStep = ...
known_local_step.env.items()  # <-- mypy is happy bc it knows that this is fine
                              #      and devs do not need to make a "is not none" check here
 ```

This would break and going the other direction though (returning a `T | None` where a `T` is expected) and mypy _should_ be able to report that error.

ankona: Great! I forgot that mypy would save me from myself. TYS!

ankona (approved):

PR #420, opened 2023-11-14
SmartDashboard docs

AlyssaCote:
I added a smartdashboard.rst file to kickstart the SmartDashboard docs within SmartSim.


doc/smartdashboard.rst
```diff
@@ -0,0 +1,7 @@
+
+**********
```
al-rigazzi, line 2: The decorator `**********` should be at least as long as the text which is enclosing

al-rigazzi (changes requested):
Just a change requested, otherwise LGTM

doc/smartdashboard.rst
ankona: +1 - fixed!

al-rigazzi (approved):
LGTM!

ankona (commented):
Moved main body of docs to other repo and tweaked SmartSim `make docks` to include those instead.

PR #425, opened 2023-11-28
Fix CLI bug hiding arguments matching root parser

ankona:
Discovered a bug where using the CLI "plugin" to execute smart-dash did not return the correct help text. e.g. `smart dashboard -h` returned the default help text. This fix corrects the failure to pass arguments known by the smart CLI parser to the plugin.

After fixing the initial defect:

- A bug hiding the final output of the plugin was discovered. Moved `print(plugin_stdout)` to ensure plugin stdout was displayed
- Fixed bug where tests run in vscode test runner fail to load plugin. Mitigated with importlib
- Added test to verify passthrough output for initial fix
- Added test to verify outputs when plugin is not installed
- Revert removal of positional-only arguments

### DIFF
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -357,7 +357,7 @@ def _format_incompatible_python_env_message(
 
 
 def execute(
-    args: argparse.Namespace, _unparsed_args: t.Optional[t.List[str]] = None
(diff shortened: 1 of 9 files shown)


MattToast (approved):
LGTM!!

PR #426, opened 2023-11-29
Add support for producing dashboard outputs

ankona:
Add support for producing & consuming dashboard outputs.

- Adds `telemetry monitor` to check for updates and produce events for the dashboard
- Updates `controller` to conditionally start `telemetry monitor`
- Updates `controller` to produce a runtime manifest to trigger telemetry collection
- Adds `indirect proxy` to produce events for the dashboard for unmanaged tasks
- Adds CLI capability to launch dashboard

### DIFF
--- a/smartsim/_core/_cli/__main__.py
+++ b/smartsim/_core/_cli/__main__.py
@@ -27,12 +27,28 @@
 import sys
 
 from smartsim._core._cli.cli import default_cli
+from smartsim._core._cli.utils import SMART_LOGGER_FORMAT
+from smartsim.error.errors import SmartSimCLIActionCancelled
+from smartsim.log import get_logger
+
+
+logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 
 def main() -> int:
     smart_cli = default_cli()
 
-    return smart_cli.execute(sys.argv)
+    try:
+        return smart_cli.execute(sys.argv)
+    except SmartSimCLIActionCancelled as ssi:
+        logger.debug(ssi, exc_info=True)
+        logger.info(ssi)
+    except KeyboardInterrupt:
+        msg = "SmartSim was terminated by user"
+        logger.debug(msg, exc_info=True)
+        logger.info(msg)
+
+    return 0
 
 
 if __name__ == "__main__":
--- a/smartsim/_core/_cli/__main__.py
+++ b/smartsim/_core/_cli/__main__.py
@@ -38,16 +38,16 @@ logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 def main() -> int:
     smart_cli = default_cli()
-    excpetion_trace_back_msg = "SmartSim exited with the following exception info:"
+    exception_trace_back_msg = "SmartSim exited with the following exception info:"
 
     try:
         return smart_cli.execute(sys.argv)
     except SmartSimCLIActionCancelled as ssi:
         logger.info(str(ssi))
-        logger.debug(excpetion_trace_back_msg, exc_info=ssi)
+        logger.debug(exception_trace_back_msg, exc_info=ssi)
     except KeyboardInterrupt as e:
         logger.info("SmartSim was terminated by user")
-        logger.debug(excpetion_trace_back_msg, exc_info=e)
+        logger.debug(exception_trace_back_msg, exc_info=e)
     return os.EX_OK
 
 
--- a/smartsim/_core/_cli/__main__.py
+++ b/smartsim/_core/_cli/__main__.py
@@ -24,6 +24,7 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import os
 import sys
 
 from smartsim._core._cli.cli import default_cli
@@ -47,8 +48,7 @@ def main() -> int:
         msg = "SmartSim was terminated by user"
         logger.debug(msg, exc_info=True)
         logger.info(msg)
-
-    return 0
+    return os.EX_OK
 
 
 if __name__ == "__main__":
--- a/smartsim/_core/_cli/__main__.py
+++ b/smartsim/_core/_cli/__main__.py
@@ -31,6 +31,7 @@ from smartsim._core._cli.cli import default_cli
 
 def main() -> int:
     smart_cli = default_cli()
+
     return smart_cli.execute(sys.argv)
 
 
--- a/smartsim/_core/_cli/__main__.py
+++ b/smartsim/_core/_cli/__main__.py
@@ -38,16 +38,16 @@ logger = get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 def main() -> int:
     smart_cli = default_cli()
+    excpetion_trace_back_msg = "SmartSim exited with the following exception info:"
 
     try:
         return smart_cli.execute(sys.argv)
     except SmartSimCLIActionCancelled as ssi:
-        logger.debug(ssi, exc_info=True)
-        logger.info(ssi)
-    except KeyboardInterrupt:
-        msg = "SmartSim was terminated by user"
-        logger.debug(msg, exc_info=True)
-        logger.info(msg)
+        logger.info(str(ssi))
+        logger.debug(excpetion_trace_back_msg, exc_info=ssi)
+    except KeyboardInterrupt as e:
+        logger.info("SmartSim was terminated by user")
+        logger.debug(excpetion_trace_back_msg, exc_info=e)
     return os.EX_OK
 
 
--- a/smartsim/_core/_cli/cli.py
+++ b/smartsim/_core/_cli/cli.py
@@ -52,7 +52,6 @@ class SmartCli:
             prog="smart",
             description="SmartSim command line interface",
         )
-        self.args: t.Optional[argparse.Namespace] = None
 
         self.subparsers = self.parser.add_subparsers(
             dest="command",
@@ -77,13 +76,15 @@ class SmartCli:
             self.parser.print_help()
             return 2
 
-        if menu_item.is_plugin:
-            self.args, unparsed_args = self.parser.parse_known_args(app_args)
-        else:
-            unparsed_args = []
-            self.args = self.parser.parse_args(app_args)
+        args = argparse.Namespace()
+        unparsed_args = []
 
-        return menu_item.handler(self.args, unparsed_args)
+        if menu_item.is_plugin:
+            unparsed_args = app_args[1:]
+        else:
+            args = self.parser.parse_args(app_args)
+
+        return menu_item.handler(args, unparsed_args)
 
     def _register_menu_item(self, item: MenuItemConfig) -> None:
         parser = self.subparsers.add_parser(
--- a/smartsim/_core/_cli/cli.py
+++ b/smartsim/_core/_cli/cli.py
@@ -27,6 +27,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import argparse
+import os
 import typing as t
 
 from smartsim._core._cli.build import configure_parser as build_parser
@@ -66,7 +67,7 @@ class SmartCli:
     def execute(self, cli_args: t.List[str]) -> int:
         if len(cli_args) < 2:
             self.parser.print_help()
-            return 2
+            return os.EX_USAGE
 
         app_args = cli_args[1:]  # exclude the path to executable
         subcommand = cli_args[1]  # first positional arg is the subcommand
@@ -74,7 +75,7 @@ class SmartCli:
         menu_item = self.menu.get(subcommand, None)
         if not menu_item:
             self.parser.print_help()
-            return 2
+            return os.EX_USAGE
 
         args = argparse.Namespace()
         unparsed_args = []
--- a/smartsim/_core/_cli/cli.py
+++ b/smartsim/_core/_cli/cli.py
@@ -41,46 +41,67 @@ from smartsim._core._cli.validate import (
     execute as validate_execute,
     configure_parser as validate_parser,
 )
+from smartsim._core._cli.plugin import plugins
 from smartsim._core._cli.utils import MenuItemConfig
 
 
 class SmartCli:
     def __init__(self, menu: t.List[MenuItemConfig]) -> None:
-        self.menu: t.Dict[str, MenuItemConfig] = {item.command: item for item in menu}
-        parser = argparse.ArgumentParser(
+        self.menu: t.Dict[str, MenuItemConfig] = {}
+        self.parser = argparse.ArgumentParser(
             prog="smart",
             description="SmartSim command line interface",
         )
-        self.parser = parser
         self.args: t.Optional[argparse.Namespace] = None
 
-        subparsers = parser.add_subparsers(
+        self.subparsers = self.parser.add_subparsers(
             dest="command",
             required=True,
             metavar="<command>",
             help="Available commands",
         )
 
-        for cmd, item in self.menu.items():
-            parser = subparsers.add_parser(
-                cmd, description=item.description, help=item.description
-            )
-            if item.configurator:
-                item.configurator(parser)
+        self.register_menu_items(menu)
+        self.register_menu_items([plugin() for plugin in plugins])
 
     def execute(self, cli_args: t.List[str]) -> int:
         if len(cli_args) < 2:
             self.parser.print_help()
-            return 0
+            return 2
 
-        app_args = cli_args[1:]
-        self.args = self.parser.parse_args(app_args)
+        app_args = cli_args[1:]  # exclude the path to executable
+        subcommand = cli_args[1]  # first positional arg is the subcommand
 
-        if not (menu_item := self.menu.get(app_args[0], None)):
+        menu_item = self.menu.get(subcommand, None)
+        if not menu_item:
             self.parser.print_help()
-            return 0
+            return 2
 
-        return menu_item.handler(self.args)
+        if menu_item.is_plugin:
+            self.args, unparsed_args = self.parser.parse_known_args(app_args)
+        else:
+            unparsed_args = []
+            self.args = self.parser.parse_args(app_args)
+
+        return menu_item.handler(self.args, unparsed_args)
+
+    def _register_menu_item(self, item: MenuItemConfig) -> None:
+        parser = self.subparsers.add_parser(
+            item.command, description=item.description, help=item.description
+        )
+        if item.configurator:
+            item.configurator(parser)
+
+        if item.command in self.menu:
+            raise ValueError(
+                f"{item.command} cannot overwrite existing CLI command"
+            )
+
+        self.menu[item.command] = item
+
+    def register_menu_items(self, menu_items: t.List[MenuItemConfig]) -> None:
+        for item in menu_items:
+            self._register_menu_item(item)
 
 
 def default_cli() -> SmartCli:
--- a/smartsim/_core/_cli/dbcli.py
+++ b/smartsim/_core/_cli/dbcli.py
@@ -31,7 +31,7 @@ from smartsim._core._cli.utils import get_db_path
 
 
 def execute(
-    _args: argparse.Namespace, _unparsed_args: t.Optional[t.List[str]] = None
+    _args: argparse.Namespace, _unparsed_args: t.Optional[t.List[str]] = None, /
 ) -> int:
     if db_path := get_db_path():
         print(db_path)
--- a/smartsim/_core/_cli/dbcli.py
+++ b/smartsim/_core/_cli/dbcli.py
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import argparse
+import os
 import typing as t
 
 from smartsim._core._cli.utils import get_db_path
@@ -35,6 +36,6 @@ def execute(
 ) -> int:
     if db_path := get_db_path():
         print(db_path)
-        return 0
+        return os.EX_OK
     print("Database (Redis or KeyDB) dependencies not found")
-    return 1
+    return os.EX_SOFTWARE
--- a/smartsim/_core/_cli/dbcli.py
+++ b/smartsim/_core/_cli/dbcli.py
@@ -25,11 +25,14 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import argparse
+import typing as t
 
 from smartsim._core._cli.utils import get_db_path
 
 
-def execute(_args: argparse.Namespace) -> int:
+def execute(
+    _args: argparse.Namespace, _unparsed_args: t.Optional[t.List[str]] = None
+) -> int:
     if db_path := get_db_path():
         print(db_path)
         return 0
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -5,8 +5,11 @@ import sys
 import subprocess as sp
 import typing as t
 
-from smartsim._core._cli.utils import MenuItemConfig
+import smartsim.log
+from smartsim._core._cli.utils import MenuItemConfig, SMART_LOGGER_FORMAT
 from smartsim.error.errors import SmartSimCLIActionCancelled
+
+_LOGGER = smartsim.log.get_logger("Smart", fmt=SMART_LOGGER_FORMAT)
 
 
 def dynamic_execute(
@@ -15,29 +18,22 @@ def dynamic_execute(
     def process_execute(
         _args: argparse.Namespace, unparsed_args: t.List[str], /
     ) -> int:
-        not_found = f"{cmd} plugin not found. Please ensure it is installed"
         try:
             spec = importlib.util.find_spec(cmd)
             if spec is None:
                 raise AttributeError()
         except (ModuleNotFoundError, AttributeError):
-            print(not_found)
+            _LOGGER.error(f"{cmd} plugin not found. Please ensure it is installed")
             return os.EX_CONFIG
 
         combined_cmd = [sys.executable, "-m", cmd] + unparsed_args
 
         try:
-            with sp.Popen(combined_cmd, stdout=sp.PIPE, stderr=sp.PIPE) as process:
-                stdout, _ = process.communicate()
-                while process.returncode is None:
-                    stdout, _ = process.communicate()
-
-                plugin_stdout = stdout.decode("utf-8")
-                print(plugin_stdout)
-                return process.returncode
+            completed_proc = sp.run(combined_cmd)
         except KeyboardInterrupt as ex:
             msg = f"{plugin_name} terminated by user"
             raise SmartSimCLIActionCancelled(msg) from ex
+        return completed_proc.returncode
 
     return process_execute
 
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -5,9 +5,12 @@ import subprocess as sp
 import typing as t
 
 from smartsim._core._cli.utils import MenuItemConfig
+from smartsim.error.errors import SmartSimCLIActionCancelled
 
 
-def dynamic_execute(cmd: str) -> t.Callable[[argparse.Namespace, t.List[str]], int]:
+def dynamic_execute(
+    cmd: str, plugin_name: str
+) -> t.Callable[[argparse.Namespace, t.List[str]], int]:
     def process_execute(
         _args: argparse.Namespace, unparsed_args: t.List[str], /
     ) -> int:
@@ -21,14 +24,19 @@ def dynamic_execute(cmd: str) -> t.Callable[[argparse.Namespace, t.List[str]], i
             return 1
 
         combined_cmd = [sys.executable, "-m", cmd] + unparsed_args
-        with sp.Popen(combined_cmd, stdout=sp.PIPE, stderr=sp.PIPE) as process:
-            stdout, _ = process.communicate()
-            while process.returncode is None:
-                stdout, _ = process.communicate()
 
-            plugin_stdout = stdout.decode("utf-8")
-            print(plugin_stdout)
-            return process.returncode
+        try:
+            with sp.Popen(combined_cmd, stdout=sp.PIPE, stderr=sp.PIPE) as process:
+                stdout, _ = process.communicate()
+                while process.returncode is None:
+                    stdout, _ = process.communicate()
+
+                plugin_stdout = stdout.decode("utf-8")
+                print(plugin_stdout)
+                return process.returncode
+        except KeyboardInterrupt as ex:
+            msg = f"{plugin_name} terminated by user"
+            raise SmartSimCLIActionCancelled(msg) from ex
 
     return process_execute
 
@@ -37,7 +45,7 @@ def dashboard() -> MenuItemConfig:
     return MenuItemConfig(
         "dashboard",
         "Start the SmartSim dashboard",
-        dynamic_execute("smartdashboard.Experiment_Overview"),
+        dynamic_execute("smartdashboard.Experiment_Overview", "Dashboard"),
         is_plugin=True,
     )
 
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -1,0 +1,33 @@
+import argparse
+import shutil
+import subprocess as sp
+import typing as t
+
+from smartsim._core._cli.utils import MenuItemConfig
+
+
+def dynamic_execute(cmd: str) -> t.Callable[[argparse.Namespace, t.List[str]], int]:
+    def process_execute(_args: argparse.Namespace, unparsed_args: t.List[str]) -> int:
+        if not shutil.which(cmd):
+            raise ValueError(f"{cmd} plugin not found. Please ensure it is installed")
+        combined_cmd = [cmd] + unparsed_args
+        with sp.Popen(combined_cmd, stdout=sp.PIPE, stderr=sp.PIPE) as process:
+            stdout, _ = process.communicate()
+            while process.returncode is None:
+                stdout, _ = process.communicate()
+                print(stdout.decode("utf-8"))
+
+            return process.returncode
+
+    return process_execute
+
+
+def dashboard() -> MenuItemConfig:
+    return MenuItemConfig(
+        "dashboard",
+        "Start the SmartSim dashboard",
+        dynamic_execute("smart-dash"),
+        is_plugin=True,
+    )
+
+plugins = (dashboard, )
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -1,5 +1,6 @@
 import argparse
 import importlib
+import os
 import sys
 import subprocess as sp
 import typing as t
@@ -21,7 +22,7 @@ def dynamic_execute(
                 raise AttributeError()
         except (ModuleNotFoundError, AttributeError):
             print(not_found)
-            return 1
+            return os.EX_CONFIG
 
         combined_cmd = [sys.executable, "-m", cmd] + unparsed_args
 
@@ -44,7 +45,12 @@ def dynamic_execute(
 def dashboard() -> MenuItemConfig:
     return MenuItemConfig(
         "dashboard",
-        "Start the SmartSim dashboard",
+        (
+            "Start the SmartSim dashboard to monitor experiment output from a "
+            "graphical user interface. This requires that the SmartSim Dashboard "
+            "Package be installed. For more infromation please visit "
+            "https://github.com/CrayLabs/SmartDashboard"
+        ),
         dynamic_execute("smartdashboard.Experiment_Overview", "Dashboard"),
         is_plugin=True,
     )
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -1,5 +1,5 @@
 import argparse
-import importlib
+import importlib.util
 import os
 import sys
 import subprocess as sp
@@ -21,7 +21,7 @@ def dynamic_execute(
         try:
             spec = importlib.util.find_spec(cmd)
             if spec is None:
-                raise AttributeError()
+                raise AttributeError
         except (ModuleNotFoundError, AttributeError):
             _LOGGER.error(f"{cmd} plugin not found. Please ensure it is installed")
             return os.EX_CONFIG
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -29,7 +29,7 @@ def dynamic_execute(
         combined_cmd = [sys.executable, "-m", cmd] + unparsed_args
 
         try:
-            completed_proc = sp.run(combined_cmd)
+            completed_proc = sp.run(combined_cmd, check=False)
         except KeyboardInterrupt as ex:
             msg = f"{plugin_name} terminated by user"
             raise SmartSimCLIActionCancelled(msg) from ex
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -47,7 +47,7 @@ def dashboard() -> MenuItemConfig:
             "Package be installed. For more infromation please visit "
             "https://github.com/CrayLabs/SmartDashboard"
         ),
-        dynamic_execute("smartdashboard.Experiment_Overview", "Dashboard"),
+        dynamic_execute("smartdashboard", "Dashboard"),
         is_plugin=True,
     )
 
--- a/smartsim/_core/_cli/plugin.py
+++ b/smartsim/_core/_cli/plugin.py
@@ -1,5 +1,6 @@
 import argparse
-import shutil
+import importlib
+import sys
 import subprocess as sp
 import typing as t
 
@@ -7,16 +8,26 @@ from smartsim._core._cli.utils import MenuItemConfig
 
 
 def dynamic_execute(cmd: str) -> t.Callable[[argparse.Namespace, t.List[str]], int]:
-    def process_execute(_args: argparse.Namespace, unparsed_args: t.List[str]) -> int:
-        if not shutil.which(cmd):
-            raise ValueError(f"{cmd} plugin not found. Please ensure it is installed")
-        combined_cmd = [cmd] + unparsed_args
+    def process_execute(
+        _args: argparse.Namespace, unparsed_args: t.List[str], /
+    ) -> int:
+        not_found = f"{cmd} plugin not found. Please ensure it is installed"
+        try:
+            spec = importlib.util.find_spec(cmd)
+            if spec is None:
+                raise AttributeError()
+        except (ModuleNotFoundError, AttributeError):
+            print(not_found)
+            return 1
+
+        combined_cmd = [sys.executable, "-m", cmd] + unparsed_args
         with sp.Popen(combined_cmd, stdout=sp.PIPE, stderr=sp.PIPE) as process:
             stdout, _ = process.communicate()
             while process.returncode is None:
                 stdout, _ = process.communicate()
-                print(stdout.decode("utf-8"))
 
+            plugin_stdout = stdout.decode("utf-8")
+            print(plugin_stdout)
             return process.returncode
 
     return process_execute
@@ -26,8 +37,9 @@ def dashboard() -> MenuItemConfig:
     return MenuItemConfig(
         "dashboard",
         "Start the SmartSim dashboard",
-        dynamic_execute("smart-dash"),
+        dynamic_execute("smartdashboard.Experiment_Overview"),
         is_plugin=True,
     )
 
-plugins = (dashboard, )
+
+plugins = (dashboard,)
--- a/smartsim/_core/config/config.py
+++ b/smartsim/_core/config/config.py
@@ -212,6 +212,10 @@ class Config:
     def telemetry_enabled(self) -> bool:
         return bool(os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1))
 
+    @property
+    def telemetry_cooldown(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_COOLDOWN", 90))
+
 @lru_cache(maxsize=128, typed=False)
 def get_config() -> Config:
     # wrap into a function with a cached result
--- a/smartsim/_core/config/config.py
+++ b/smartsim/_core/config/config.py
@@ -204,6 +204,13 @@ class Config:
         # no account by default
         return os.environ.get("SMARTSIM_TEST_ACCOUNT", None)
 
+    @property
+    def telemetry_frequency(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_FREQUENCY", 5))
+
+    @property
+    def telemetry_enabled(self) -> bool:
+        return bool(os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1))
 
 @lru_cache(maxsize=128, typed=False)
 def get_config() -> Config:
--- a/smartsim/_core/config/config.py
+++ b/smartsim/_core/config/config.py
@@ -210,7 +210,7 @@ class Config:
 
     @property
     def telemetry_enabled(self) -> bool:
-        return bool(os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1))
+        return int(os.environ.get("SMARTSIM_FLAG_TELEMETRY", "0")) > 0
 
     @property
     def telemetry_cooldown(self) -> int:
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -343,6 +343,10 @@ class Controller:
         Orchestrators are always launched first so that the
         address of the database can be given to following entities
 
+        :param exp_name: The name of the launching experiment
+        :type exp_name: str
+        :param exp_path: path to location of ``Experiment`` directory if generated
+        :type exp_path: str
         :param manifest: Manifest of deployables to launch
         :type manifest: Manifest
         """
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -130,7 +130,8 @@ class Controller:
         )
 
         # launch a telemetry monitor to track job progress
-        self.start_telemetry_monitor(exp_path)
+        if CONFIG.telemetry_enabled:
+            self.start_telemetry_monitor(exp_path)
 
         # block until all non-database jobs are complete
         if block:
@@ -814,9 +815,6 @@ class Controller:
 
     def start_telemetry_monitor(self, exp_dir: str) -> None:
         logger.debug("Starting telemetry monitor process")
-        if not CONFIG.telemetry_enabled:
-            return
-
         if (
             self._telemetry_monitor is None
             or self._telemetry_monitor.returncode is not None
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -27,8 +27,11 @@
 from __future__ import annotations
 
 import os.path as osp
+import pathlib
 import pickle
 import signal
+import subprocess
+import sys
 import threading
 import time
 import typing as t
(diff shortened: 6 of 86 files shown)


smartsim/_core/control/controller.py
```diff
@@ -739,13 +812,64 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
...
+            or self._telemetry_monitor.returncode is not None
+        ):
+            cmd = [
+                sys.executable,
+                "-m",
+                "smartsim._core.entrypoints.telemetrymonitor",
+                "-exp_dir",
+                exp_dir,
+                "-frequency",
+                str(frequency),
```
ankona, line 831: do we care about this inconstent config frequency passing (vs Config.telemetry_cooldown below)

smartsim/_core/config/config.py
```diff
@@ -204,6 +204,17 @@ def test_account(self) -> t.Optional[str]:  # pragma: no cover
         # no account by default
         return os.environ.get("SMARTSIM_TEST_ACCOUNT", None)
 
+    @property
+    def telemetry_frequency(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_FREQUENCY", 5))
+
+    @property
+    def telemetry_enabled(self) -> bool:
+        return bool(os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1))
```
MattToast, line 213: Per discussion: Turn this off by default

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+        exp_dir = pathlib.Path(manifest_path).parent.parent.parent
+
+        for run in runs:
+            for entity in run.flatten(
+                filter_fn=lambda e: e.key not in self._tracked_jobs and e.is_managed
+            ):
+                entity.path = str(exp_dir)
+
+                self._tracked_jobs[entity.key] = entity
+                track_event(
```
ankona, line 392: is track_event needed here? the WLM will trigger start/stop & indirect will handle for unmanaged. This should be a no-op in any non-running exp_dir?

MattToast (commented):
Small notes from discussion

smartsim/_core/entrypoints/indirect.py
```diff
@@ -0,0 +1,237 @@
...
+            get_ts(),
+            proxy_pid,
+            "",  # step_id for unmanaged task is always empty
+            etype,
+            "start",
+            status_path,
+            logger,
+            detail=start_detail,
+            return_code=start_rc,
+        )
```
MattToast, line 102: Small fun bug here for the validate CLI:

If a user were to run ``smart validate`` with the TM enabled, this entry-point process could live on beyond the time the validation experiment has concluded. This means that this ``track_event`` call may actually recreate the temporary directory that the validation driver was using as an experiment directory in order to write out the ``stop.json`` file (this was at least consistently the case for me when I ran ``smart validate``).

There is nothing "wrong" with this, but it would be a nice QoL feature if the ``validate`` command did not leave artifacts around if the validation was successful. The easiest ways I can think of to guarantee this would be to either:

1) Automatically turn the TM off if a user runs ``smart validate``, regardless of the ``SMARTSIM_FLAG_TELEMETRY`` env var (this is my preferred solution).
1) Validate that all TM and indirect processes have finished before the ``validate`` driver cleans up the temp exp dir.

smartsim/_core/control/controller.py
ankona: removed frequency param & used `CONFIG.telemetry_frequency`

smartsim/_core/control/controller.py
```diff
@@ -739,13 +812,64 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
...
+        launched_step_name, step = data
+        launched_step_map = launcher.step_mapping[launched_step_name]
+        out_file, err_file = step.get_output_files()
+        return (
+            launched_step_map.step_id,
+            launched_step_map.task_id,
+            launched_step_map.managed,
+            out_file,
+            err_file,
+            pathlib.Path(step.meta["status_dir"]),
```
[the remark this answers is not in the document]
ankona: Replaced with:

```
pathlib.Path(step.meta.get("status_dir", step.cwd)),
```

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,233 @@
...
+
+TStepLaunchMetaData = t.Tuple[
+    t.Optional[str], t.Optional[str], t.Optional[bool], str, str, Path
+]
+TELMON_SUBDIR: t.Final[str] = ".smartsim/telemetry"
+MANIFEST_FILENAME: t.Final[str] = "manifest.json"
+
+
+def save_launch_manifest(manifest: _Manifest[TStepLaunchMetaData]) -> None:
+    manifest.metadata.run_telemetry_subdirectory.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
ankona: Added

smartsim/_core/config/config.py
ankona: Default modified.

smartsim/_core/entrypoints/indirect.py
ankona: Added `_disable_telmon` context manager to disable telemetry monitor for the test.

MattToast (changes requested):
Found some things I would like addressed in the TM tests before we approve for merge

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+    num_db_nodes: int,
+    launcher: str,
+):
+    """Run an experiment with multiple start calls to ensure that telemetry is
+    saved correctly for each run"""
+    test_dir = fileutils.make_test_dir()
+
+    if run_command != "auto" and not shutil.which(run_command):
+        assert True, f"{run_command} not supported on test environment"
+        return
```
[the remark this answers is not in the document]
ankona: 100. Replaced with:

```
    if run_command != "auto" and not shutil.which(run_command):
        pytest.skip(reason=f"{run_command} not supported on test environment")
```

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+        assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+        try:
+            exp.start(ens, block=True, summary=True)
+            assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+        finally:
+            exp.stop(db)
+            assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+            time.sleep(3)  # time for telmon to write db stop event
+
+    assert True, "TODO: check telemetry output"
```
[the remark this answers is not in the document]
ankona: Done & Done!

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+    """Ensure expected output type"""
+    ts = get_ts()
+    assert isinstance(ts, int)
+
+
+@pytest.mark.parametrize(
+    ["etype", "task_id", "step_id", "timestamp", "evt_type"],
+    [
+        pytest.param("ensemble", "", "123", get_ts(), "start", id="start event"),
+        pytest.param("ensemble", "", "123", get_ts(), "start", id="stop event"),
```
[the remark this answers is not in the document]
ankona: 👁️ 👁️ !!

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+    )
+    assert manifest.launcher == "Slurm"
+    assert len(manifest.runs) == 2
+
+    assert len(manifest.runs[0].orchestrators) == 1
+    assert len(manifest.runs[1].models) == 1
+
+
+def test_load_manifest_db_and_models_1run(fileutils: FileUtils):
+    """Ensure that the runtime manifest loads correctly when containing models & orchestrator"""
```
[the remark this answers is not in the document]
ankona: Great catch!

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+    assert "world" in cmd
+
+
+@pytest.mark.parametrize(
+    "run_command, num_db_nodes, launcher",
+    [
+        pytest.param("auto", 1, "local", id="use auto"),
+        pytest.param("mpirun", 1, "local", id="use mpirun"),
+        pytest.param("srun", 1, "slurm", id="use srun"),
+        pytest.param("srun", 2, "slurm", id="use srun w/2 db"),
```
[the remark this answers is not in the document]
ankona: Swapped to 3

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+        run_settings=rs_e,
+        perm_strategy="all_perm",
+        params={
+            "START": ["spam", "foo"],
+            "MID": ["eggs", "bar"],
+            "END": ["ham", "baz"],
+        },
+    )
+
+    yo_path = fileutils.get_test_conf_path("printing_model.py")
```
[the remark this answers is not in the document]
ankona: renamed to `test_script_path`

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+    rs_e.set_tasks(1)
+    ens = exp.create_ensemble(
+        "my-ens",
+        run_settings=rs_e,
+        perm_strategy="all_perm",
+        params={
+            "START": ["spam", "foo"],
+            "MID": ["eggs", "bar"],
+            "END": ["ham", "baz"],
+        },
```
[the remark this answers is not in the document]
ankona: reduced ensemble member count to 2

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
...
+
+        exp.start(db, block=False)
+        assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+        try:
+            exp.start(ens, block=True, summary=True)
+            assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+        finally:
+            exp.stop(db)
+            assert tm_pid == tm_pid == exp._control._telemetry_monitor.pid
+            time.sleep(3)  # time for telmon to write db stop event
```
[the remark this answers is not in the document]
ankona: The reduction in size of ensemble will turn 8 members into 2, which should go a long way to reducing test duration w/only 3 nodes & 2 dbs (leaving only 1 being able to run non-databases). I'd like to push for another node in the test allocation instead of reducing test complexity further... especially when we're considering _adding_ db nodes in comment above

smartsim/_core/control/controller.py
```diff
@@ -739,13 +812,64 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
...
 class _AnonymousBatchJob(EntityList[Model]):
-    def __init__(
-        self, name: str, path: str, batch_settings: BatchSettings, **kwargs: t.Any
-    ) -> None:
-        super().__init__(name, path)
-        self.batch_settings = batch_settings
+    def __init__(self, model: Model) -> None:
+        super().__init__(model.name, model.path)
+        self.entities = [model]
+        self.batch_settings = model.batch_settings
```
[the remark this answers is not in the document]
ankona: added a validation method and called it. worried about how this might break existing tests...

smartsim/_core/config/config.py
MattToast, line 213: If a user were to explicitly
```sh
export SMARTSIM_FLAG_TELEMETRY=0
```
then ``os.environ.get("SMARTSIM_FLAG_TELEMETRY", 1)`` would eval `"0"` (as a str type) and `bool("0")` would eval ``True``, when the user almost certainly intended for the TM to be off and for this to eval ``False``. We should make an explicit cast to `int` before casting to `bool`

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1142 @@
+# BSD 2-Clause License
```
MattToast, line 1: Several tests in this file require the TM to be active in order for them to not pass. Therefore, if a tester were to explicitly `export SMARTSIM_FLAG_TELEMETRY=0` they will get spurious errors.

By my estimate these tests are:
- `test_telemetry_single_model`
- `test_telemetry_single_model_nonblocking`
- `test_telemetry_serial_models`
- `test_telemetry_serial_models_nonblocking`
- `test_telemetry_db_only_with_generate`
- `test_telemetry_db_only_without_generate`
- `test_telemetry_db_and_model`
- `test_telemetry_ensemble`
- `test_telemetry_colo`
- `test_telemetry_autoshutdown` (all perms)
- `test_multistart_experiment` (all perms)

Maybe add a:
```py
@pytest.fixture(autouse=True)
def turn_on_tm(monkeypatch):
    import smartsim._core.config.config
    monkeypatch.setattr(
        smartsim._core.config.config.Config,
        "telemetry_enabled",
        property(lambda self: True))
    yield
```
to the top of this module?

tests/test_telemetry_monitor.py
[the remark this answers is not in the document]
ankona: swapped to 1

tests/test_telemetry_monitor.py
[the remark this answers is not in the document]
MattToast: was decided to drop the clustered db perm from this test as it was not the focus of what is being tested

smartsim/_core/_cli/cli.py
```diff
@@ -41,46 +41,68 @@
...
-        self.args = self.parser.parse_args(app_args)
+        app_args = cli_args[1:]  # exclude the path to executable
+        subcommand = cli_args[1]  # first positional arg is the subcommand
 
-        if not (menu_item := self.menu.get(app_args[0], None)):
+        menu_item = self.menu.get(subcommand, None)
+        if not menu_item:
             self.parser.print_help()
-            return 0
+            return 2
```
ashao, line 77: Same here: `os.EX_USAGE`?

smartsim/_core/_cli/dbcli.py
```diff
@@ -25,11 +25,14 @@
...
 from smartsim._core._cli.utils import get_db_path
 
 
-def execute(_args: argparse.Namespace) -> int:
+def execute(
+    _args: argparse.Namespace, _unparsed_args: t.Optional[t.List[str]] = None, /
+) -> int:
     if db_path := get_db_path():
         print(db_path)
         return 0
```
ashao, line 38: `os.EX_OK`

smartsim/_core/_cli/plugin.py
```diff
@@ -0,0 +1,53 @@
...
+            msg = f"{plugin_name} terminated by user"
+            raise SmartSimCLIActionCancelled(msg) from ex
+
+    return process_execute
+
+
+def dashboard() -> MenuItemConfig:
+    return MenuItemConfig(
+        "dashboard",
+        "Start the SmartSim dashboard",
```
ashao, line 47: Make this more descriptive, maybe to the point of linking to the repo?

smartsim/_core/control/controller.py
```diff
@@ -104,12 +119,19 @@ def start(
...
         # start the job manager thread if not already started
         if not self._jobs.actively_monitoring:
             self._jobs.start()
 
+        serialize.save_launch_manifest(
+            launched.map(_look_up_launched_data(self._launcher))
+        )
+
+        # launch a telemetry monitor to track job progress
+        self.start_telemetry_monitor(exp_path)
```
ashao, line 133: `start_telemetry_monitor` checks whether telemetry is enabled or not. Move this check up here, otherwise a plain reading makes it seem that telemetry is enabled unconditionally

smartsim/_core/control/controller.py
```diff
@@ -739,13 +812,71 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
                             if db_script not in ensemble.db_scripts:
                                 set_script(db_script, client)
 
+    def start_telemetry_monitor(self, exp_dir: str) -> None:
+        logger.debug("Starting telemetry monitor process")
```
ashao, line 816: See above comment, about moving the check on `CONFIG.telemetry_enabled` one level up the stack. At the least, this debug line should be moved down, otherwise an erroneous message will be admitted to the debug log

smartsim/_core/control/job.py
```diff
@@ -27,10 +27,43 @@
...
 
 
+@dataclass(frozen=True)
+class _JobKey():
+    step_id: str
+    task_id: str
+
+
+class JobEntity:
+    """Minimum API required for a job processed in the JobManager with support for
```
ashao, line 42: Remove "Minimum"

smartsim/_core/control/jobmanager.py
```diff
@@ -145,13 +148,8 @@ def __getitem__(self, entity_name: str) -> Job:
...
         """
         with self._lock:
-            if entity_name in self.db_jobs:
-                return self.db_jobs[entity_name]
-            if entity_name in self.jobs:
-                return self.jobs[entity_name]
-            if entity_name in self.completed:
-                return self.completed[entity_name]
-            raise KeyError
+            entities = ChainMap(self.db_jobs, self.jobs, self.completed)
```
ashao, line 151: Nice cleanup!

smartsim/_core/control/manifest.py
```diff
@@ -214,3 +223,130 @@ def has_db_scripts(
...
+
+class _LaunchedManifestMetadata(t.NamedTuple):
+    run_id: str
+    exp_name: str
+    exp_path: str
+    launcher_name: str
+
+    @property
+    def exp_telemetry_subdirectory(self) -> pathlib.Path:
+        return _fmt_exp_telemetry_path(self.exp_path)
```
ashao, line 237: Slight preference for spelling out `fmt`

smartsim/_core/control/manifest.py
```diff
@@ -214,3 +223,130 @@ def has_db_scripts(
...
+    def run_telemetry_subdirectory(self) -> pathlib.Path:
+        return _fmt_run_telemetry_path(self.exp_path, self.exp_name, self.run_id)
+
+    @property
+    def manifest_file_path(self) -> pathlib.Path:
+        return self.exp_telemetry_subdirectory / _serialize.MANIFEST_FILENAME
+
+
+@dataclass(frozen=True)
+class LaunchedManifest(t.Generic[_T]):
```
ashao, line 249: Beautiful, what a great imrpvoement

smartsim/_core/entrypoints/indirect.py
```diff
@@ -0,0 +1,237 @@
...
+SIGNALS = [signal.SIGINT, signal.SIGTERM, signal.SIGQUIT, signal.SIGABRT]
+
+
+def main(
+    cmd: str,
+    etype: str,
+    cwd: str,
+    status_dir: str,
+) -> int:
+    """Execute the step command and emit tracking events"""
```
ashao, line 56: Can you expand the description of this a bit more to describe that this launches the original command, but then also spins up an event tracker

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+
+    return the_runs
+
+
+def load_manifest(file_path: str) -> t.Optional[RuntimeManifest]:
+    """Load a persisted manifest and return the content"""
+    manifest_dict: t.Optional[t.Dict[str, t.Any]] = None
+    try_count = 1
+
+    while manifest_dict is None and try_count < 6:
```
ashao, line 203: Let's make `try_count` a module variable at least

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+
+
+def faux_return_code(step_info: StepInfo) -> t.Optional[int]:
+    """Create a faux return code for a task run by the WLM. Must not be
+    called with non-terminal statuses or results may be confusing"""
+    if step_info.status not in TERMINAL_STATUSES:
+        return None
+
+    if step_info.status == STATUS_COMPLETED:
+        return 0
```
ashao, line 293: I think we should also be able to change these to the `os.EX_` numbers?

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+            entity.type,
+            "stop",
+            write_path,
+            self._logger,
+            detail=detail,
+            return_code=faux_return_code(step_info),
+        )
+
+    def on_timestep(self, timestamp: int) -> None:
+        """Called at polling frequency .to request status updates on
```
ashao, line 477: You spilled your `.` into the docstring

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+        if manifest_path.exists():
+            # a manifest may not exist depending on startup timing
+            action_handler.process_manifest(str(manifest_path))
+
+        observer.schedule(log_handler, experiment_dir, recursive=True)  # type:ignore
+        observer.schedule(action_handler, experiment_dir, recursive=True)  # type:ignore
+        observer.start()  # type: ignore
+
+        event_loop(observer, action_handler, frequency, logger, telemetry_cooldown)
+        return 0
```
ashao, line 602: Here again. `os.EX_`

smartsim/_core/launcher/step/slurmStep.py
ashao, line 1: Why doesn't slurm need the proxyable_run_cmd decorator?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def _dictify_run_settings(run_settings: RunSettings) -> t.Dict[str, t.Any]:
+    return {
+        "exe": run_settings.exe,
+        # TODO: We should try to move this back
+        # "exe_args": run_settings.exe_args,
+        "run_command": run_settings.run_command,
+        "run_args": run_settings.run_args,
+        # TODO: We currently do not have a way to represent MPMD commands!
```
ashao, line 188: This doesn't seem like it should be that hard to do? MPMD should just be lists of the extra slurm/pbs commands, e.g. `-N 1 -n 1 -c 32 : -N 1 -n32 -c 1`

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+            "runs": [new_run],
+        }
+    else:
+        manifest_dict["runs"].append(new_run)
+    finally:
+        with open(manifest.metadata.manifest_file_path, "w", encoding="utf-8") as file:
+            json.dump(manifest_dict, file, indent=2)
+
+
+def _dictify_model(
```
ashao, line 97: Wouldn't these dictify methods be more appropriate to add to `__dict__` on each entity's class definition?

ashao (changes requested):
In the interest of getting feedback quickly, I'll do two reviews. One which goes through the implementation (here). I'll post any comments about the tests later on.

smartsim/_core/_cli/cli.py
```diff
@@ -41,46 +41,68 @@
...
-            if item.configurator:
-                item.configurator(parser)
+        self.register_menu_items(menu)
+        self.register_menu_items([plugin() for plugin in plugins])
 
     def execute(self, cli_args: t.List[str]) -> int:
         if len(cli_args) < 2:
             self.parser.print_help()
-            return 0
+            return 2
```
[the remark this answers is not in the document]
MattToast: `2` was based on me half-remembering that was the "default" exit code from `argparse.ArguementParser.parse_args` when the arguments were malformed. I see no reason why we should be beholden to this and would be more than wiling to replace with `os.EX_USAGE`!

smartsim/_core/_cli/cli.py
MattToast: swapped!

smartsim/_core/entrypoints/telemetrymonitor.py
MattToast: Good call!

smartsim/_core/launcher/step/slurmStep.py
MattToast: Slurm doesn't need to be proxied because the because `SlurmStep`s are managed and can be tracked through the WLM, so we don't need to spawn an a additional process to check to see when command has completed as the JM already has this capability.

The rule of thumb at this point in time is that managed steps do not need to be run through the proxy as the WLM can keep track of their lifetime, while unmanged steps do need to be run through the proxy. This is part of the reason that I believe `manged` should be a class attr on a `Step` or we should have two distinct types of `ManagedStep` and `UnmanagedStep` so that we could remove the need for users to manually apply the `proxyable_run_cmd` decorator, but that change was deemed "out of scope" for this PR.

smartsim/_core/utils/serialize.py
MattToast: I would agree if the dashboard were expecting a `str`. Unfortunately right now the dashboard expects a `dict[str, str]` and we cannot "re-add" keys for the dashboard to display (for e.g. in your example it would be non-trivial to display the two different values of `-n`).

We could turn the `"run_args"` key into a `list[dict[str, str]]`, but I would have to check with @AlyssaCote to see how much effort it would take for the dashboard to render this info on the frontend!

smartsim/_core/control/controller.py
MattToast: Good call! I think original idea was to limit the number of `if telemeter monitor enabled` touch points, but if it is only used here, it seems reasonable to move the check here (at the very least I agree it becomes more apparent what the control flow is).

smartsim/_core/control/controller.py
MattToast: moved the `if CONFIG.telemetry_enabled` check

smartsim/_core/entrypoints/telemetrymonitor.py
MattToast: `try_count` or move the `6` to a module level `MAX_TRY_COUNT` variable? I don't really want to move `try_count` outside the scope of this fn if this fn is going to mutate the variable.

smartsim/_core/utils/helpers.py
```diff
@@ -64,21 +66,22 @@ def unpack_colo_db_identifier(db_id: str) -> str:
...
 def check_dev_log_level() -> bool:
-    try:
-        lvl = os.environ["SMARTSIM_LOG_LEVEL"]
-        if lvl == "developer":
-            return True
-        return False
-    except KeyError:
-        return False
+    lvl = os.environ.get("SMARTSIM_LOG_LEVEL", "")
+    if lvl == "developer":
```
al-rigazzi, line 82: Aren't lines 82-83 equivalent to `return lvl=="developer"`?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+]
+TELMON_SUBDIR: t.Final[str] = ".smartsim/telemetry"
+MANIFEST_FILENAME: t.Final[str] = "manifest.json"
+
+
+def save_launch_manifest(manifest: _Manifest[TStepLaunchMetaData]) -> None:
+    if not CONFIG.telemetry_enabled:
+        return
+
+    manifest.metadata.run_telemetry_subdirectory.mkdir(parents=True, exist_ok=True)
```
al-rigazzi, line 57: Codecov seems convinced that this file could use some more testing. Any chance we make it happier?

smartsim/_core/utils/serialize.py
al-rigazzi, line 188: Do we have a ticket for the two TODOs listed in this function?

smartsim/_core/utils/serialize.py
```diff
@@ -0,0 +1,237 @@
...
+        if ens.batch_settings else {},
+        "models": [
+            _dictify_model(model, *launching_metadata)
+            for model, launching_metadata in members
+        ],
+    }
+
+
+def _dictify_run_settings(run_settings: RunSettings) -> t.Dict[str, t.Any]:
+    return {
```
al-rigazzi, line 182: Can this function be directly tested?

tests/backends/test_dbmodel.py
```diff
@@ -398,7 +398,7 @@ def test_colocated_db_model_tf(fileutils, wlmutils, mlutils):
     test_script = fileutils.get_test_conf_path("run_tf_dbmodel_smartredis.py")
 
     # Create SmartSim Experience
-    exp = Experiment(exp_name, launcher=test_launcher)
+    exp = Experiment(exp_name, launcher=test_launcher, exp_path=test_dir)
```
al-rigazzi, line 401: I will not address hypothetical issue in this file or in any file which is heavily touched by the Slurm test PR, as I think that will fix most of them

smartsim/_core/utils/serialize.py
MattToast: I would actually argue they should not be instance methods on the `Model`/`Ensemble`/etc. The reason for this is I didn't want to clutter up the entities with a method that is dedicated specifically for serialization. For instance, I could see a another feature in SmartSim eventually needing to cast entities to dictionaries that are either subsets or superset of these dictionaries and I didn't want to start the pattern of making methods `Model.to_dict_for_serialization()`, `Model.to_dict_for_other_feature()`, etc.

MattToast: In other words, my goal was to try and encapsulate all logic regarding serialization into dedicated a `serialization` module. If these fns are not used outside of this module I didn't see a reason to expose them more than necessary.

Let me know if you think this seems justified, but if not I would be happy to revisit this decision!

tests/test_configs/telemetry/colocatedmodel.json
```diff
@@ -0,0 +1,69 @@
...
+                "inter_op_parallelism": null,
+                "intra_op_parallelism": null
+              },
+              "extra_db_args": {}
+            },
+            "scripts": [],
+            "models": []
+          },
+          "telemetry_metadata": {
+            "status_dir": "/lus/cls01029/mcbridch/ss/tests/test_output/test_telemetry_monitor/test_load_manifest_colo_model/my-exp/.smartsim/telemetry/telemetry_ensemble/002816b/model/colocated_model",
```
al-rigazzi, line 56: Anonymize paths

tests/test_configs/telemetry/db_and_model.json
```diff
@@ -0,0 +1,86 @@
+{
+    "schema info": {
+        "schema_name": "entity manifest",
+        "version": "0.0.1"
+    },
+    "experiment": {
+        "name": "my-exp",
+        "path": "/lus/cls01029/mcbridch/ss/tests/test_output/test_telemetry_monitor/test_telemetry_db_and_model/my-exp",
```
al-rigazzi, line 8: Anonymize paths

tests/test_configs/telemetry/db_and_model_1run.json
```diff
@@ -0,0 +1,79 @@
...
+    "launcher": "Slurm"
+  },
+  "runs": [
+    {
+      "run_id": "4b5507a",
+      "timestamp": 1699038661491043211,
+      "model": [
+        {
+          "name": "perroquet",
+          "path": "/lus/cls01029/mcbridch/ss/tests/test_output/test_telemetry_monitor/test_telemetry_db_and_model/my-exp/perroquet",
```
al-rigazzi, line 18: Anonymize paths

tests/test_configs/telemetry/ensembles.json
```diff
@@ -0,0 +1,329 @@
...
+            "models": [
+              {
+                "name": "my-ens_0",
+                "path": "/home/chris/code/ss",
+                "exe_args": [
+                  "yo.py"
+                ],
+                "run_settings": {
+                  "exe": [
+                    "/home/chris/.pyenv/versions/3.9.16/envs/ss/bin/python"
```
al-rigazzi, line 44: Consider anonymizing path

tests/backends/test_dbmodel.py
ashao, line 401: Al and I found that this is also needed to get the tests to run on horizon. Out of curiosity, do you know what the behaviour is for dashboard if you omit `exp_path`?

tests/test_configs/telemetry/serialmodels.json
```diff
@@ -0,0 +1,186 @@
+{
+    "schema info": {
+        "schema_name": "entity manifest",
+        "version": "0.0.1"
+    },
+    "experiment": {
+        "name": "my-exp",
+        "path": "/lus/cls01029/mcbridch/ss/tests/test_output/test_telemetry_monitor/test_telemetry_serial_models/my-exp",
```
al-rigazzi, line 8: Anonymize paths

tests/test_configs/telemetry/telemetry.json
```diff
@@ -0,0 +1,946 @@
+{
+    "experiment": {
+      "name": "my-exp",
+      "path": "/lus/cls01029/drozt/playground/ss/dash-int/my-exp",
```
al-rigazzi, line 4: Anonymize paths

tests/test_indirect.py
```diff
@@ -0,0 +1,193 @@
...
+    if len(msg_tuple) < 2:
+        assert False, "error message indicates no missing arguments"
+
+    actual_missing = msg_tuple[1].strip()
+    for exp in expected:
+        assert f"{exp}/" not in actual_missing
+
+
+def test_cleanup(capsys, monkeypatch):
+    """Ensure cleanup attempts termination of correcct process"""
```
ashao, line 79: One too many `c`s in "correcct"

tests/test_telemetry_monitor.py
```diff
@@ -0,0 +1,1145 @@
...
+
+        # Create SmartSim Experiment
+        exp = Experiment(exp_name, launcher=test_launcher, exp_path=test_dir)
+
+        # create run settings
+        app_settings = exp.create_run_settings("python", test_script)
+        app_settings.set_nodes(1)
+        app_settings.set_tasks_per_node(1)
+
+        #  # Create the SmartSim Model
```
al-rigazzi, line 509: Commented commented line a few times in this file. Is this intentional?

ashao (changes requested):
Some very minor comments. Otherwise the tests look really good!

al-rigazzi (changes requested):
Great work so far! Just requesting a couple of minor fixes to docs and maybe some more tests for files for which coverage seems a bit low.

smartsim/_core/launcher/step/slurmStep.py
ashao: Yah ok, I see that and buy that justification.

smartsim/_core/utils/serialize.py
ashao: Ahhhh, I see. Ok, as a stop gap for now let's throw a warning here and put up a ticket.

tests/backends/test_dbmodel.py
MattToast: If you omit `exp_path` and leave have the `SMARTSIM_FLAG_TELEMETRY` flag enabled, it would lead to placement of `.smartsim` directories in the CWD. In order to make sure that these dir always ended up in the `test_output` dir it was decided to give every experiment an appropriate path.

smartsim/_core/control/manifest.py
```diff
@@ -214,3 +223,130 @@ def has_db_scripts(
...
+@dataclass(frozen=True)
+class LaunchedManifestBuilder(t.Generic[_T]):
+    """A mutable class used to build a ``LaunchedManifest`` while going through
+    the launching process.
+    """
+
+    exp_name: str
+    exp_path: str
+    launcher_name: str
+    run_id: str = field(default_factory=_helpers.create_short_id_str)
```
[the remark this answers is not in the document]
MattToast: The class still has some mutable attrs, namely the `_models`, `_ensembles`, and `_databases` lists which can all be appended to. The wording of the docstring could probably make this more clear! Will change to "A class comprised of mutable collections of SmartSim entities that is used to build a \`\`LaunchedManifest\`\`"!

smartsim/_core/launcher/step/step.py
```diff
@@ -107,3 +116,49 @@ def add_to_batch(self, step: Step) -> None:
...
+) -> t.Callable[[_StepT], t.List[str]]:
+    @functools.wraps(fn)
+    def _get_launch_cmd(self: _StepT) -> t.List[str]:
+        original_cmd_list = fn(self)
+
+        if not CONFIG.telemetry_enabled:
+            return original_cmd_list
+
+        if self.managed:
+            raise UnproxyableStepError(
```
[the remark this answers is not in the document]
MattToast: Hmmmm.... I thought I wrote a test to test explicitly this....let me look into why this wasn't included in the coverage

smartsim/_core/launcher/cobalt/cobaltLauncher.py
```diff
@@ -117,8 +117,7 @@ def run(self, step: Step) -> t.Optional[str]:
             # aprun doesn't direct output for us.
             out, err = step.get_output_files()
 
-            # LocalStep.run_command omits env, include it here
-            passed_env = step.env if isinstance(step, LocalStep) else None
+            passed_env = step.env
```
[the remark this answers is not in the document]
MattToast: Makes sense! My `grep`/`sed` skills just weren't good enough to rm the useless variable 😞.

Updated and rm-ed those useless vars by hand!

smartsim/_core/launcher/local/local.py
```diff
@@ -104,16 +105,19 @@ def run(self, step: Step) -> str:
...
         cmd = step.get_launch_cmd()
 
-        # LocalStep.run_command omits env, include it here
-        passed_env = step.env if isinstance(step, LocalStep) else None
+        # pylint: disable-next=consider-using-with
+        output = open(out, "w+", encoding="utf-8")
+        # pylint: disable-next=consider-using-with
+        error = open(err, "w+", encoding="utf-8")
+
+        passed_env = step.env
```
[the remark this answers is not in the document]
MattToast: `grep`/`sed` is hard

smartsim/_core/launcher/lsf/lsfLauncher.py
```diff
@@ -115,12 +115,11 @@ def run(self, step: Step) -> t.Optional[str]:
...
             step_id = self._get_lsf_step_id(step)
             logger.debug(f"Gleaned jsrun step id: {step_id} for {step.name}")
-        else:  # isinstance(step, MpirunStep) or isinstance(step, LocalStep)
+        else:
             # mpirun and local launch don't direct output for us
             out, err = step.get_output_files()
 
-            # LocalStep.run_command omits env, include it here
-            passed_env = step.env if isinstance(step, LocalStep) else None
+            passed_env = step.env
```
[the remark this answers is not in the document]
MattToast: `grep`/`sed` is hard

smartsim/_core/launcher/pbs/pbsLauncher.py
```diff
@@ -111,8 +111,7 @@ def run(self, step: Step) -> t.Optional[str]:
             # aprun/local doesn't direct output for us.
             out, err = step.get_output_files()
 
-            # LocalStep.run_command omits env, include it here
-            passed_env = step.env if isinstance(step, LocalStep) else None
+            passed_env = step.env
```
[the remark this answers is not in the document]
MattToast: `grep`/`sed` is hard

smartsim/_core/launcher/slurm/slurmLauncher.py
```diff
@@ -155,8 +155,7 @@ def run(self, step: Step) -> t.Optional[str]:
             # MPI/local steps don't direct output like slurm steps
             out, err = step.get_output_files()
 
-            # LocalStep.run_command omits env, include it here
-            passed_env = step.env if isinstance(step, LocalStep) else None
+            passed_env = step.env
```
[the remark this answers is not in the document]
MattToast: `grep`/`sed` is hard

smartsim/_core/launcher/local/local.py
```diff
@@ -60,10 +56,11 @@ def create_step(self, name: str, cwd: str, step_settings: SettingsBase) -> Step:
...
                 f"Local Launcher only supports entities with RunSettings, not {type(step_settings)}"
             )
-        step = LocalStep(name, cwd, step_settings)
-        return step
+        return LocalStep(name, cwd, step_settings)
 
-    def get_step_update(self, step_names: t.List[str]) -> t.List[t.Tuple[str, t.Optional[StepInfo]]]:
+    def get_step_update(
+        self, step_names: t.List[str]
+    ) -> t.List[t.Tuple[str, t.Optional[StepInfo]]]:
```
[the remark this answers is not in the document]
MattToast: Re-synced the typehint and docstring!

Looks the de-sync was coming because docstring was _technically_ "more correct" but we were expecting a `list` return type which is invariant and thus breaking liskov's substitution principle. Might be worth going through these methods and updating them with the co-variant equivalents (not part of this ticket)!

smartsim/_core/launcher/local/local.py
```diff
@@ -24,29 +24,25 @@
...
+import os
+import sys
 import typing as t
 
 from ..launcher import Launcher
 from ....log import get_logger
 from ....settings import RunSettings, SettingsBase
-from ..step import LocalStep
-from ..step import Step
+from ..step.localStep import LocalStep
```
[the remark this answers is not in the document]
MattToast: No reason! Just saw they were on different lines before and figured this was intended! Unified to `from ..step import LocalStep, Step`

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -0,0 +1,677 @@
...
+    :param observer: (optional) a preconfigured watchdog Observer to inject
+    :type observer: t.Optional[BaseObserver]
+    :param action_handler: The manifest event processor instance
+    :type action_handler: ManifestEventHandler
+    :param frequency: frequency (in seconds) of update loop
+    :type frequency: t.Union[int, float]
+    :param experiment_dir: the experiement directory to monitor for changes
+    :type experiment_dir: pathlib.Path
+    :param logger: a preconfigured Logger instance
+    :type logger: logging.Logger"""
```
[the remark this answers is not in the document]
MattToast: Added!

smartsim/_core/control/manifest.py
```diff
@@ -214,3 +223,130 @@ def has_db_scripts(
...
+    _ensembles: t.List[t.Tuple[Ensemble, t.Tuple[t.Tuple[Model, _T], ...]]] = field(
+        default_factory=list, init=False
+    )
+    _databases: t.List[
+        t.Tuple[Orchestrator, t.Tuple[t.Tuple[DBNode, _T], ...]]
+    ] = field(default_factory=list, init=False)
+
+    @property
+    def exp_telemetry_subdirectory(self) -> pathlib.Path:
+        return _fmt_exp_telemetry_path(self.exp_path)
```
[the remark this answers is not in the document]
MattToast: Added test!

smartsim/_core/control/controller.py
```diff
@@ -739,13 +812,71 @@ def _set_dbobjects(self, manifest: Manifest) -> None:
                             if db_script not in ensemble.db_scripts:
                                 set_script(db_script, client)
 
+    def start_telemetry_monitor(self, exp_dir: str) -> None:
```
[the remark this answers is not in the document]
MattToast: Added!

smartsim/_core/control/controller.py
```diff
@@ -462,31 +520,42 @@ def _launch_step(
...
+            step = self._create_job_step(entity, telemetry_dir)
+            substeps.append(step)
             batch_step.add_to_batch(step)
-        return batch_step
+        return batch_step, substeps
 
-    def _create_job_step(self, entity: SmartSimEntity) -> Step:
+    def _create_job_step(
+        self, entity: SmartSimEntity, telemetry_path: pathlib.Path
+    ) -> Step:
```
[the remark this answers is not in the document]
MattToast: Added and conformed name to `telemetry_dir` to match `_create_batch_job_step`

smartsim/_core/control/controller.py
```diff
@@ -348,33 +375,51 @@ def _launch(self, manifest: Manifest) -> None:
...
             self._launch_step(step, entity)
 
-    def _launch_orchestrator(self, orchestrator: Orchestrator) -> None:
+        return manifest_builder.finalize()
+
+    def _launch_orchestrator(
+        self,
+        orchestrator: Orchestrator,
+        manifest_builder: LaunchedManifestBuilder[t.Tuple[str, Step]],
+    ) -> None:
```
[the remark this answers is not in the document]
MattToast: Added!

smartsim/_core/control/controller.py
```diff
@@ -462,31 +520,42 @@ def _launch_step(
...
+        entity_list: t.Union[Orchestrator, Ensemble, _AnonymousBatchJob],
+        telemetry_dir: pathlib.Path,
+    ) -> t.Tuple[Step, t.List[Step]]:
         """Use launcher to create batch job step
 
         :param entity_list: EntityList to launch as batch
         :type entity_list: EntityList
-        :return: job step instance
-        :rtype: Step
+        :return: batch job step instance and a list of run steps to be
```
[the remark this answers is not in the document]
MattToast: Good catch!

smartsim/_core/config/config.py
```diff
@@ -204,6 +204,17 @@ def test_account(self) -> t.Optional[str]:  # pragma: no cover
...
+    def telemetry_frequency(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_FREQUENCY", 5))
+
+    @property
+    def telemetry_enabled(self) -> bool:
+        return int(os.environ.get("SMARTSIM_FLAG_TELEMETRY", "0")) > 0
+
+    @property
+    def telemetry_cooldown(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_COOLDOWN", 90))
```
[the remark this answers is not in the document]
MattToast: Test added!

smartsim/_core/config/config.py
```diff
@@ -204,6 +204,17 @@ def test_account(self) -> t.Optional[str]:  # pragma: no cover
         # no account by default
         return os.environ.get("SMARTSIM_TEST_ACCOUNT", None)
 
+    @property
+    def telemetry_frequency(self) -> int:
+        return int(os.environ.get("SMARTSIM_TELEMETRY_FREQUENCY", 5))
```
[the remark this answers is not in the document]
MattToast: Test added!

smartsim/_core/_cli/plugin.py
```diff
@@ -0,0 +1,53 @@
...
+    def process_execute(
+        _args: argparse.Namespace, unparsed_args: t.List[str], /
+    ) -> int:
+        not_found = f"{cmd} plugin not found. Please ensure it is installed"
+        try:
+            spec = importlib.util.find_spec(cmd)
+            if spec is None:
+                raise AttributeError()
+        except (ModuleNotFoundError, AttributeError):
+            print(not_found)
```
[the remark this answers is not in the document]
MattToast: Good call!

smartsim/_core/launcher/step/step.py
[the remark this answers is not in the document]
MattToast: Confirmed that I do have a `test_proxy_launch_cmd_decorator_errors_if_attempt_to_proxy_a_managed_step` test in the `test_telemetry_monitor.py` test module!

I think because the new test modules were added after the CI refactor that split our tests into groups, none of the runners are picking up the new test modules and thus they are not contributing to coverage. Need to look into this a bit further, but that is my running hypothesis!

smartsim/_core/utils/helpers.py
MattToast: It absolutely is! Rm'd the noise!

smartsim/_core/control/job.py
```diff
@@ -42,7 +75,7 @@ def __init__(
         self,
         job_name: str,
         job_id: t.Optional[str],
-        entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]],
+        entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity], JobEntity],
```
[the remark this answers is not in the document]
MattToast: Updated!

tests/test_telemetry_monitor.py
MattToast: I'm guessing not. Rm'd

smartsim/_core/utils/serialize.py
MattToast: We do for moving `exe_args` back to (at least in my opinion) is a more appropriate location. I still need to write up one for the MPMD gap

smartsim/_core/control/jobmanager.py
```diff
@@ -185,7 +183,8 @@ def add_job(
         job = Job(job_name, job_id, entity, launcher, is_task)
         if isinstance(entity, (DBNode, Orchestrator)):
             self.db_jobs[entity.name] = job
-
+        elif isinstance(entity, JobEntity) and entity.is_db:
+            self.db_jobs[entity.name] = job
```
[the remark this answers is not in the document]
MattToast: Looks like this clause is now covered after registering new test modules with the CI!

smartsim/_core/utils/serialize.py
MattToast: Seems to be covered now that new test modules are picked up by CI

smartsim/_core/utils/serialize.py
MattToast: Seems to be covered now that CI is picking up the new test modules. Also added a test for a warning requested by @ashao

smartsim/_core/control/job.py
```diff
@@ -27,10 +27,43 @@
...
+    telemetry monitoring"""
+
+    def __init__(self) -> None:
+        self.name: str = ""
+        self.path: str = ""
+        self.step_id: str = ""
+        self.task_id: str = ""
+        self.type: str = ""
+        self.timestamp: int = 0
+        self.status_dir: str = ""
```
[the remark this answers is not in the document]
MattToast: Looks like it is now covered now that the new test modules are being picked up by the CI!

smartsim/_core/_cli/__main__.py
```diff
@@ -27,11 +27,28 @@
...
 
 def main() -> int:
     smart_cli = default_cli()
-    return smart_cli.execute(sys.argv)
+
+    try:
+        return smart_cli.execute(sys.argv)
+    except SmartSimCLIActionCancelled as ssi:
+        logger.debug(ssi, exc_info=True)
+        logger.info(ssi)
```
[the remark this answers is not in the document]
MattToast: Nope, that is absolutely the case! will give the debug logger call a generic "here is the traceback" message

smartsim/_core/_cli/__main__.py
```diff
@@ -27,11 +27,28 @@
...
+    except KeyboardInterrupt:
+        msg = "SmartSim was terminated by user"
+        logger.debug(msg, exc_info=True)
```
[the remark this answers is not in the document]
MattToast: ditto here

smartsim/_core/control/controller.py
```diff
@@ -312,7 +335,9 @@ def init_launcher(self, launcher: str) -> None:
         else:
             raise TypeError("Must provide a 'launcher' argument")
 
-    def _launch(self, manifest: Manifest) -> None:
+    def _launch(
+        self, exp_name: str, exp_path: str, manifest: Manifest
+    ) -> LaunchedManifest[t.Tuple[str, Step]]:
```
al-rigazzi, line 340: Missed this on first review, can you please update the docstring to reflect the new function signature?

smartsim/_core/control/job.py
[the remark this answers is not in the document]
al-rigazzi: Looks good!

smartsim/_core/control/manifest.py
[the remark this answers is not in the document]
al-rigazzi: Thanks, possibly not needed after you explained, but really appreciate making this clearer for devs

smartsim/_core/launcher/step/step.py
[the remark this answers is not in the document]
al-rigazzi: Great

smartsim/experiment.py
```diff
@@ -878,3 +881,35 @@ def append_to_db_identifier_list(self, db_identifier: str) -> None:
...
+
+    def enable_telemetry(self) -> None:
+        """Experiments will start producing telemetry for all entities run
+        through ``Experiment.start``
+
+        .. warning::
+
+            This method is currently implemented so that ALL ``Experiment``
+            instances will begin producing telemetry data. In the future it
+            planned to have this method work on a "per instance" basis!
```
al-rigazzi, line 893: Minor: "it planned" -> "it is planned"

al-rigazzi (changes requested):
Two super-minor fixes to docs (I missed on first review) and then it looks good to me!

smartsim/experiment.py
ankona: fixed!

smartsim/_core/control/controller.py
ankona: Added `exp_path` & `exp_name` params in docstring

smartsim/experiment.py
MattToast: Can you all tell that I have set up a spell checker, but not a grammar checker in my IDE 🙃

al-rigazzi (approved):
LGTM! Thanks for addressing the very last remarks!

smartsim/_core/utils/serialize.py
ashao: Punted this into a different ticket

ashao (approved):

PR #435, opened 2023-12-08
Remove conflicting setuptools instructions

MattToast:
This PR fixes some conflicting directives in the SmartSim packaging instructions:

- `setup.py` manually listed an incomplete list of packages to include in SmartSim while the `setup.cfg` was using `find_packages`. This PR default to `setup.cfg` with a slightly refined `include` directive.
- `setup.py` manually listed `package_data` directives while the `setup.cfg` set `include_package_data=True`. This PR defaults to `setup.cfg`.
- In order to exclude the `__pycache__` created from SmartSim modules imported during the `setup.py` script, `__pycache__` and related files were explicitly ignored in the `MANIFEST.in`.
- In order to ensure the the `smartsim._core.launcher.local` package was found by `find_packages` without reverting to `find_namespace_packages` an `__init__.py` module was added to the directory.


setup.py
```diff
@@ -200,10 +200,6 @@ def has_ext_modules(_placeholder):
 setup(
     version=smartsim_version,
     install_requires=deps,
-    packages=["smartsim"],
-    package_data={"smartsim": [
-        "_core/bin/*",
```
ankona, line 205: Is there any value in the explicitly defined package data location? `graft` is obviously simpler, but doesn't tell anybody that we're doing something on purpose (which, I think, is to include db binaries).

ankona (approved):

setup.py
MattToast: Good point! Originally removed the `package_data` section for two reasons:
1) It is technically redundant due to the `include_package_data=True`/`MANIFEST.in`
2) It is [slightly mis-formatted according to `setuptools`](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files)

But I see what you mean about it providing some insight into _what_ package data we are expecting to be included. In fact, I think the original error becomes more easy to parse if we properly format this section (e.g. `smartsim._core.bin` is not a package!).

Will add this section back (but under the `setup.cfg`, just to try and reduce the number of files dealing with `package_data`)!

al-rigazzi (approved):

PR #439, opened 2023-12-14
Add concurrency group to test workflow

al-rigazzi:
This PR adds concurrency groups to GitHub's CI/CD workflows, preventing multiple workflows from the same PR to be launched concurrently.


MattToast (approved):
LGTM!! Thanks for helping to save minutes/free up runners!

PR #446, opened 2023-12-19
Update actions

al-rigazzi:
This PR updates actions which are part of GitHub workflows. The update is needed because of several warning which are issued at every CI/CD run on GH.


ashao (approved):
Thanks for tracking these down. Always nice to have quieter CI/CD.

PR #447, opened 2024-01-03
Experiment Preview

juliaputko:


smartsim/experiment.py
```diff
@@ -797,6 +796,56 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+
+    def _preview(self) -> str:
+        """String formatting"""
+
+        preview = Viewexp(self).render()
+        logger.info(preview)
+
+        return preview
+
+    def get_launcher(self) -> str:
```
ankona, line 845: If we need this, would a normal read property do?

```py
@property
def launcher(self) -> str:
    return self._launcher
```

smartsim/experiment.py
```diff
@@ -797,6 +796,56 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        debug: Display user defined field and entities and auto generated
+        fields.
+        developer: Display user defined field and entities, auto generated
+        fields, and run commands.
+        Defaults to info.
+        type verbosity_level: str
+        """
+
+        if output_format:
+            raise NotImplementedError
```
ankona, line 829: Are these just temporary until you continue this ticket?

smartsim/_core/control/view.py
```diff
@@ -0,0 +1,43 @@
...
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import typing as t
+import jinja2
+
+
+class Viewexp:
```
ankona, line 30: I think you'll probably want to `PascalCase` this class name.

Consider using the `._prop_name` underscore prefix for private members.

Consider tightening the type instead of `t.Any`

smartsim/experiment.py
```diff
@@ -797,6 +796,56 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        if output_filename:
+            raise NotImplementedError
+        if verbosity_level:
+            raise NotImplementedError
+
+        return self._preview()
+
+    def _preview(self) -> str:
```
ankona, line 837: Consider renaming `ViewExp` to `ExpPreviewRenderer` to clearly indicate its purpose

Consider getting rid of the `ViewExp` class altogether and just using a function since no members are really used. If not, consider static methods so you can get rid of `Experiment._preview` and use:

```py
return ExpView.render(self)
```

tests/test_preview.py
```diff
@@ -0,0 +1,82 @@
...
+from smartsim import Experiment
+
+
+def test_experiment_preview(test_dir, wlmutils):
+    """Test correct preview output items for Experiment preview"""
+    test_launcher = wlmutils.get_test_launcher()
+    exp_name = "test_prefix"
+    exp = Experiment(exp_name, exp_path=test_dir, launcher=test_launcher)
+    # Call method for string formatting for testing
+    output = exp._preview().strip()
```
ankona, line 38: The `exp.preview` method doesn't really do anything. This low-level test case might be better if it targeted `ExpView.render` directly. Alternatively, to write fewer test cases, consider testing the public method `exp.preview` so the entire call-chain is covered?

Consider doing the `.strip()` inside the preview function (because all the tests have to repeat it), or preferably force the template to produce the desired output (check out whitespace management in jinja docs)

smartsim/_core/control/view.py
al-rigazzi: Agree with @ankona, also I think we usually stick to full names, so I'd prefer `ExperimentView` as class name.

setup.py
```diff
@@ -168,6 +173,7 @@ def has_ext_modules(_placeholder):
     "filelock>=3.4.2",
     "protobuf~=3.20",
     "watchdog>=3.0.0",
+    "jinja2",
```
al-rigazzi, line 176: Please pin a (minimum) version.

smartsim/experiment.py
```diff
@@ -797,6 +796,56 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+    def preview(
+        self,
+        output_format: t.Optional[str] = None,
+        output_filename: t.Optional[str] = None,
+        verbosity_level: t.Optional[str] = None,
+    ) -> str:
+        """Preview entity information prior to launch. This method
+        aggregates multiple pieces of information to give users insight
+        into what and how entities will be launched.  Any instance of
+        ``Model``, ``Ensemble`` or ``Orchestrator`` created by the
```
al-rigazzi, line 808: May I ask you to use the Oxford comma for consistency with the rest of the documentation?

smartsim/experiment.py
```diff
@@ -797,6 +796,56 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        Experiment can be passed as an argument to the preview method.
+        param output_format: Set output destination. The possible accepted
```
al-rigazzi, line 810: Missing `:` before `param` and `type`, the syntax should be `:param param_name:` and `:type param_name:` to avoid Sphinx errors or warnings.

al-rigazzi (changes requested):
Good starting point, I agree with @ankona's suggestion to use static methods. @MattToast may want to chime in as well.

smartsim/_core/control/view.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, exp_entity: t.Any) -> None:
+        self.exp_entity = exp_entity
+
+    def render(self) -> str:
+        """
+        Render the template from the supplied entity
+        """
+        loader = jinja2.PackageLoader("templates")
```
ankona, line 38: templates should probably be in `smartsim.config`?

preview_test.html
```diff
@@ -0,0 +1,5 @@
+
+=== Experiment Overview ===
+Experiment: test_prefix
+Experiment Path: /lus/cls01029/putko/test/smartsim/tests/test_output/test_preview/test_preview_output_format_html
```
ankona, line 4: I don't think you meant to include this test output in the commit

smartsim/experiment.py
```diff
@@ -797,6 +796,89 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+            raise ValueError("The only valid currently available is html")
+
+    @staticmethod
+    def _check_verbosity_level(
+        verbosity_level: t.Literal["info", "debug", "developer"]
+    ) -> None:
+        """
+        Check verbosity_level
+        """
+        if verbosity_level == "debug":
```
ankona, line 862: Are we duplicating the verbosity levels of the logging system but using them inside of the templates? If so, can we use something that already exists in `smartsim.log` instead of having something in preview that also couples tightly to log levels?

smartsim/log.py
```diff
@@ -129,3 +129,23 @@ def log_to_file(filename: str, log_level: str = "debug") -> None:
...
+    logger: logging.Logger,
+) -> logging.Handler:
+    """Installs a filestream handler to the root logger,
+    allowing logging calls to be sent to filename for preview.
+    :param filename: the name of the desired preview file.
+    :type filename: str
+    """
+    logger.propagate = False
+    fmt = None
+    formatter = logging.Formatter(fmt)
```
ankona, line 145: Do you really want a None here? would `"%(message)s"` do the job? 

I don't even know what happens if you have a null formatter!

smartsim/log.py
```diff
@@ -129,3 +129,23 @@ def log_to_file(filename: str, log_level: str = "debug") -> None:
         filename, "w+", encoding="utf-8"
     )
     coloredlogs.install(stream=stream, logger=logger, level=log_level)
+
+
+def log_to_file_preview(
+    filename: str,
+    logger: logging.Logger,
+) -> logging.Handler:
+    """Installs a filestream handler to the root logger,
```
ankona, line 138: I think this docstring is a little misleading. Since we don't propagate, it's not really accurate to say "installs...to the root logger"

smartsim/experiment.py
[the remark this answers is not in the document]
ankona: Alternatively, i don't think a bad preview verbosity should crash anything. Consider logging a warning and setting to info instead?

smartsim/experiment.py
```diff
@@ -797,6 +796,89 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        fields, and run commands.
+        Defaults to info.
+        :type verbosity_level: str
+        """
+
+        rendered_preview = previewrenderer.render(self, verbosity_level)
+
+        self._check_verbosity_level(verbosity_level)
+
+        if output_format:
```
ankona, line 832: does this need to be if/else? can self._check_output_format accept None and just do 
```py
if not output_format_arg:
  return
```

smartsim/experiment.py
```diff
@@ -797,6 +796,89 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+            if not output_filename:
+                raise ValueError(
+                    "An output filename is required when an output format is set."
+                )
+            previewrenderer.preview_to_file(rendered_preview, output_filename)
+
+        logger.info(rendered_preview)
+
+    @staticmethod
+    def _check_output_format(output_format: t.Literal["html"]) -> None:
```
ankona, line 851: Let's define these literal types at the top of the file as constants so we can reuse them.

Consider if we modify output_format options to include csv.. We can avoid changing everywhere that has this literal re-defined.

search for `_TDeviceStr = t.Literal["cpu", "gpu"]` for an example

smartsim/experiment.py
```diff
@@ -797,6 +796,89 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        if not output_format.startswith("html"):
+            raise ValueError("The only valid currently available is html")
+
+    @staticmethod
+    def _check_verbosity_level(
+        verbosity_level: t.Literal["info", "debug", "developer"]
```
ankona, line 857: ditto on re-defining literal

PR #448, opened 2024-01-05
Remove Cobalt support

al-rigazzi:
With the retirement of Theta, we are not aware of any other system using Cobalt as a workload manager. Therefore, this PR removes support for Cobalt.


MattToast (approved):
Everything seems to LGTM!!

ashao (approved):
Looks good! The only suggestion, which may have been discussed while I was away, is to throw a more informative error if someone specifically tried to request the cobalt launcher. I wouldn't even give it it's own dummy class, just throw it on the instantiation of `Experiment`, i.e.

```
if launcher == "cobalt":
  raise SSUnsupportedError("Support for the Cobalt launcher is no longer supported")
```

al-rigazzi:
> Looks good! The only suggestion, which may have been discussed while I was away, is to throw a more informative error if someone specifically tried to request the cobalt launcher. I wouldn't even give it it's own dummy class, just throw it on the instantiation of `Experiment`, i.e.
> 
> ```
> if launcher == "cobalt":
>   raise SSUnsupportedError("Support for the Cobalt launcher is no longer supported")
> ```

Slightly reworded and added to `Experiment`. Added a test.

PR #449, opened 2024-01-05
Fix index when installing torch through smart build

ashao:
Torch changed something in their indexing when trying to install from their provided wheels. This updates the `pip install` command within `smart build` to ensure that the appropriate packages can be found.


ankona (approved):
LGTM!

PR #453, opened 2024-01-10
Override sphinx-tabs background color

mellis13:
The sphinx-tabs documentation extension uses a white background for the tabs component.  This causes readability issues with the theme that we have chosen.  This PR overrides the CSS for those components to inherit the overall theme color.


al-rigazzi (approved):
LGTM! Thanks for looking into this!

PR #461, opened 2024-01-20
Python 3.11 Support

MattToast:
Adds Python 3.11 to the SmartSim support matrix.

---

TODO before merge:
- [x] Run the full test suite against a super (I was running into "invalid instruction" problems earlier when attempting to start redis through SmartSim. Back of the envelope math says this should be unrelated to Python 3.11 support, there is a good chance I misconfigured my environment, but it's still good practice to check!)
- [x] Run the backends test against a machine with GPUs (again, back of the envelope says they _should_ be fine, but doesn't hurt to confirm!)
- [x] Update the change log!

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -57,7 +57,7 @@ jobs:
         os: [macos-12, ubuntu-20.04] # Operating systems
         compiler: [8] # GNU compiler version
         rai: [1.2.7] # Redis AI versions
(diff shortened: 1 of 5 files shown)


ashao (changes requested):
Tick those TODOs and I think you're good to go!

MattToast:
TODOs have been addressed and this should be ready for re-review/merge! 🚀

ashao (approved):
Merge it!

PR #464, opened 2024-01-25
Add isort/black check to github actions

amandarichardsonn:


.github/workflows/run_tests.yml
```diff
@@ -118,6 +118,16 @@ jobs:
...
 
+      # Run isort/black style check
+      - name: Run isort
+        run: isort --check-only --profile black ./smartsim ./tests
+
+      # Run isort/black style check
+      - name: Run black
+        run: |
+          black ./smartsim/version.py
+          black --check ./smartsim ./tests
```
MattToast, line 129: Super small request here, I'm not sure that I like the CI is making changes to the code that we generate, especially since those changes will not be made in the release CI.

Can we just have black ignore that file instead?
```suggestion
          black --exclude smartsim/version.py --check ./smartsim ./tests
```

MattToast (changes requested):
One super small change request, but otherwise LGTM!! Thanks for making sure that we keep our style black compliant!

.github/workflows/run_tests.yml
amandarichardsonn: that is a way better idea!!

MattToast (approved):
Perfect! LGTM pending tests!!

PR #465, opened 2024-01-27
Add support for Mac OSX on Apple Silicon

ashao:
With the new ml_lib_builder repository we can now ship a version of libtorch that is compiled for Mac OSX on Apple Silicon (arm64). This PR explicitly allows ARM64 as an allowed architecture and creates variants fo the PTArchive class for Linux and Mac operating systems.


MattToast (changes requested):
Looks great! I go a couple of weird corner cases for you to consider and some truly pedantic style comments for you to consider while you sort out the CI, but otherwise seems about ready to go to me!

smartsim/_core/_install/builder.py
```diff
@@ -454,7 +457,7 @@ def _get_deps_to_fetch_for(
         #       is used as script in the SmartSim `setup.py`.
         fetchable_deps: t.Sequence[t.Tuple[bool, _RAIBuildDependency]] = (
             (True, _DLPackRepository("v0.5_RAI")),
-            (self.fetch_torch, _PTArchive(os_, device, "2.0.1")),
+            (self.fetch_torch, choose_PT_variant(os_, arch, device, "2.0.1")),
```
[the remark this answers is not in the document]
ashao: Oh I see. Yeah good catch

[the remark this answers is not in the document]
ashao: Ok, not the prettiest thing I've ever seen, but I think it'll do for now until we refactor:
```
        # DLPack is always required
        fetchable_deps: t.List[_RAIBuildDependency] = [_DLPackRepository("v0.5_RAI")]
        if self.fetch_torch:
            fetchable_deps.append(choose_pt_variant(os_, arch, device, "2.0.1"))
        if self.fetch_tf:
            fetchable_deps.append(_TFArchive(os_, arch, device, "2.13.1"))
        if self.fetch_onnx:
            fetchable_deps.append(_ORTArchive(os_, device, "1.16.3"))

        return tuple(fetchable_deps)
```

smartsim/_core/_install/builder.py
```diff
@@ -797,6 +782,49 @@ def __place_for_rai__(self, target: t.Union[str, "os.PathLike[str]"]) -> Path:
         return target
 
 
+@t.final
+class _PTArchive_Linux(_PTArchive):
```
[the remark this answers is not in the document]
ashao: Fixed

smartsim/_core/_install/builder.py
```diff
@@ -797,6 +782,49 @@ def __place_for_rai__(self, target: t.Union[str, "os.PathLike[str]"]) -> Path:
...
+            raise BuildError("RedisAI does not currently support GPU on Mac OSX")
+        if self.architecture == Architecture.X64:
+            libtorch_archive= f"libtorch-macos-{self.version}.zip"
+            return f"https://download.pytorch.org/libtorch/{pt_build}/{libtorch_archive}"
+        elif self.architecture == Architecture.ARM64:
+            libtorch_archive = f"libtorch-macos-arm64-{self.version}.zip"
+            root_url = "https://github.com/CrayLabs/ml_lib_builder/releases/download/v0.1/"
+            out = f"{root_url}/{libtorch_archive}"
+            return out
+
```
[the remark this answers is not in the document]
ashao: Mypy caught this for me as a unnecessary elifs due to the multiple returns

doc/changelog.rst
```diff
@@ -41,13 +43,17 @@ Detailed Notes
   there is now an available ONNX wheel for use with Python 3.10, and wheels for
   all of SmartSim's machine learning backends with Python 3.11.
   (SmartSim-PR451_) (SmartSim-PR461_)
+- SmartSim can now be built and used on platforms using Apple Silicon
+  (ARM64). Currently, only the pyTorch backend is supported. Note that libtorch
```
MattToast, line 47: Correct me if I'm wrong but isn't the preferred capitalization
```suggestion
  (ARM64). Currently, only the PyTorch backend is supported. Note that libtorch
```
? At least I thought that was how they titled their releases.

smartsim/_core/_install/builder.py
```diff
@@ -385,10 +395,21 @@ def __init__(
...
+    def _check_backends_arm64(self) -> None:
+        if self._architecture == Architecture.ARM64:
+            unsupported = []
+            if self.build_tf:
+                unsupported.append("Tensorflow")
+            if self.build_onnx:
+                unsupported.append("ONNX")
+            if unsupported:
+                raise BuildError(
+                    f"The {'.'.join(unsupported)} backends are not "
```
MattToast, line 410:
```suggestion
                    f"The {', '.join(unsupported)} backend(s) are not "
```

smartsim/_core/_install/builder.py
```diff
@@ -385,10 +395,21 @@ def __init__(
...
+                    "supported on ARM64. Run with `smart build --no_tf`"
+                )
```
MattToast, line 412: I like the decision to try and move as much of the validation of dependencies into the constructor to error out early!

But I don't necessarily love the existence of this method. It's going to become very difficult to maintain if we were to expand out into different architectures w/o full dependency support and/or we want to become more prescriptive in our builds (e.g. different versions of libraries for different architectures, etc.)

What would you think of, instead, expanding the `_RAIBuildDependency` interface so that it included an `__is_satisfiable__` method:
```py
class _RAIBuildDependency(ABC):
    ...

    @abstractmethod
    def __is_satisfiable__(self) -> bool: ...
```

which can then be implemented by all dependencies that implement the interface:
```py
class _TFArchive(_WebTGZ, _RAIBuildDependency):
    ...

    def __is_satisfiable__(self) -> bool:
        try:
            self.url
        except BuildError:
            return False
        else:
             return True

# And presumably very similar implementations 
# for `_PTArchiveLinux`, `_PTArchiveMacOSX`, and `_ORTArchive`
```

and then checking for dependency satisfiability in `RedisAIBuilder.__init__` with a simple:
```py
class RedisAIBuilder(Builder):
    ...

    def __init__(self, ...) -> None:
        ...
        if not all(
            dep.__is_satisfiable__() for dep
            in sequence_of_deps_that_the_rai_builder_needs
        ):
            raise BuildError(f"{type(self).__name__} failed to satisfy "
                              "dependencies for this platform")
            # We can workshop the error message if want to be a bit 
            # more prescriptive, but you get the idea!
```

tests/install/test_builder.py
```diff
@@ -111,6 +123,27 @@ def test_rai_builder_raises_if_attempting_to_place_deps_in_nonempty_dir(
...
+    dict(build_tf=True, build_onnx=False),
+]
+invalid_build_ids = [
+    ",".join([f"{key}={value}" for key, value in d.items()])
+    for d in invalid_build_arm64
+]
+
+
+@pytest.mark.parametrize("build_options", invalid_build_arm64, ids=invalid_build_ids)
+def test_rai_builder_raises_if_unsupported_deps_on_arm64(monkeypatch, build_options):
```
MattToast, line 138: ditto unneeded `monkeypatch`
```suggestion
def test_rai_builder_raises_if_unsupported_deps_on_arm64(build_options):
```

tests/install/test_builder.py
```diff
@@ -205,3 +242,40 @@ def _some_long_io_op(_):
...
+    all_archs = (build.Architecture.ARM64, build.Architecture.X64)
+    for arch in all_archs:
+        os_ = build.OperatingSystem.DARWIN
+        assert isinstance(
+            build.choose_pt_variant(os_, arch, "cpu", RAI_VERSIONS.torch),
+            build._PTArchiveMacOSX,
+        )
+
+
+def test_PTArchive_MacOSX_url():
```
MattToast, line 265: Guessing this is hold over from the class name change. Maybe
```suggestion
def test_PTArchiveMacOSX_url():
```
or 
```suggestion
def test_pt_archive_macosx_url():
```

tests/install/test_builder.py
```diff
@@ -205,3 +242,40 @@ def _some_long_io_op(_):
...
+    pt_linux_cpu = build._PTArchiveLinux(build.Architecture.X64, "cpu", pt_version)
+    x64_prefix = "https://download.pytorch.org/libtorch/"
+    assert x64_prefix in pt_linux_cpu.url
+
+    pt_macosx_cpu = build._PTArchiveMacOSX(build.Architecture.ARM64, "cpu", pt_version)
+    arm64_prefix = "https://github.com/CrayLabs/ml_lib_builder/releases/download/"
+    assert arm64_prefix in pt_macosx_cpu.url
+
+
+def test_PTArchive_MacOSX_gpu_error():
```
MattToast, line 279: ditto looks like class name changed. Maybe change test name to
```suggestion
def test_PTArchiveMacOSX_gpu_error():
```
or
```suggestion
def test_pt_archive_macosx_gpu_error():
```

smartsim/_core/_install/builder.py
```diff
@@ -385,10 +395,21 @@ def __init__(
...
+        # Sanity checks
+        self._check_backends_arm64()
+
+    def _check_backends_arm64(self) -> None:
+        if self._architecture == Architecture.ARM64:
+            unsupported = []
+            if self.build_tf:
+                unsupported.append("Tensorflow")
+            if self.build_onnx:
+                unsupported.append("ONNX")
```
MattToast, line 407: Should these checks be on `self.build_{tf,onnx}` or `self.fetch_{tf,onnx}`. Is there a reason we would want to try and stop someone from compiling and trying to use their own ARM64 version of TF (or onnx)?

smartsim/_core/_install/builder.py
```diff
@@ -793,6 +801,54 @@ def __place_for_rai__(self, target: t.Union[str, "os.PathLike[str]"]) -> Path:
...
+
+def choose_pt_variant(
+    os_: OperatingSystem, arch: Architecture, device: TDeviceStr, version: str
+) -> t.Union[_PTArchiveLinux, _PTArchiveMacOSX]:
+    if os_ == OperatingSystem.DARWIN:
+        return _PTArchiveMacOSX(arch, device, version)
+    if os_ == OperatingSystem.LINUX:
+        return _PTArchiveLinux(arch, device, version)
+
+    raise BuildError(f"Unsupported OS for pyTorch: {os_}")
```
MattToast, line 849: Ditto "preferred capitalization" for torch here:
```suggestion
    raise BuildError(f"Unsupported OS for PyTorch: {os_}")
```

MattToast (changes requested):
Couple of small tweaks suggested, but otherwise LGTM!!

smartsim/_core/_install/builder.py
```diff
@@ -378,7 +381,14 @@ def __init__(
         verbose: bool = False,
     ) -> None:
         super().__init__(build_env or {}, jobs=jobs, verbose=verbose)
+
         self.rai_install_path: t.Optional[Path] = None
+        if _os not in OperatingSystem:
+            raise BuildError(f"Unsupported operating system: {_os}")
+        self._os = _os
+        if architecture not in Architecture:
+            raise BuildError(f"Unsupported architecture: {architecture}")
```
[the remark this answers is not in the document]
ashao: Now we're allowing os and arch to be passed into the Client constructor, so it's entirely possible that someone could pass in their own enum and/or inherit from `OperatingSystem` and `Architecture` and add their own.

tests/install/test_builder.py
```diff
@@ -34,10 +34,26 @@
...
+valid_platforms = (
+    dict(_os=build.OperatingSystem.DARWIN, architecture=build.Architecture.X64),
+    dict(_os=build.OperatingSystem.DARWIN, architecture=build.Architecture.ARM64),
+    dict(_os=build.OperatingSystem.LINUX, architecture=build.Architecture.X64),
+)
+extract_names = lambda a, b : ".".join((a.name, b.name)).lower()
+platform_ids = (
+    extract_names(*platform.values()) for platform in valid_platforms
+)
+for_each_platform = pytest.mark.parametrize(
```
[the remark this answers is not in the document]
ashao: Yep, good call this is leftover from when I was playing around with just testing "valid" platform configurations and then threw it away as not useful to add in this PR.

tests/install/test_builder.py
```diff
@@ -52,27 +68,24 @@
 toggle_build_pt = _toggle_build_optional_backend("pt")
 toggle_build_ort = _toggle_build_optional_backend("ort")
 
-
 @pytest.mark.parametrize(
     "mock_os", [pytest.param(os_, id=f"os='{os_}'") for os_ in ("Windows", "Java", "")]
 )
-def test_rai_builder_raises_on_unsupported_op_sys(monkeypatch, mock_os):
-    monkeypatch.setattr(platform, "system", lambda: mock_os)
+def test_os_enum_raises_on_unsupported(monkeypatch, mock_os):
```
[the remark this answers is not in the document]
ashao: Nice Catch

doc/changelog.rst
ashao: You're totally correct

smartsim/_core/_install/builder.py
ashao: Oh yes good call

smartsim/_core/_install/builder.py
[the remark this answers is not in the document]
MattToast: Correct me if I'm wrong, but I'm like 99% sure that enums will raise a `TypeError` if they are extended, so I think we are still safe to keep this static?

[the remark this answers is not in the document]
MattToast: Can confirm that they do raise if extended
```py
>>> import enum as e
>>> class E(e.Enum):
...     ONE = e.auto()
...
>>> class E2(E):
...     TWO = e.auto()
...
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.11/3.11.7/Frameworks/Python.framework/Versions/3.11/lib/python3.11/enum.py", line 485, in __prepare__
    metacls._check_for_existing_members_(cls, bases)
  File "/usr/local/Cellar/python@3.11/3.11.7/Frameworks/Python.framework/Versions/3.11/lib/python3.11/enum.py", line 936, in _check_for_existing_members_
    raise TypeError(
TypeError: <enum 'E2'> cannot extend <enum 'E'>
```

smartsim/_core/_install/builder.py
MattToast: Per conversation yesterday, this solution is not viable as we do not have all the information necessary at `__init__` time to construct the dependencies (e.g. we do not know if we are building for GPU), which should definitely be changed, but is also definitely not in the scope of this PR.

Switching to having all `_RAIBuildDependency`s to have a static method to return platform/architecture combinations that they can be built for. This will move the platform checks to each dependency which will be at least slightly more maintainable until we can move more build details into the `RedisAIBuilder.__init__`.

Makefile
```diff
@@ -66,6 +66,10 @@ clobber: clean
 # help:
 # help: Style
 # help: -------
+# help: check-all                      - Performs all the style-related checks
+.PHONY: check-all
+check-all: check-style check-format check-sort-imports check-lint check-mypy
+	$(info All style checks PASSED)
```
MattToast, line 72: good idea adding this!

doc/changelog.rst
MattToast: Re-surfacing this `pyTorch` -> `PyTorch`

smartsim/_core/_install/builder.py
```diff
@@ -756,30 +795,20 @@ def _extract_download(
...
-            raise BuildError(f"Unexpected OS for the PT Archive: {self.os_}")
-        return f"https://download.pytorch.org/libtorch/{pt_build}/{libtorch_arch}"
+    @staticmethod
+    def supported_platforms() -> t.Sequence[t.Tuple[OperatingSystem, Architecture]]:
+        # TODO: Rework this to do a recursive search
+        return tuple(
+            itertools.chain.from_iterable(
+                var.supported_platforms() for var in _PTArchive.__subclasses__()
+            )
+        )
```
MattToast, line 811: Not wrong just be warned: If someone subclass this class and forgets to re-implement this method, that will lead to a recursion error

MattToast (approved):
Couple of last minute small call-outs, but nothing worth holding approval up over. LGTM pending tests!!

smartsim/_core/_install/builder.py
ashao: Yah that was the point of the above comment, but I'll make it a bit clearer

MattToast: Oh! Is that what that meant? I thought that was talking about if someone where to subclass `_PTArchiveLinux` or `_PTArchiveMacOSX`.


I meant something like
```py
calss _PTArchiveWindows(_PTArchive):
    # Very explicitly not overwriting `supported_platforms`
    ...
```
will now raise a
```
RecursionError: maximum recursion depth exceeded
```

smartsim/_core/_install/builder.py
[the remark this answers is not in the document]
ashao: Ah good. Ok then, removing lines.

smartsim/_core/_install/builder.py
ashao: Ohh I see, yeah that makes sense.

PR #466, opened 2024-01-29
Model preview

juliaputko:


templates/templates/preview/base_html.template
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+<body>
+{% include "preview/experiment_html.template" %}
+{%- if manifest.all_entity_lists|length > 0 %}
+<p>=== Entity Preview ===<p/>
```
ankona, line 6: invalid html tag -> meant `</p>`

ankona, line 6: - Consider using container elements to group items.
- Consider setting classes on semantic elements to enable applying CSS 
- Consider looping in the parent (base_html.template) for databases so the database template doesn't have to worry about "sets of databases"

```
<div class='preview'>   <------- container for a preview will enable things like expand/contract on the entire element  -->
    <div class='header'>Experiment Preview</div>   <------- header class lets us bold/font-face/font-size w/CSS
    <div class='experiment'>
        {% include "preview/experiment_html.template %}
    </div>
    <div class='childContainer'>     <------- could help us do things like alternate colors on rows for different types of children
    <div class='childEntity orchestrator'>  <----- can help us do clever things like click a filter in a front end to show only orchestrators, etc
        {% for orc in manifest.dbs %} <----- loop in parent
            {% include "preview/orchestrator_html.template" %}
        {% endfor %}
    </div>
    </div>
</div>
```

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _check_verbosity_level(
+    verbosity_level: _VerbosityLevelString,
+) -> _VerbosityLevelString:
+    """
+    Check that the given verbosity level is valid.
+    """
+
+    if verbosity_level not in ["info", "debug", "developer"]:
```
ankona, line 111: Conditional+message is not logical. Why check for three values and output that info is the only valid one? why not check for info?

templates/templates/preview/model.test
```diff
@@ -0,0 +1,112 @@
+
+== Models ==
```
ankona: If you need templates for unit tests, they shouldn't be in the set that is shipped. 

- Consider moving these to `/tests/test_configs` instead
- Consider using the same naming scheme
  - currently: `model.test` and `model.template`
  - maybe: `model_XXX.tpl` and `model_YYY.tpl`
  - I don't know the real difference for why each template is unique at a glance, but replace xxx/yyy w/purpose

templates/templates/preview/test.template
```diff
@@ -0,0 +1,117 @@
+{% for model in manifest.models %}
```
ankona: - ditto on location of test template
- ditto on naming convention

templates/templates/preview/test.template
```diff
@@ -0,0 +1,117 @@
...
+{% if model is defined %}
+== Models ==
```
ankona: There appear to be no test templates for HTML output. Are we not testing HTML/file-based output?

tests/test_preview.py
```diff
@@ -0,0 +1,271 @@
...
+    hello_world_model.attach_generator_files(to_configure=[config])
+    exp.generate(hello_world_model, overwrite=True)
+
+    preview_manifest = Manifest(hello_world_model, spam_eggs_model)
+
+    # Execute preview method
+    rendered_preview = previewrenderer.render(exp, preview_manifest)
+
+    # Evaluate output for hello_world_model
+    assert "Model name" in rendered_preview
```
ankona, line 193: Consider separate tests or renaming test.
- test named "test w/tagged files" but also validates model params, batch args, password, etc
- split test so asserts for model params don't cloud tests for outputs related to files

tests/test_preview.py
```diff
@@ -0,0 +1,271 @@
...
+    model = exp.create_model("model_test", run_settings=rs1)
+    # enable key prefixing on ensemble
+    model.enable_key_prefixing()
+    exp.generate(model, overwrite=True)
+
+    # Run preview render
+    preview_manifest = Manifest(db, model)
+    output = previewrenderer.render(exp, preview_manifest)
+
+    assert "Client Configuration" in output
```
ankona, line 249: consider splitting tests per above logic:

- this named test would only look for key prefixing outputs
- I think ONLY line `253` w/ `... (key prefixing)` looks at prefix related output
  - is there anything else it can lok for?
  - should we check that the _prefix_ is in the output?

tests/test_preview.py
```diff
@@ -0,0 +1,271 @@
...
+    Test error when invalid ouput format is given.
+    """
+    # Prepare entities
+    exp_name = "test_output_format"
+    exp = Experiment(exp_name)
+
+    # Execute preview method
+    with pytest.raises(ValueError) as ex:
+        exp.preview(output_format="hello")
+    assert "The only valid currently available is html" in ex.value.args[0]
```
ankona, line 271: - try not to replicate entire strings in tests (because this one is nonsense and needs to change already)
- Consider creating a `PreviewException` base class
  - Consider raising a `PreviewFormat` child class

templates/templates/preview/experiment_html.template
```diff
@@ -0,0 +1,6 @@
+<p>=== Experiment Overview ===</p>
```
ankona, line 1: - add your parent div here (around entire entity)

templates/templates/preview/experiment_html.template
```diff
@@ -0,0 +1,6 @@
...
+<ul style="list-style-type:none;">
```
ankona, line 2: Don't do inline styles.
- move style declarations to the base template, set class here to something like `entityList` so we don't have to change inline styles

templates/templates/preview/model.template
```diff
@@ -0,0 +1,87 @@
...
+        {% endif %}
+        {%- if model.files %}
+        {%- if model.files.tagged %}
+        Tagged Files for model configuration: 
+            {{ model.files.tagged[0] }} -> 
+            {{ model.path }}
+        {%- endif -%}
+        {%- if model.files.copy %}
+        Copy files: 
+            {{ model.files.copy[0] }} ->
```
ankona, line 29: is hardcoding the 0-th element ok?

templates/templates/preview/model.template
```diff
@@ -0,0 +1,87 @@
...
+            {{ model.path }}
+        {%- endif %}
+        {%- if model.files.link %}
+        Symlink: 
+            {{ model.files.link[0] }} ->
```
ankona, line 34: ditto on 0 index hardcoding

templates/templates/preview/model.template
```diff
@@ -0,0 +1,87 @@
...
+        {% if model.params -%}
+        Model parameters: 
+        {%- for param, value in model.params.items() %}
+            {{ param }}: {{ value }}
+        {%- endfor -%}
+        {% endif %}
+        {%- if model.files %}
+        {%- if model.files.tagged %}
+        Tagged Files for model configuration: 
+            {{ model.files.tagged[0] }} ->
```
ankona, line 24: ditto on 0 index hardcoding

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,120 @@
...
+    return rendered_preview
+
+
+def preview_to_file(content: str, filename: str) -> None:
+    """
+    Output preview to a file if output format and filename
+    are specified.
+    """
+
+    with open(filename, "w", encoding="utf-8") as prev_file:
```
ankona, line 92: docstring is misleading. this function attempts to output, no matter what the content/filename are

templates/templates/preview/base.template
```diff
@@ -0,0 +1,17 @@
...
+    == Models == 
+{%- for model in manifest.models %}   
+    = {% include "preview/model.template" %}
+    {% if manifest.dbs or model.run_settings.colocated_db_settings %}
+    {% include "preview/clientconfig.template" %}
+    {%- endif %}
+    =
+{%- endfor %}
+{%- endif -%}
+{%- endif -%}
```
ankona, line 17: don't anger the github gods! add your missing trailing newline

templates/templates/preview/clientconfig.template
```diff
@@ -0,0 +1,47 @@
...
+                DataSets: On 
+                Models/Scripts: Off
+                Aggregation Lists: On
+            {%- endif %}
+            {%- if model.query_key_prefixing() == False %}
+                Tensors: Off 
+                DataSets: Off
+                Models/Scripts: Off
+                Aggregation Lists: Off
+            {%- endif %}
```
ankona, line 47: don't anger the github gods! add your missing trailing newline

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,120 @@
...
+    if verbosity_level not in ["info", "debug", "developer"]:
+        raise ValueError("The only valid verbosity level currently available is info")
+
+    if verbosity_level in ("debug", "developer"):
+        logger.warning(
+            f"'{verbosity_level}' is an unsupported verbosity level requested.\
+Setting verbosity to: info"
+        )
+        return "info"
+    return "info"
```
ankona, line 120: - why are we repeatedly hard-coding these strings?
    - [consider](https://stackoverflow.com/questions/58608361/string-based-enum-in-python) `class Verbosity(str, enum)`

smartsim/experiment.py
```diff
@@ -44,6 +44,9 @@
 
 logger = get_logger(__name__)
 
+_OutputFormatString = t.Optional[t.Literal["html"]]
+_VerbosityLevelString = t.Literal["info", "debug", "developer"]
```
ankona, line 48: - _VerbosityLevelString is redefined in experiment after definition in `previewrenderer.py`
- _OutputFormatString is redefined in experiment after definition in `previewrenderer.py`

tests/test_preview.py
```diff
@@ -0,0 +1,271 @@
...
+    exp = Experiment(exp_name, exp_path=test_dir, launcher=test_launcher)
+
+    # Execute method for template rendering
+    output = previewrenderer.render(exp)
+
+    # Evaluate output
+    summary_lines = output.split("\n")
+    summary_lines = [item.replace("\t", "").strip() for item in summary_lines[-3:]]
+    assert 3 == len(summary_lines)
+    summary_dict = dict(row.split(": ") for row in summary_lines)
```
ankona, line 63: this is clever and i love it, even though i hate having to do the strips to get here...

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,120 @@
...
+
+    verbosity_level = _check_verbosity_level(verbosity_level)
+
+    loader = jinja2.PackageLoader("templates")
+    env = jinja2.Environment(loader=loader, autoescape=True)
+
+    version = f"_{output_format}" if output_format else ""
+    tpl_path = f"preview/base{version}.template"
+
+    if output_format:
```
ankona, line 71: I dislike that we use `output_format` to indicate both the format AND that it will output to a file. 
- Consider setting the default output_format to `text`
- Consider avoiding a side effect (writing a file) when rendering
  - allow _caller_ to decide what to do with it


Consider this example that lets us add CLI arguments to let the user specify format & path. Who cares if they output text to a file or html to console. If that's what they want, _give it to them_.

```py 
class Experiment:
  ...
  def preview(self, args: arparse.Namespace) -> None:
    format = args.format
    
    # the renderer doesn't render AND write, it just renders
    rendered = previewrenderer.render(self, format=format, ...)
    

    if args.output_path:
        # write file doesn't care if it's a preview, only a path & some content
        # so it's less likely to have to add branching vs nesting it in render code
        self.write_file(args.output_path, content=rendered)
```

ankona: an alternative might be to pass `output_path` to `previewrenderer.render` to avoid having `format` represent both an output format AND place to render

templates/templates/preview/base_html.template
juliaputko: resolved

smartsim/_core/control/previewrenderer.py
juliaputko: changed

templates/templates/preview/model.test
juliaputko: removed test templates (a mistake)

templates/templates/preview/test.template
juliaputko: removed

templates/templates/preview/test.template
juliaputko: removed

juliaputko: added html output templates

templates/templates/preview/experiment_html.template
juliaputko: changed

templates/templates/preview/model.template
juliaputko: good point. Changed to for loop. Will have to double check how multiple files should be displayed

tests/test_preview.py
```diff
@@ -0,0 +1,271 @@
...
+from smartsim.settings import RunSettings
+
+
+@pytest.fixture
+def choose_host():
+    def _choose_host(wlmutils, index=0):
+        hosts = wlmutils.get_test_hostlist()
+        if hosts:
+            return hosts[index]
+        else:
```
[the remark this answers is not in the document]
juliaputko: changed

smartsim/_core/control/previewrenderer.py
juliaputko: changed

smartsim/experiment.py
juliaputko: removed

smartsim/_core/control/previewrenderer.py
juliaputko: changed

templates/templates/preview/model.template
juliaputko: changed

templates/templates/preview/model.template
juliaputko: changed

tests/test_preview.py
juliaputko: added previewexception and previewformaterror

tests/test_preview.py
juliaputko: separated tests

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,120 @@
...
+    with open(filename, "w", encoding="utf-8") as prev_file:
+        prev_file.write(content)
+
+
+def _check_output_format(output_format: str) -> None:
+    """
+    Check that the output format given is valid.
+    """
+    if not output_format == "html":
+        raise ValueError("The only valid currently available is html")
```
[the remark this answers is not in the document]
juliaputko: changed

tests/test_preview.py
juliaputko: separated tests

templates/templates/preview/experiment_html.template
juliaputko: changed

templates/templates/preview/clientconfig.template
juliaputko: added newline

smartsim/_core/control/previewrenderer.py
juliaputko: getting mypy error about the return type. Will not accept enum as the same type as _VerbosityLevelString = t.Literal["info", "debug", "developer"]

smartsim/_core/control/previewrenderer.py
juliaputko: That is what the purpose of output_filename is now. If a user passes a filename, the output will be output to that file instead of to stdout. Do you think that is okay?

smartsim/_core/control/manifest.py
```diff
@@ -109,6 +109,12 @@ def all_entity_lists(self) -> t.List[EntitySequence[SmartSimEntity]]:
 
         return _all_entity_lists
 
+    @property
```
ankona, line 112: The new property `all_entities` is only used to evaluate as a boolean for presence of entities in the manifest. If we didn't already have a ton of properties listing each individual item, I probably wouldn't notice it. 

### Do you think the API is better if we instead indicate what you want to know directly? 

```py
@property
def has_deployable(self) -> bool:
   """Return True if the manifest contains entities that must be physically deployed"""
   return bool(self._deployables)

```
which results in
```html
{%- if manifest.has_deployables %}
...
```

### Could we just avoid adding another list property? The current usage is only to show models, so we could just do:

```html
{%- if manifest.models %}
...
```

ankona: ### Creating a View Model

An alternative is to build a facade over the manifest that gives the preview-centric view properties. This is a "ViewModel" whose purpose is to aggregate or reshape existing data into a format that is useful for display, but doesn't fulfill the original purpose of the object you're interested in.

Views don't make a lot of immediate sense, but by the time it does we might have added a ton of manifest properties that we no longer recognize as only used by the manifest. 

Consider:

```py

class ManifestPreviewView:
  def __init__(self, manifest: Manifest) -> None:
    self._manifest = manifest

  def has_deployables(self) -> bool:
    return any(manifest.models)
  
  def __attr__(self, attr_name: string) -> t.Any:
    if hasattr(self._manifest, attr_name):
      return getattr(self._manifest, attr_name)

```
 
This creates a purpose built way to aggregate things used in a preview and doesn't start to degrade the API of an existing object

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,119 @@
...
+from ..._core.control import Manifest
+from ...error.errors import PreviewFormatError
+from ...log import get_logger
+
+logger = get_logger(__name__)
+
+if t.TYPE_CHECKING:
+    from smartsim import Experiment
+
+_OutputFormatString = t.Optional[t.Literal["plain_text"]]
```
ankona, line 42: [optional nit] Let's consider the same question discussed offline here... Do we want to expose an enum further up the call stack and avoid passing hardcoded strings, similar to `Verbosity`?

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    return rendered_preview
+
+
+def preview_to_file(content: str, filename: str) -> None:
+    """
+    Output preview to file.
+    """
+
+    with open(filename, "w", encoding="utf-8") as prev_file:
```
ankona, line 95: Are you ok with overwriting old files? Do you want to proactively append a unique-ifier to the file name if a user wants to compare previews?

```py
def find_available_filename(filename: str) -> str:
    """Iterate through potentially unique names until one is found that does 
    not already exist. Return an unused name variation"""
    path = pathlib.Path(filename)
    candidate_path = pathlib.Path(filename)
    index = 1

    while candidate_path.exists():
        candidate_path = path.with_stem(f"{path.stem}_{index}")
        index += 1
    
    return candidate_path.name
```

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,119 @@
...
+        prev_file.write(content)
+
+
+def _check_file_output_format(output_format: _OutputFormatString) -> None:
+    """
+    Check that a valid file output format is given.
+    """
+    if not output_format == "plain_text":
```
ankona, line 103: potentially want to use your enum for `_OutputFormatString` instead

smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+
+def _check_verbosity_level(
+    verbosity_level: Verbosity,
+) -> Verbosity:
+    """
+    Check that the given verbosity level is valid.
+    """
+    if verbosity_level not in (Verbosity.INFO, Verbosity.DEBUG, Verbosity.DEVELOPER):
```
ankona, line 115: I think moving to our `Verbosity` enum makes this check obsolete?

smartsim/entity/dbobject.py
```diff
@@ -275,6 +276,7 @@ def __init__(
         self.min_batch_timeout = min_batch_timeout
         self.tag = tag
         self.inputs, self.outputs = self._check_tensor_args(inputs, outputs)
+        self.model_file = model_file
```
ankona, line 279: an easier to grok validation for model/model_file might be:

```py
if not any(model, model_file):
   
```

templates/templates/preview/clientconfig_plain_text.template
```diff
@@ -0,0 +1,51 @@
+   
+        Client Configuration:
+            {%- if manifest.dbs %}
+            {%- if manifest.dbs[0].name %}
+            Database identifier: {{ manifest.dbs[0].name }}
```
ankona, line 5: is hardcoded 0-th index ok?

templates/templates/preview/clientconfig_plain_text.template
```diff
@@ -0,0 +1,51 @@
...
+            {%- endif %}
+            {%- endif %}
+            {%- if incoming in model.incoming_entities %}
+            Incoming entities (available data sources): 
+            {%- for incoming in model.incoming_entities %}
+                {{ incoming.name }}
+            {%- endfor %}
+            {%- endif %}
+            Outgoing key collision prevention (key prefixing): 
+            {%- if model.query_key_prefixing() == True %}
```
ankona, line 40: is ` == True` extraneous?

templates/templates/preview/clientconfig_plain_text.template
```diff
@@ -0,0 +1,51 @@
...
+                Tensors: On 
+                DataSets: On 
+                Models/Scripts: Off
+                Aggregation Lists: On
+            {%- endif %}
+            {%- if model.query_key_prefixing() == False %}
```
ankona, line 46: `{% else %}`?

templates/templates/preview/clientconfig_plain_text.template
```diff
@@ -0,0 +1,51 @@
...
+            {%- if incoming in model.incoming_entities %}
+            Incoming entities (available data sources): 
+            {%- for incoming in model.incoming_entities %}
+                {{ incoming.name }}
+            {%- endfor %}
+            {%- endif %}
+            Outgoing key collision prevention (key prefixing): 
+            {%- if model.query_key_prefixing() == True %}
+                Tensors: On 
+                DataSets: On
```
ankona, line 42: Could we turn this into a single section without nesting using a custom filter? Does that pattern help us simplify other things... if not, ignore...

```
{% if xxx %}
                Tensors: Off 
                DataSets: Off
                Models/Scripts: Off
                Aggregation Lists: Off
{% else %}
                Tensors: On 
                DataSets: On
                Models/Scripts: On
                Aggregation Lists: On
{% endif %}
```

could become:

```
                Tensors: {% model.query_key_prefixing | as_toggle %} 
                DataSets: {% model.query_key_prefixing | as_toggle %} 
                Models/Scripts: {% model.query_key_prefixing | as_toggle %} 
                Aggregation Lists: {% model.query_key_prefixing | as_toggle %} 
```

with a [custom filter](https://jinja.palletsprojects.com/en/3.1.x/api/#custom-filters) in the vein of:

```py
@pass_eval_context
def as_toggle(eval_ctx, value: bool):
  return "On" if value else "Off"

```

templates/templates/preview/model_plain_text.template
```diff
@@ -0,0 +1,97 @@
...
+                {%- endfor %}
+            {%- endfor %}
+            {%- endif %}
+            {%- endif %}
+            {%- if manifest.models %}
+            {%- if manifest.models[0].query_key_prefixing() == True %}  
+            Key prefix: {{ model.name }}
+            {%- endif %}
+            {%- endif %}
+    {%- endif -%}
```
ankona, line 97: nitpick: our if/endif wrapping the whole thing don't have the same indentation. could we fix the endif?

maybe also use jinja to get rid of whitespace and shift the first line down off the if clause just for readability?

smartsim/_core/control/previewrenderer.py
juliaputko: good catch

smartsim/_core/control/previewrenderer.py
juliaputko: changed to enum

smartsim/_core/control/previewrenderer.py
juliaputko: changed to enum

templates/templates/preview/clientconfig_plain_text.template
juliaputko: yes. Removed

templates/templates/preview/clientconfig_plain_text.template
juliaputko: Changed

templates/templates/preview/model_plain_text.template
juliaputko: changed

smartsim/_core/control/previewrenderer.py
juliaputko: included

templates/templates/preview/clientconfig_plain_text.template
juliaputko: did some refactor of how client configuration was shown for each entity

templates/templates/preview/clientconfig_plain_text.template
juliaputko: implemented a custom filter

PR #467, opened 2024-01-29
Fix test_logs to prevent generation of dir

al-rigazzi:
This PR fixes the `test_logs.py::test_context_leak` test that was erroneously creating a directory named `some value` in SmartSim's root directory.


MattToast (approved):
Good catch! Thanks for cleaning up the clutter!!

ankona (approved):
Thanks for cleaning up the mess I left behind me!

PR #468, opened 2024-01-29
Expose SS Typehints

MattToast:
Add and ship `py.typed` marker per [pep-561](https://peps.python.org/pep-0561/). This change will help change the `mypy` output for this script
```py
from smartredis import Client

exp = Experiment(b"my-exp")  # Should err expected str
client = Client(False)

reveal_type(exp)
reveal_type(client)
```
from this
```sh
$ mypy script.py
script.py:1: error: Skipping analyzing "smartsim": module is installed, but missing library stubs or py.typed marker  [import-untyped]
script.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
script.py:2: error: Skipping analyzing "smartredis": module is installed, but missing library stubs or py.typed marker  [import-untyped]
script.py:7: note: Revealed type is "Any"
script.py:8: note: Revealed type is "Any"
Found 2 errors in 1 file (checked 1 source file)
```
to this
```sh
$ mypy script.py
script.py:4: error: Argument 1 to "Experiment" has incompatible type "bytes"; expected "str"  [arg-type]
script.py:7: note: Revealed type is "smartsim.experiment.Experiment"
script.py:8: note: Revealed type is "smartredis.client.Client"
Found 1 error in 1 file (checked 1 source file)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,18 +91,32 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 disallow_untyped_decorators = true
 
+# Probably Unintended Branches/Blocks
+# warn_unreachable = true
+
 # Safety/Upgrading Mypy
 warn_unused_ignores = true
 warn_redundant_casts = true
 warn_unused_configs = true
 show_error_codes = true
 
+# Misc Strictness Settings
+strict_concatenate = false
+strict_equality = true
+
+# Additional Error Codes
+enable_error_code = [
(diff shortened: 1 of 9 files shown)


al-rigazzi (approved):
LGTM! Thanks for taking another step towards modern Python!

al-rigazzi (approved):
Re-approving and praising the effort of catching all those type errors!

PR #469, opened 2024-01-29
Ensemble preview

juliaputko:


smartsim/_core/control/previewrenderer.py
```diff
@@ -0,0 +1,123 @@
...
+from ...log import get_logger
+
+logger = get_logger(__name__)
+
+if t.TYPE_CHECKING:
+    from smartsim import Experiment
+
+
+_OutputFormatString = t.Optional[t.Literal["plain_text", "html"]]
+_VerbosityLevelString = t.Literal["info", "debug", "developer"]
```
ankona, line 44: 1. Do we need both `_VerbosityLevelString` and `Verbosity`?
2. If we add `Verbosity.DEFAULT = "info"` does it allow us to avoid hardcoding defaults (see [line 122](https://github.com/CrayLabs/SmartSim/pull/469/files#diff-3c7fd203fcb3ac4ad98cde50342ee0f7fb41da902f2f806b88c6c92e6012a3bcR122))?

tests/test_preview.py
```diff
@@ -0,0 +1,706 @@
...
+    # Execute preview method
+    exp.preview(
+        hello_world_model,
+        spam_eggs_model,
+        output_format="html",
+        output_filename=str(path),
+    )
+
+    # Evaluate output
+    assert path.exists()
```
ankona, line 135: 1. Should we check that HTML tags exist or anything else about the content of the output file? I'd consider looking for something like `<html>` just to be sure it didn't fall back on a "default" format

2. What happens if i pass a path that is invalid? We might already have tests for the core functionality of preview like handling bad paths.

3. What happens if i give a output format that is valid but the template is incorrect. Ditto on potentially already covered.

tests/test_preview.py
```diff
@@ -0,0 +1,706 @@
...
+    spam_eggs_model = exp.create_model("echo-spam", run_settings=rs2)
+    hello_ensemble = exp.create_ensemble("echo-ensemble", run_settings=rs1, replicas=3)
+
+    exp.generate(hello_world_model, spam_eggs_model, hello_ensemble)
+
+    preview_manifest = Manifest(hello_world_model, spam_eggs_model, hello_ensemble)
+    output = previewrenderer.render(exp, preview_manifest)
+
+    assert "Models" in output
+    assert "echo-hello" in output
```
ankona, line 346: nitpic: we should use vars to avoid minor typos or having to update multiple places in the case of test updates. e.g.
```py
hw_name = "echo-hello"
...
assert hw_name in output
```

tests/test_preview.py
```diff
@@ -0,0 +1,706 @@
...
+    assert "generator_files/test_dir" in output
+    assert "generator_files/to_copy_dir" in output
+    assert "generator_files/to_symlink_dir" in output
+    assert "/SmartSim" in output
+    for model in ensemble:
+        assert "generator_files/test_dir" in model.files.tagged[0]
+        for copy in model.files.copy:
+            assert "generator_files/to_copy_dir" in copy
+        for link in model.files.link:
+            assert "generator_files/to_symlink_dir" in link
```
ankona, line 438: Does each copy/link in each ensemble always get the same value? if they have slight variations

- /smartsim/generator/test/*modela*/generator_files/to_copy
- /smartsim/generator/test/*modelb*/generator_files/to_copy
- /smartsim/generator/test/*modelc*/generator_files/to_copy

it might be worthwhile to ensure a bug didn't crop up where we put the same thing for all of them through a template mistake. If they are the same, we're good.

tests/test_preview.py
```diff
@@ -0,0 +1,706 @@
...
+        "colocated_ens", run_settings=colo_settings, replicas=2
+    )
+
+    # Create colocated SmartSim Model
+    colo_model = exp.create_model("colocated_model", colo_settings)
+
+    # Create and save ML model to filesystem
+    content = "empty test"
+    path = pathlib.Path(test_dir) / "model1.pt"
+    with open(path, "w", encoding="utf-8") as prev_file:
```
ankona, line 471: pathlib lets you do something like:
```py
path.write_text(content)
```
so you could avoid the extra context manager line

tests/test_preview.py
```diff
@@ -0,0 +1,706 @@
...
+        prev_file.write(content)
+
+    model_path = test_dir + "/model1.pt"
```
ankona, line 474: same path as `path` above? intentional? can we re-use it w/`str(path)` as arg to `.add_ml_model`?

smartsim/_core/control/previewrenderer.py
juliaputko: 1. I like _verbositylevelstring for type stuff. I think you suggested the Verbosity class to replace "info", but it is causing conflicts with _Verbositylevelstring
2.  I was getting a mypy conflict when I tried that:  error: Incompatible return value type (got "Literal[Verbosity.INFO]", expected "Literal['info', 'debug', 'developer']") .

tests/test_preview.py
juliaputko: removing html from this PR. Will keep this note for future

tests/test_preview.py
juliaputko: changed to vars in tests

tests/test_preview.py
juliaputko: paths all look the same.

tests/test_preview.py
juliaputko: change to write_text

templates/templates/preview/model_plain_text.template
```diff
@@ -0,0 +1,96 @@
...
+            Tagged Files for model configuration: 
+                {%- for tagged in model.files.tagged %}
+                {{ tagged }} -> 
+                {{ model.path }}
+                {%- endfor %}
+            {%- endif -%}
+            {%- if model.files.copy %}
+            Copy files: 
+                {%- for copy in model.files.copy %}
+                {{ copy }} ->
```
ankona, line 32: not sure but did you want newlines between these `source -> target` things? Multiple copies might be weird looking in output, e.g.

```
aaa ->
bbb
ccc ->
ddd
eee ->
fff
```
vs 
```
aaa -> bbb
ccc -> ddd
eee -> fff
```

juliaputko: left as they because readability it easier this way

PR #470, opened 2024-01-30
Dragon Launcher Prototype

al-rigazzi:
This is the first prototype of the new Dragon-based launcher.

Issues to fix (or to defer to a future PR):
- [x] Several telemetry test errors
- [ ] Missing batch settings for Dragon launcher
- [x] Use Process Groups for MPI-based applications - we have process groups but Dragon team is working on a more complete API


smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,158 @@
...
+            "stop": self.stop,
+            "handshake": DragonBackend.handshake,
+        }
+        self._proc_groups: t.Dict[str, t.Tuple[ProcessGroup, t.List[int]]] = {}
+        self._step_id_lock = RLock()
+        self._step_id = 0
+
+    def _get_new_id(self) -> str:
+        with self._step_id_lock:
+            self._step_id += 1
```
ankona, line 78: Should we avoid auto-incrementing IDs and use UUID instead? Even w/the lock here, it makes me feel safer moving forward and avoids mistakes where you could mistakenly point to the wrong thing by passing the wrong integer.

smartsim/_core/launcher/dragon/dragonLauncher.py
```diff
@@ -0,0 +1,362 @@
...
+        self._context.setsockopt(zmq.SNDTIMEO, value=timeout)
+        self._context.setsockopt(zmq.RCVTIMEO, value=timeout)
+
+    def connect_to_dragon(self, path: str) -> None:
+        with self._comm_lock:
+            # TODO use manager instead
+            if self.is_connected:
+                return
+
+            dragon_config_log = os.path.join(path, "dragon_config.log")
```
ankona, line 123: Do you want to push the output file name into config.py::Config to make it easier to let other components touch it?

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: Yeah, do you know if a short version of UUIDs exists? Long UUIDs look intimidating in the TaskManager output

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,158 @@
...
+            return str(self._step_id)
+
+    def process_request(self, request: DragonRequest) -> DragonResponse:
+        req_type = DragonRequest.parse_obj(request).request_type
```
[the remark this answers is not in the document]
al-rigazzi: Quite sure you're right

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,158 @@
...
+        proc = TemplateProcess(
+            target=run_request.exe,
+            args=run_request.exe_args,
+            cwd=run_request.path,
+            env={**run_request.current_env, **run_request.env},
+            # stdout=Popen.PIPE,
+            # stderr=Popen.PIPE,
+        )
+
+        grp = ProcessGroup(restart=False, pmi_enabled=True)
```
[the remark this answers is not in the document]
al-rigazzi: Indeed we did.

smartsim/_core/launcher/dragon/dragonLauncher.py
```diff
@@ -0,0 +1,362 @@
...
+        self._context.setsockopt(zmq.RCVTIMEO, value=self._timeout)
+        self._dragon_head_socket: t.Optional[zmq.Socket[t.Any]] = None
+        self._dragon_head_process: t.Optional[subprocess.Popen[bytes]]
+        self._comm_lock = RLock()
+
+    @property
+    def is_connected(self) -> bool:
+        return self._dragon_head_socket is not None
+
+    def _handsake(self, address: str) -> None:
```
MattToast, line 94: Small typo I found while implementing against this PR that I do not want to forget to address:
```suggestion
    def _handshake(self, address: str) -> None:
```

al-rigazzi: LOL

smartsim/_core/launcher/dragon/dragonBackend.py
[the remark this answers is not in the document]
al-rigazzi: It's been fixed by @MattToast

smartsim/_core/launcher/dragon/dragonLauncher.py
al-rigazzi: Done done!

MattToast (commented):
Just a couple more small things I found while implementing against this PR

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+def main(args: argparse.Namespace) -> int:
+    interface, address = get_best_interface_and_address()
+    if not interface:
+        raise ValueError("Net interface could not be determined")
+    dragon_head_address = f"tcp://{address}"
+
+    if args.launching_address:
+        if str(args.launching_address).split(":", maxsplit=1)[0] == dragon_head_address:
+            address = "localhost"
+            dragon_head_address = "tcp://localhost:5555"
```
ankona, line 119: consider reducing nested/repetitive string-building and pulling out single parts being computed. it could make it easier to find the one place where the string is put together and ensure they're consistent.

```py
def main()...
  ...
  transport = "tcp://"
  port = 5555
  if dragon_head_address not in args.launching_address:
    address = args.launching_address
  if port not in args.launching_port:  # example of another component, not really here
    port = args.launching_port

  dragon_head_address = f"{transport}{address}:{port}"
    
```

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+
+    run(dragon_head_address=dragon_head_address)
+
+    print("Shutting down! Bye bye!")
+    return 0
+
+
+def cleanup() -> None:
+    global SHUTDOWN_INITIATED  # pylint: disable=global-statement
+    print("Cleaning up", flush=True)
```
ankona, line 151: Can we replace the prints with a `logging.Logger`?

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+Dragon server entrypoint script
+"""
+
+DBPID: t.Optional[int] = None
+
+
+def print_summary(network_interface: str, ip_address: str) -> None:
+    zmq_config = {"interface": network_interface, "address": ip_address}
+
+    with open("dragon_config.log", "w", encoding="utf-8") as dragon_config_log:
```
ankona, line 75: Can we pull this file name out and put it in `smartsim._core.config.Config` in case something else needs to touch it (and we want to avoid potentially changing the name in multiple places)

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+        dragon_config_log.write(
+            textwrap.dedent(f"""\
+                -------- Dragon Configuration --------
```
ankona, line 78: Is this output consistent with expectations for the preview feature? Do we want it to be able to be used there?

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+
+def handle_signal(signo: int, _frame: t.Optional[FrameType]) -> None:
+    if not signo:
+        print("Received signal with no signo")
+    else:
+        print(f"Received {signo}")
+    cleanup()
+
+
+context = zmq.Context()
```
ankona, line 63: Can we move `context` instantiation out of global space?

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+    print(f"Opening socket {dragon_head_address}")
+    dragon_head_socket = context.socket(zmq.REP)
+    dragon_head_socket.bind(dragon_head_address)
+    dragon_backend = DragonBackend()
+
+    while not SHUTDOWN_INITIATED:
+        print(f"Listening to {dragon_head_address}")
+        req = dragon_head_socket.recv_json()
+        print(f"Received request: {req}")
+        drg_req = request_serializer.deserialize_from_json(str(req))
```
ankona, line 102: I think the `SchemaSerializer` might not be the best choice. Pydantic should give us a ton of power over JSON serialization behaviors already.

What's wrong with built-ins?

```
import pydantic


class Book(pydantic.BaseModel):
    title: str
    num_pages: int


book = Book(title="The shortest title in the entire world", num_pages=1)

json = book.model_dump_json()
dict_ = book.model_dump()

print(json)
print(dict_)


```

output:

```bash
-- /Users/chris.mcbride/code/zmq_multi/.vscode/pydantic_serialization.py 
{"title":"The shortest title in the entire world","num_pages":1}
{'title': 'The shortest title in the entire world', 'num_pages': 1}
```

smartsim/_core/entrypoints/dragon.py
```diff
@@ -0,0 +1,178 @@
...
+            raise ValueError(
+                "Could not receive connection confirmation from launcher. Aborting."
+            )
+
+    print_summary(interface, dragon_head_address)
+
+    run(dragon_head_address=dragon_head_address)
+
+    print("Shutting down! Bye bye!")
+    return 0
```
ankona, line 146: We don't catch any exceptions in here. Can we add something so we ensure errors are logged?

smartsim/_core/entrypoints/redis.py
```diff
@@ -179,6 +191,14 @@ def cleanup() -> None:
         action="store_true",
         help="Specify if this orchestrator shard is part of a cluster",
     )
+    parser.add_argument(
+        "+redirect_output",
+        action="store_true",
+        help=(
+            "Specify if stdout and stder of this script should be redirected. "
```
ankona, line 198: obligatory typo nitpick! stder -> stderr

smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -591,6 +599,7 @@ def main(
     )
 
     cooldown_duration = cooldown_duration or CONFIG.telemetry_cooldown
+    logger.warning(cooldown_duration)
```
ankona, line 602: Can we add text here to say what the number is? 

`logger.warning(f"Cooldown period (seconds): {cooldown_duration}")`

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+    and will only be called by the Dragon entry-point script or
+    by threads spawned by it.
+    """
+
+    def __init__(self) -> None:
+        self._proc_groups: t.Dict[str, t.Tuple[ProcessGroup, t.List[int]]] = {}
+        self._step_id_lock = RLock()
+        self._step_id = 0
+
+    def _get_new_id(self) -> str:
```
ankona, line 71: Should we avoid auto-incremented IDs to ensure we don't have errors that look fine but are hard to solve because the IDs are easily guessable/mistakeable (e.g. i ask for status with "item_count" instead of "item_id" and always get item "2")?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+        with self._step_id_lock:
+            self._step_id += 1
+            return str(self._step_id)
+
+    @functools.singledispatchmethod
+    # Deliberately suppressing errors so that overloads have the same signature
+    # pylint: disable-next=no-self-use
+    def process_request(self, request: DragonRequest) -> DragonResponse:
+        raise TypeError("Unsure how to process a `{type(request)}` request")
```
ankona, line 80: f-string appears to miss an `f"..."`

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @process_request.register
+    def _(self, request: DragonRunRequest) -> DragonRunResponse:
+
+        proc = TemplateProcess(
+            target=request.exe,
+            args=request.exe_args,
+            cwd=request.path,
+            env={**request.current_env, **request.env},
```
ankona, line 89: If req.current_env and req.env both have the same value, i think the `current_env` value will be overwritten. Is that the precedence order you want?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+        step_id = self._get_new_id()
+        grp.init()
+        grp.start()
+        self._proc_groups[step_id] = (grp, grp.puids)
+
+        return DragonRunResponse(step_id=step_id)
+
+    @process_request.register
+    def _(self, request: DragonUpdateStatusRequest) -> DragonUpdateStatusResponse:
+        updated_statuses: t.Dict[str, t.Tuple[str, t.Optional[t.List[int]]]] = {}
```
ankona, line 105: Would it improve readability to reduce our nesting here?

```py

running_items = [item for item in request.step_ids if item.status == "running"]
groups = [item for item in running_items if item.step_id in self._proc_groups]

for running_group in groups:
  updated_statuses[running_group.step_id] = (STATUS_RUNNING, return_codes)

...
```

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+                                Process(None, ident=puid).returncode
+                                for puid in proc_group_tuple[1]
+                            ]
+                        except (ValueError, TypeError):
+                            return_codes = [-1 for _ in proc_group_tuple[1]]
+                    else:
+                        return_codes = [0]
+                    status = (
+                        STATUS_FAILED
+                        if any(return_codes) or proc_group_tuple[0].status == "Error"
```
ankona, line 125: can we avoid case-sensitive exact string matches to be more resilient to changes in the proc_group_tuple content? Maybe using a status enum here `proc_group_tuple[0].status == STATUS_ERROR`?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -0,0 +1,155 @@
...
+
+        return DragonUpdateStatusResponse(statuses=updated_statuses)
+
+    @process_request.register
+    def _(self, request: DragonStopRequest) -> DragonStopResponse:
+        if request.step_id in self._proc_groups:
+            # Technically we could just terminate, but what if
+            # the application intercepts that and ignores it?
+            proc_group = self._proc_groups[request.step_id][0]
+            if proc_group.status == "Running":
```
ankona, line 140: Ditto on case-sensistive, repeated comparisons to `"Running"`

smartsim/_core/launcher/dragon/dragonParser.py
```diff
@@ -0,0 +1,149 @@
...
+
+import typing as t
+from shutil import which
+
+"""
+Parsers for various slurm functions.
+"""
+
+
+def parse_salloc(output: str) -> t.Optional[str]:
```
ankona, line 35: Can we avoid re-creating what the slurm launcher already does?

smartsim/_core/schemas/dragonRequests.py
```diff
@@ -0,0 +1,86 @@
...
+    name: t.Optional[NonEmptyStr]
+    pmi_enabled: bool = True
+
+
+@request_serializer.register("run")
+class DragonRunRequest(DragonRunRequestView):
+    current_env: t.Dict[str, t.Optional[str]] = {}
+
+    def __str__(self) -> str:
+        return str(DragonRunRequestView.parse_obj(self.dict(exclude={"current_env"})))
```
ankona, line 63: This seems like a lot of work to get to a built-in!

Could we just use:

`return self.model_dump_json(exclude={"current_env"})`

smartsim/_core/schemas/dragonResponses.py
```diff
@@ -0,0 +1,69 @@
...
+from pydantic import BaseModel
+
+import smartsim._core.schemas.utils as _utils
+from smartsim._core.schemas.types import NonEmptyStr
+
+# Black and Pylint disagree about where to put the `...`
+# pylint: disable=multiple-statements
+
+
+class DragonResponse(BaseModel):
```
ankona, line 38: Should we consider separating the schemas for the bootstrap process? That request isn't really related to dragon, it's a SmartSim _control_ event about a resource being set up. I could see a similar event being raised to indicate that any other `SmartSimEntity` has been created...

smartsim/_core/schemas/types.py
```diff
@@ -0,0 +1,31 @@
...
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import pydantic
+
+
+class NonEmptyStr(pydantic.ConstrainedStr):
+    min_length = 1
```
ankona, line 31: I'm not super against this, but it does force us into an inheritance style relationship when we could be more flexible and avoid "TwoCharString" and "FiveCharString" later...

```
from pydantic import BaseModel, Field


class Foo(BaseModel):
    short: str = Field(min_length=3)
```

smartsim/_core/schemas/utils.py
```diff
@@ -0,0 +1,55 @@
+import json
+import typing as t
+
+import pydantic
+
+_KeyT = t.TypeVar("_KeyT")
+_SchemaT = t.TypeVar("_SchemaT", bound=pydantic.BaseModel)
+
+
+class SchemaSerializer(t.Generic[_KeyT, _SchemaT]):
```
ankona, line 10: I'm convinced we don't need this

smartsim/_core/utils/helpers.py
```diff
@@ -302,3 +306,43 @@ def decode_cmd(encoded_cmd: str) -> t.List[str]:
...
+                  .then(calls)
+                  .get_result())
+
+    without the need to introduce confusing temporary variable names
+    """
+
+    _val: _T
+
+    def then(self, fn: t.Callable[[_T], _U], /) -> "_Pipeline[_U]":
+        return _Pipeline(fn(self._val))
```
ankona, line 341: I am not a huge fan of this. It seems like we start to mess with our logged call stacks and just make life harder.

I can also no longer put breakpoints to debug, either.... Are you sure this is desirable?

smartsim/_core/utils/helpers.py
```diff
@@ -302,3 +306,43 @@ def decode_cmd(encoded_cmd: str) -> t.List[str]:
     cleaned_cmd = decoded_cmd.decode("ascii").split("|")
 
     return cleaned_cmd
+
+
+@t.final
+@dataclass(frozen=True)
+class _Pipeline(t.Generic[_T]):
```
ankona, line 313: The builder pattern normally returns the original object instead of creating a new _Pipeline each time. One bonus is that you have an actual description of the expected operations and potentially lazy execution.

smartsim/_core/utils/network.py
```diff
@@ -86,3 +87,13 @@ def current_ip(interface: str = "lo") -> str:  # pragma: no cover
...
+
+
+def get_best_interface_and_address() -> t.Tuple[t.Optional[str], t.Optional[str]]:
+    available_ifs = psutil.net_if_addrs()
+    # TODO make this a CONFIG-time parameter
+    known_ifs = ["hsn", "ipogif", "ib"]
+    for interface in available_ifs:
+        if any(interface.startswith(if_prefix) for if_prefix in known_ifs):
+            return interface, get_ip_from_interface(interface)
+    return None, None
```
ankona, line 99: Could we return an actual object instead of a tuple, or even a named tuple. There is no indication of which position in the resulting tuple is the address and which is the interface in this style.

smartsim/_core/utils/redis.py
```diff
@@ -241,7 +241,8 @@ def shutdown_db_node(host_ip: str, port: int) -> t.Tuple[int, str, str]:  # cov-
 
     if returncode != 0:
         logger.error(out)
-        logger.error(err)
+        if err:
+            logger.error(err)
```
ankona, line 245: if the RC is non-zero, we shouldn't care about the error message to log... that just hides that the error occurred.

i think you might just get rid of `if err` to make sure we don't hide a problem:

`logger.error("something happened. rc: {returncode}, err: {err}")`

smartsim/database/orchestrator.py
```diff
@@ -71,7 +72,7 @@
 def _detect_command(launcher: str) -> str:
     if launcher in by_launcher:
         for cmd in by_launcher[launcher]:
-            if launcher == "local":
+            if launcher in ["local", "dragon"]:
```
ankona, line 75: how sure are we of lowercase inputs?

smartsim/database/orchestrator.py
```diff
@@ -105,10 +106,15 @@ def _check_run_command(launcher: str, run_command: str) -> None:
...
+def _get_single_command(
+    run_command: str, launcher: str, batch: bool, single_cmd: bool
+) -> bool:
     if not single_cmd:
         return single_cmd
 
+    if launcher == "dragon":
+        return False
+
     if run_command == "srun" and getenv("SLURM_HET_SIZE") is not None:
```
ankona, line 118: This test doesn't look like it should live in an orchestrator. Perhaps something in a slurm launcher or a slurm commands module could be imported?

smartsim/database/orchestrator.py
```diff
@@ -847,6 +855,8 @@ def _get_start_script_args(
         ]
         if cluster:
             cmd.append("+cluster")  # is the shard part of a cluster
+        if self.launcher == "dragon":
```
ankona, line 858: i see a LOT of `if launcher == "dragon"` repeats here... can we pull the things that deal w/dragon into a single location? Maybe a `DragonStep` would have this line, for example, and we'd do `step.get_cmd()` 

As-is, we've completely coupled an orchestrator script to dragon and this way lies madness!

smartsim/ml/data.py
```diff
@@ -108,12 +108,13 @@ def download(self, client: Client) -> None:
...
         try:
             info_ds = client.get_dataset(self._ds_name)
-        except RedisReplyError:
+        except RedisReplyError as e:
             # If the info was not published, proceed with default parameters
             logger.warning(
                 "Could not retrieve data for DataInfo object, the following "
                 "values will be kept."
             )
+            logger.debug(f"Original error from Redis was {e}")
```
ankona, line 117: I don't think the original error should be a debug level message. An exception should always be logged, shouldn't it?

smartsim/settings/dragonRunSettings.py
```diff
@@ -0,0 +1,208 @@
...
+            exe_args,
+            run_command="",
+            run_args=run_args,
+            env_vars=env_vars,
+            **kwargs,
+        )
+        self.alloc = alloc
+        self.mpmd: t.List[RunSettings] = []
+
+    reserved_run_args = {"chdir", "D"}
```
ankona, line 78: Should we make this immutable or turn it into a property returning a `Mapping[str, str]`

smartsim/settings/dragonRunSettings.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def set_nodes(self, nodes: int) -> None:
+        """Set the number of nodes
+
+        :param nodes: number of nodes to run with
+        :type nodes: int
+        """
+        self.run_args["nodes"] = int(nodes)
```
ankona, line 86: extraneous `int(xxx)`?

smartsim/settings/dragonRunSettings.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def set_hostlist(self, host_list: t.Union[str, t.List[str]]) -> None:
```
ankona, line 88: It looks like this host list method is copied in multiple places (e.g. AlpsSettings has the only difference being the key `node-list` instead of `nodelist`.

Can we move this up into the base class or pull it out so we don't repeat ourselves?

If we need a different key, consider instead adding to the base class:
```
@abc.abstractmethod
@property
def _nodelist_key(self) -> str:
  ...
```

so all implementations are only the _difference_ and the line in `set_hostlist` can be:

`self.run_args[self._nodelist_key] = ",".join(host_list)`

smartsim/settings/dragonRunSettings.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        self.run_args["time"] = str(walltime)
+
+    def format_env_vars(self) -> t.List[str]:
+        """Build bash compatible environment variable string for Slurm
+
+        :returns: the formatted string of environment variables
+        :rtype: list[str]
+        """
+        return [f"{k}={v}" for k, v in self.env_vars.items()]
```
ankona, line 208: description is not consistent with behavior. it doesn't "build a formatted string" - it builds a list of strings. If we turn around and do `",".join(obj.format_env_vars)`, why not do it here and not return a list?

smartsim/_core/entrypoints/dragon.py
MattToast: The problem with the pydantic serialization is that it does not encode any type information about the original schemas, so if we want the dragon backend to be able to dispatch based on message type, we need some additional entity to provide it (it cannot be an instance attr on the pydantic schema itself, like it was originally, as pydantic will happly override that "defualt" meaning you could end up in a situation where you have a `DragonStartRequest` where `dragon_start_request.msg_type == "dragon_stop_request"`). Under the hood, all the schema serializer is doing is adding a `"__{request,response}_type__"` type key to the dict that pydantic natively dumps out.

The only other way I could think to encode this info would be to set up a req/res channel for each message type, which I could be swayed to, but I would be VERY hesitant to have pydantic encode type information give how passive and coercive it tends to be.

smartsim/settings/settings.py
```diff
@@ -171,7 +173,7 @@ def create_run_settings(
 
     def _detect_command(launcher: str) -> str:
         if launcher in by_launcher:
-            if launcher == "local":
+            if launcher in ["local", "dragon"]:
```
ankona, line 176: we repeated this change in a few places. is it time to get detect command into a shared place?

tests/on_wlm/test_containers_wlm.py
```diff
@@ -49,10 +49,9 @@ def test_singularity_wlm_smartredis(fileutils, test_dir, wlmutils):
     """
 
     launcher = wlmutils.get_test_launcher()
-    print(launcher)
-    if launcher not in ["pbs", "slurm"]:
+    if launcher not in ["pbs", "slurm", "dragon"]:
         pytest.skip(
-            f"Test only runs on systems with PBS or Slurm as WLM. Current launcher: {launcher}"
+            f"Test only runs on systems with PBS, Dragon, or Slurm as WLM. Current launcher: {launcher}"
```
ankona, line 54: how about making things that are annoyingly easy to miss unmissable!

```
supported_launchers = ",".join([key.capitalize() for key in by_launcher.keys()])
f"Test only runs on ...{supported_launchers}. you supplied: {launcher}
```

smartsim/_core/entrypoints/dragon.py
ankona: then the schema is missing data it needs. 

we should add something like:

```
@computed_field
def message_type(self) -> str:
  return type(self).__name__
```

smartsim/_core/utils/helpers.py
MattToast: That's a fair point! I made this to try and avoid the need for so many intermediate variables to the nature of
```py
request = ...
request_str = str(request)
request_schema = Schma.parse_obj(request_str)
```
and I didn't want to resort to
```py
calls(fn(nested(deeply(very(some(data))))))
```
but I think it ended up being less useful that I originally had hoped. I still like pipe-lining fns in but am willing to be out voted if others think its unreadable.

smartsim/_core/utils/helpers.py
MattToast: I was actually trying to explicitly avoid the builder pattern and was going for something similar to bash's `|` or OCaml's `|>`to keep everything "immutable" (you can see that I do end up using this with fns that have side effects, hahaha).

We can still do lazy execution, we just need to move where the fn gets called to the `get_value` method instead
```py
class _Pipeline[T]:
    val: Callable[[], T]

    @classmethod
    def from(val: T) -> _Pipeline[T]:
        return cls(lambda: val)

    def then(self, fn: Callable[U, V]) -> _Pipeline[U, V]:
        return _Pipeline(lambda: fn(self.get_value()))

    def get_value(self) -> T:
        return self.val()

some_data = ...
p = (_Pipeline.from(some_data)
     .then(these)
     .then(methods)
     .then(wont)
     .then(be)
     .then(evaluated))

# until now
p.get_value()
```

MattToast: But again, totally willing to scrap this whole data structure if we agree that it is unreadable!

smartsim/_core/entrypoints/dragon.py
MattToast: Sure, but unless we want to devolve into `eval(payload["message_type"]).parse_obj(payload)`, we still need a mapping of "payload type strings" to "fns to parse the payload" and at that point I would argue that we just remade the `SchemaSerializer` class

ankona: then make the subclasses implement it:

` 
import abc
import pydantic
import typing as t


class Base(pydantic.BaseModel):

    @pydantic.computed_field
    def message_type(self) -> str:
        return self.message_key()

    @staticmethod
    @abc.abstractmethod
    def message_key() -> str: ...


class Book(Base):
    title: str
    num_pages: int

    @staticmethod
    def message_key() -> str:
        return "book"


class Car(Base):
    make: str
    model: str

    @staticmethod
    def message_key() -> str:
        return "car"


book = Book(title="The shortest title in the entire world", num_pages=1)
car = Car(make="Ford", model="F150")

book_json = book.model_dump_json()
car_json = car.model_dump_json()


print(book_json)
print(car_json)


def deserialize_message(message: str) -> t.Optional[Base]:
    for msg_class in [Book, Car]:
        try:
            item = msg_class.model_validate_json(message)
            return item
        except pydantic.ValidationError:
            ...

    return None


obj1 = deserialize_message(book_json)
obj2 = deserialize_message(car_json)

print(obj1)
print(obj2)

`

ankona: Even easier... If we need to filter messages, we can also just swap to a different zmq socket type. PUB/SUB pairs let us set up an explicit filter...

Or, set up a socket for each message type and be dirt simple?

Plenty of options without rebuilding pydantic deserialization .

smartsim/_core/utils/helpers.py
ankona: Clearly, you are a lisp programmer at heart and should be feared and respected!

smartsim/_core/entrypoints/dragon.py
MattToast: Per offline conversation:

WRT the first example: This works GREAT iff models have distinct sets of attrs w/o default values. However, if two models have the same shape, for e.g.

```py
import abc
import pydantic
import typing as t


class Base(pydantic.BaseModel):

    @pydantic.computed_field
    def message_type(self) -> str:
        return self.message_key()

    @staticmethod
    @abc.abstractmethod
    def message_key() -> str: ...


class Book(Base):
    title: str = "default"   # <-- NOTE: new default values here 
    num_pages: int = 123     #           and here!

    @staticmethod
    def message_key() -> str:
        return "book"


class Car(Base):
    make: str = "car brand"     # <-- NOTE: new default values here 
    model: str = "the new one"  #           and here!

    @staticmethod
    def message_key() -> str:
        return "car"


book = Book(title="The shortest title in the entire world", num_pages=1)
car = Car(make="Ford", model="F150")

book_json = book.model_dump_json()
car_json = car.model_dump_json()


print(book_json)
print(car_json)


def deserialize_message(message: str) -> t.Optional[Base]:
    for msg_class in [Book, Car]:
        try:
            item = msg_class.model_validate_json(message)
            return item
        except pydantic.ValidationError as e:
            print(e)

    return None


obj1 = deserialize_message(book_json)
obj2 = deserialize_message(car_json)

print(type(obj1))
print(obj1)
print('============')
print(type(obj2))
print(obj2)
```
will erroneously print out two books. And right now we have multiple schemas that violate that precondition, which runs into the same problem, which is why I am still VERY hesitant to allow pydantic handle de-serialization alone. I think it is too passive.

---

WRT to option two, I would argue that is functionally equivalent to
```py
str_to_cls_map = {cls.message_type: cls for cls in (Book, Car)}
#                     ^^^^^^^^^^^^ For sake of argument assume that this a class property

def load(payload):
    d = json.loads(message_json)
    return str_to_cls_map[d['message_type']].model_validate_json(d)
```
which I would argue is near identical to `SchemaSerializer.deserialize_from_json`.

----

Currently exploring setting up different ZMQ channels to route specific requests based on type and removing the need for the `SchemaSerializer` entirely. Failing that, trying to reduce the scope of `SchemaSerializer` to explicitly work on hydrating requests only (no serialization).

smartsim/_core/schemas/dragonRequests.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class DragonRequest(BaseModel): ...
+
+
+request_serializer = _utils.SchemaSerializer[str, DragonRequest]("request_type")
+
+
+class DragonRunRequestView(DragonRequest):
+    exe: NonEmptyStr
```
ankona, line 45: Could we add our class level docstrings to explain usage? I'm not sure why this guy is a "view" and the others are not

smartsim/_core/schemas/types.py
MattToast: Makes sense to me! I originally had to make this bc python 3.8 didn't support `typing.Annotated` for nested types of multiple `str`s and I didn't want to add a dependency on `typing_extensions` if I could help it. But since dragon also doesn't support 3.8 and because it is so close to EOL, it doesn't really seem relevant anymore!

It also saves us a couple of nasty casts elsewhere! Will remove!!

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: OK, what about a shortened `uuid` and the step id? I prefer seeing the step id as it gives an idea of how many steps were run, but I can also understand the benefit of something not guessable.

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: Yes, the idea is that `request.env` is what the user *asked for*, while `request.current_env` is the env seen by the launcher. Without this, I think it could not find details about the Slurm job.

smartsim/_core/utils/network.py
al-rigazzi: Yeah, I felt like it was half baked. Named tuple it is.

smartsim/_core/entrypoints/dragon.py
al-rigazzi: Done as part of https://github.com/al-rigazzi/SmartSim/pull/7

smartsim/_core/entrypoints/dragon.py
al-rigazzi: I believe you guys reached a conclusion, for which a PR is up already, please feel free to resolve this conversation if it is correct.

smartsim/_core/entrypoints/dragon.py
al-rigazzi: Added a generic `except Exception` and returned a `os.EX_SOFTWARE` code.

smartsim/_core/entrypoints/redis.py
al-rigazzi: Done in https://github.com/al-rigazzi/SmartSim/pull/7

smartsim/_core/entrypoints/telemetrymonitor.py
al-rigazzi: Yes, but demoted to `debug` level.

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: Done as part of https://github.com/al-rigazzi/SmartSim/pull/7

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: OK, done following the Dragon standards, as part of https://github.com/al-rigazzi/SmartSim/pull/7.

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: Ditto as part of https://github.com/al-rigazzi/SmartSim/pull/7.

smartsim/_core/launcher/dragon/dragonParser.py
al-rigazzi: This file has not reason to exist. Deleted as part of https://github.com/al-rigazzi/SmartSim/pull/7.

smartsim/_core/schemas/dragonRequests.py
al-rigazzi: I'm afraid `model_dump_json` is part of pydantic 2.X, but we are constrained to 1.X because of TF (iirc)

smartsim/_core/schemas/dragonResponses.py
al-rigazzi: I added a field about the `pid` which is strictly a dragon thing (allows us to shut it down), so I'm going to push back for a little, and we can think about this later if we believe we can generalize it, which I find an intriguing idea.

smartsim/_core/utils/redis.py
al-rigazzi: Thanks, done as part of the other PR.

smartsim/database/orchestrator.py
al-rigazzi: So, this very check will go away once Dragon process can redirect output to a file. The one above (`single_cmd`) is actually something that will go away once we implement MPMD call for Dragon (which should not be difficult at all, just need to understand how to chain calls or sth similar). May I ask you to postpone this correct criticism to the final review?

smartsim/ml/data.py
al-rigazzi: Promoted to `error`.

smartsim/settings/dragonRunSettings.py
al-rigazzi: That is a good catch for other `RunSettings`-derived classes. Here we can just remove it.

smartsim/settings/dragonRunSettings.py
al-rigazzi: I think we don't need it like that in this class, as we don't pass things as strings. Can you open a ticket for the refactor for other `RunSettings`-derived classes?

smartsim/settings/dragonRunSettings.py
al-rigazzi: Removed, but I guess your comment applies to `SrunSettings`.

smartsim/_core/launcher/dragon/dragonBackend.py
ankona: Ok, super. I wasn't sure - in my head "current" sounded like it was a modified version so this was just me not knowing how it worked.

smartsim/_core/entrypoints/dragon.py
ankona: Upon review, it looks like we already pulled the filename int `Config` but just missed updating this reference.

smartsim/_core/utils/helpers.py
MattToast: `_Pipeline` has been removed with https://github.com/CrayLabs/SmartSim/pull/470/commits/97292df741aba1acb644f39312abfb44380cb556 🎉

smartsim/_core/schemas/dragonRequests.py
MattToast: iirc, we could, in theory, do something similar with `return self.json(exclude={'current_env'})`?

PR #471, opened 2024-01-30
Validate Slurm Timing format

amandarichardsonn:
This PR merges in functionality to validate the timing format when requesting a slurm allocation. Previously, no check was required leading to the WLM responsibility to throw an error. With the new code, SmartSim will catch and throw.

### DIFF
--- a/smartsim/settings/slurmSettings.py
+++ b/smartsim/settings/slurmSettings.py
@@ -244,7 +244,7 @@ class SrunSettings(RunSettings):
         self.run_args["bcast"] = dest_path
 
     @staticmethod
-    def _fmt_walltime(hours: int, minutes: int, seconds: int) -> str:
+    def fmt_walltime(hours: int, minutes: int, seconds: int) -> str:
         """Convert hours, minutes, and seconds into valid walltime format
 
         Converts time to format HH:MM:SS
--- a/smartsim/settings/slurmSettings.py
+++ b/smartsim/settings/slurmSettings.py
@@ -244,7 +244,7 @@ class SrunSettings(RunSettings):
         self.run_args["bcast"] = dest_path
 
     @staticmethod
-    def fmt_walltime(hours: int, minutes: int, seconds: int) -> str:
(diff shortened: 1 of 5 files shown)


smartsim/wlm/slurm.py
```diff
@@ -273,6 +274,29 @@ def _get_alloc_cmd(
...
+        hours, minutes, seconds = map(int, time.split(":"))
+    except ValueError as e:
+        raise ValueError(
+            "Input time must be formatted as `HH:MM:SS` with valid Integers."
+        ) from e
+    delta = datetime.timedelta(hours=hours, minutes=minutes, seconds=seconds)
+    fmt_str = str(delta)
+    if delta.seconds // 3600 < 10:
+        fmt_str = "0" + fmt_str
+    return fmt_str
```
MattToast, line 297: Thoughts on breaking these lines out into a helper function that can be used both here and in `smartsim.settings.slurmSettings.SrunSettings._fmt_walltime`? That way if we have to update how to format the walltime for slurm we only need to touch one spot.

MattToast (changes requested):
One small question/suggestion! Otherwise, LGTM!!

smartsim/wlm/slurm.py
amandarichardsonn: hm hm `_fmt_walltime ` in `smartsim.settings.slurmSettings.SrunSettings` is never used, so Ill add it to `set_walltime` and break it out into a helper func, I was hesitant on that at first bc I wasnt sure about using helper funcs in outside folders but that makes more sense

smartsim/settings/slurmSettings.py
```diff
@@ -256,13 +256,9 @@ def _fmt_walltime(hours: int, minutes: int, seconds: int) -> str:
         :param seconds: number of seconds to run job
         :type seconds: int
         :returns: Formatted walltime
-        :rtype
+        :rtype: str
```
MattToast, line 259: good catch!! Thanks for fix!!

MattToast (approved):
LGTM!! Thanks for making sure our slurm rules stay consistent!!

PR #472, opened 2024-01-31
Add eval() to remove Torch warnings during testing

mellis13:
When running the SmartSim test suite with the Torch backend installed, warnings are surfaced as shown below:

```
tests/backends/test_dbmodel.py::test_pt_db_model
  /Users/torch/jit/_trace.py:1084: TracerWarning: Output nr 1. of the traced function does not match the corresponding output of the Python function. Detailed error:
  Tensor-likes are not close!
  
  Mismatched elements: 10 / 10 (100.0%)
  Greatest absolute difference: 0.0451509952545166 at index (0, 9) (up to 1e-05 allowed)
  Greatest relative difference: 0.019849935548005467 at index (0, 9) (up to 1e-05 allowed)
    _check_trace(
```

This appears to stem from not having the ``eval()`` function invoked before jit tracing and saving the model.  It appears after adding this call that warnings no longer appear.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -45,12 +45,12 @@ Detailed Notes
   all of SmartSim's machine learning backends with Python 3.11.
   (SmartSim-PR451_) (SmartSim-PR461_)
 
-.. _SmartSim-PR472: https://github.com/CrayLabs/SmartSim/pull/472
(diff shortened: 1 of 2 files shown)


ashao (approved):
Good catch @mellis13, thanks for chasing it down.

PR #473, opened 2024-02-02
Use developer log level, protect logger defaults in test

al-rigazzi:
The current `log.get_logger()` function has two issues.

1. it uses `_get_log_level()` to access the value of `SMARTSIM_LOG_LEVEL`, and uses its return value to define whether the logger name should just be SmartSim, or the user-provided one. The problem is that the user-provided one is used only when `_get_log_level()` returns `developer`, but that is never the case, as `developer` gets translated to `debug`.
2. As a consequence of 1, calling `get_logger(name="MY_LOGGER", log_level="INFO")` will always get the logger named `SmartSim` and set its level to info. When this is done in a test, then the log level will be set for all subsequent tests, no matter what the name is set to.

I proposed to replace the `_get_log_level()` with `_translate_log_level()` and use the return value to set the log level when log level is not set to `developer`. To be honest, I think it is kind of weird that users can provide a name for the logger and that it gets thrown away 99.9% of the times, but I guess they could also use a different logger if they wanted something more flexible, and with the changes of this PR, the API is consistent with the docs.

### DIFF
--- a/smartsim/log.py
+++ b/smartsim/log.py
@@ -26,7 +26,6 @@
 
 import functools
 import logging
-import os
 import pathlib
 import sys
 import threading
--- a/smartsim/log.py
+++ b/smartsim/log.py
@@ -63,7 +63,7 @@ if t.TYPE_CHECKING:
     _PR = ParamSpec("_PR")
 
 
-def _get_log_level() -> str:
+def _get_log_level() -> t.Tuple[str, str]:
     """Get the logging level based on environment variable
        SMARTSIM_LOG_LEVEL.  If not set, default to info.
 
@@ -74,20 +74,20 @@ def _get_log_level() -> str:
          - developer: Shows everything happening during execution
                       extremely verbose logging.
 
-    :returns: Log level for coloredlogs
-    :rtype: str
+    :returns: Log level for coloredlogs and value of SMARTSIM_LOG_LEVEL
+    :rtype: tuple[str,str]
     """
     log_level = os.environ.get("SMARTSIM_LOG_LEVEL", "info").lower()
     if log_level == "quiet":
-        return "warning"
+        return "warning", log_level
     if log_level == "info":
-        return "info"
+        return "info", log_level
     if log_level == "debug":
-        return "debug"
+        return "debug", log_level
     # extremely verbose logging used internally
     if log_level == "developer":
-        return "debug"
-    return "info"
+        return "debug", log_level
+    return "info", log_level
 
 
 def get_exp_log_paths() -> t.Tuple[t.Optional[pathlib.Path], t.Optional[pathlib.Path]]:
@@ -205,16 +205,17 @@ def get_logger(
     """
     # if name is None, then logger is the root logger
     # if not root logger, get the name of file without prefix.
-    user_log_level = _get_log_level()
-    if user_log_level != "developer":
+    user_log_level, env_log_level = _get_log_level()
+
+    if env_log_level != "developer":
         name = "SmartSim"
 
     logging.setLoggerClass(ContextAwareLogger)
     logger = logging.getLogger(name)
-    if log_level:
-        logger.setLevel(log_level)
-    else:
+    if not log_level:
         log_level = user_log_level
+    if isinstance(log_level, str):
+        logger.setLevel(log_level.upper())
     coloredlogs.install(level=log_level, logger=logger, fmt=fmt, stream=sys.stdout)
     return logger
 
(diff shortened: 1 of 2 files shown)


smartsim/log.py
```diff
@@ -214,7 +213,7 @@ def get_logger(
     if log_level:
         logger.setLevel(log_level)
     else:
-        log_level = user_log_level
+        log_level = _translate_log_level(user_log_level)
     coloredlogs.install(level=log_level, logger=logger, fmt=fmt, stream=sys.stdout)
```
MattToast, line 217: Oh wow! Yeah, now that you pointed out this "psuedo-caching" behavior of loggers and log levels, this definitely feels wrong. At the least, I really don't like how we call `logger.setLevel` in only one of these branches.

Definitely might be worth a more in-depth refactor later!

MattToast (changes requested):
Great catch! I have a couple of small nits for you, but otherwise LGTM!!

smartsim/log.py
```diff
@@ -77,15 +77,14 @@ def _get_log_level() -> str:
     :returns: Log level for coloredlogs
     :rtype: str
     """
-    log_level = os.environ.get("SMARTSIM_LOG_LEVEL", "info").lower()
-    if log_level == "quiet":
+    if user_log_level == "quiet":
```
ankona, line 80: can we shorten this to:

```py
if user_log_level in ['quiet', 'info', 'debug','warning']:
    return user_log_level

if user_log_level == "developer":
    return "info"
```

smartsim/log.py
```diff
@@ -205,7 +204,7 @@ def get_logger(
     """
     # if name is None, then logger is the root logger
     # if not root logger, get the name of file without prefix.
-    user_log_level = _get_log_level()
+    user_log_level = os.environ.get("SMARTSIM_LOG_LEVEL", "info")
```
ankona, line 207: I wonder what you & @MattToast think about trying to move all `os.environ['xyz']` calls into the `smartsim._core.config.Config`

tests/test_logs.py
```diff
@@ -109,14 +109,16 @@ def test_add_exp_loggers(test_dir):
...
     logger = smartsim.log.get_logger("SmartSimTest", "INFO")
     assert isinstance(logger, smartsim.log.ContextAwareLogger)
 
 
-def test_exp_logs(test_dir: str, turn_on_tm):
+def test_exp_logs(test_dir: str, turn_on_tm, monkeypatch):
     """Ensure that experiment loggers are added when context info exists"""
+    monkeypatch.setenv("SMARTSIM_LOG_LEVEL", "developer")
     test_dir = pathlib.Path(test_dir)
     test_dir.mkdir(parents=True, exist_ok=True)
```
ankona, line 123: Consider adding direct unit test of new method:

```py
@pytest.mark.parametrize(
        "input_level,exp_level",
        [
            pytest.param("INFO", "info", id="lowercasing only, INFO"),
            pytest.param("info", "info", id="input back, info"),
            pytest.param("WARNING", "warning", id="lowercasing only, WARNING"),
            pytest.param("warning", "warning", id="input back, warning"),
            pytest.param("QUIET", "warning", id="lowercasing only, QUIET"),
            pytest.param("quiet", "warning", id="translation back, quiet"),
            pytest.param("DEVELOPER", "debug", id="lowercasing only, DEVELOPER"),
            pytest.param("developer", "debug", id="translation back, developer"),
        ]
)
def test_translate_log_level(input_level: str, exp_level: str, turn_on_tm):
    """Ensure the correct logger type is instantiated"""
    translated_level = smartsim.log._translate_log_level(input_level)
    assert exp_level == translated_level
```

smartsim/log.py
al-rigazzi: +1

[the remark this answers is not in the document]
al-rigazzi: Sure, totally right

al-rigazzi: It was already there...

smartsim/log.py
```diff
@@ -63,7 +63,7 @@
     _PR = ParamSpec("_PR")
 
 
-def _get_log_level() -> str:
+def _translate_log_level(user_log_level: t.Optional[str] = "info") -> str:
```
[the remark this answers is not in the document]
al-rigazzi: You know, I don't know. I think it's better as you say, as it is an internal function we should only call one way.

smartsim/log.py
al-rigazzi: I think you missed one case, but I got the point and modified implementation accordingly.

smartsim/log.py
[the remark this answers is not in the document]
al-rigazzi: I put that in the `CONFIG.log_level` property.

MattToast (approved):
LGTM, pending tests!!

PR #475, opened 2024-02-05
Duplicate DBModel/Script prevention

amandarichardsonn:
This PR prevents duplicate ML models and scripts names being added to an Ensemble member if the names exists already.
The checks are performed for `Ensemble.add_ml_model()`, `Ensemble.add_model()`, `Ensemble.add_script()` and `Ensemble.add_function()`.

### DIFF
--- a/smartsim/entity/ensemble.py
+++ b/smartsim/entity/ensemble.py
@@ -422,6 +422,11 @@ class Ensemble(EntityList[Model]):
             inputs=inputs,
             outputs=outputs,
         )
+        dupe = next((db_model.name for ensemble_ml_model in self._db_models if ensemble_ml_model.name == db_model.name), None)
+        if dupe:
+            raise SSUnsupportedError(
+                f'A ML model with name "{db_model.name}" already exists'
+            )
         self._db_models.append(db_model)
         for entity in self.models:
             self._extend_entity_db_models(entity, [db_model])
@@ -471,6 +476,11 @@ class Ensemble(EntityList[Model]):
             devices_per_node=devices_per_node,
             first_device=first_device,
         )
+        dupe = next((db_script.name for ensemble_script in self._db_scripts if ensemble_script.name == db_script.name), None)
+        if dupe:
+            raise SSUnsupportedError(
+                f'A Script with name "{db_script.name}" already exists'
+            )
         self._db_scripts.append(db_script)
         for entity in self.models:
             self._extend_entity_db_scripts(entity, [db_script])
@@ -517,6 +527,11 @@ class Ensemble(EntityList[Model]):
             devices_per_node=devices_per_node,
             first_device=first_device,
         )
+        dupe = next((db_script.name for ensemble_script in self._db_scripts if ensemble_script.name == db_script.name), None)
+        if dupe:
+            raise SSUnsupportedError(
+                f'A Script with name "{db_script.name}" already exists'
+            )
         self._db_scripts.append(db_script)
         for entity in self.models:
             self._extend_entity_db_scripts(entity, [db_script])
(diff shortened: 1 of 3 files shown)


smartsim/entity/ensemble.py
```diff
@@ -523,15 +523,46 @@ def add_function(
...
+        This static method checks if the provided ML model names already exist in
+        the Ensemble. An SSUnsupportedError is raised if any duplicate names are
+        found. Otherwise, it appends the given list of DBModels to the Ensemble.
+
+        :param model: SmartSim Model object.
+        :type model: Model
+        :param db_models: List of DBModels to append to the Ensemble.
+        :type db_models: t.List[DBModel]
+        """
+        entity_db_models = [db_model.name for db_model in model.db_models]
```
MattToast, line 539: nit: can we make this a set so that we can look up if a name is present with the `in` stmt in constant, rather than linear time?
```suggestion
        entity_db_models = {db_model.name for db_model in model.db_models}
```
I don't ever expect this list to be too long, so it _shouldn't reeeaalllyy_ matter, but also doesn't hurt ¯\\\_(ツ)\_/¯

smartsim/entity/ensemble.py
```diff
@@ -523,15 +523,46 @@ def add_function(
...
+        in the Ensemble. An SSUnsupportedError is raised if any duplicate names
+        are found. Otherwise, it appends the given list of DBScripts to the
+        Ensemble.
+
+        :param model: SmartSim Model object.
+        :type model: Model
+        :param db_scripts: List of DBScripts to append to the Ensemble.
+        :type db_scripts: t.List[DBScript]
+        """
         entity_db_scripts = [db_script.name for db_script in model.db_scripts]
```
MattToast, line 562: ditto here
```suggestion
        entity_db_scripts = {db_script.name for db_script in model.db_scripts}
```

tests/backends/test_dbmodel.py
```diff
@@ -859,3 +859,83 @@ def test_inconsistent_params_db_model():
...
+    exp = Experiment(exp_name, exp_path=test_dir, launcher=test_launcher)
+
+    # Create RunSettings
+    run_settings = exp.create_run_settings(exe=sys.executable, exe_args=test_script)
+    run_settings.set_nodes(1)
+    run_settings.set_tasks(1)
+
+    # Create ensemble
+    smartsim_ensemble = exp.create_ensemble(
+        "smartsim_model", run_settings=run_settings, replicas=2
```
MattToast, line 890: very pedantic name change request
```suggestion
        "smartsim_ensemble", run_settings=run_settings, replicas=2
```

tests/backends/test_dbmodel.py
```diff
@@ -859,3 +859,83 @@ def test_inconsistent_params_db_model():
...
+        devices_per_node=test_num_gpus,
+        first_device=0,
+        inputs=inputs2,
+        outputs=outputs2,
+    )
+
+    # Attempt to add a duplicate ML model to Ensemble via Ensemble.add_model()
+    with pytest.raises(SSUnsupportedError) as ex:
+        smartsim_ensemble.add_model(smartsim_model)
+    assert ex.value.args[0] == 'An ML Model with name "cnn" already exists'
```
MattToast, line 941: Oooh! Nice test for this corner case!!

MattToast (approved):
Couple of small change requests but nothing worth holding up approval over! LGTM!!

smartsim/entity/ensemble.py
ankona: I think this will build the whole set no matter what and the set is better but still not a huge improvement. Given that, could we restate more simply:

```py
dupes = next(x.name for x in model.db_models if x.name == input_name, None)
if dupes:
   raise SSUnsupported
```

PR #479, opened 2024-02-07
Update instructions for Apple Silicon

ashao:
[docs.tgz](https://github.com/CrayLabs/SmartSim/files/14201965/docs.tgz)

Separate instructions are now included for Mac OSX on Apple Silicon. The most important: 
- `core.autoclrf=true` when cloning SmartSim to avoid errors when building due to line-ending conventions during the Redis build
- Only PyTorch is supported for that platform

### DIFF
--- a/doc/installation_instructions/basic.rst
+++ b/doc/installation_instructions/basic.rst
@@ -41,7 +41,7 @@ The machine-learning backends have additional requirements in order to
 use GPUs for inference
 
   - `CUDA Toolkit 11 (tested with 11.8) <https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html>`_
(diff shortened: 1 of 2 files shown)


ankona (approved):
Other than OSX vs MacOS, LGTM

doc/installation_instructions/basic.rst
```diff
@@ -270,9 +276,25 @@ Install SmartSim from Source
...
+    .. code-block:: bash
+
+      git clone https://github.com/CrayLabs/SmartSim smartsim
+
+  .. tab:: MacOS (Apple Silicon)
 
-  git clone https://github.com/CrayLabs/SmartSim smartsim
+    .. code-block:: bash
+
+      git clone --config core.autocrlf=true https://github.com/CrayLabs/SmartSim smartsim
```
ankona, line 297: I think I may have also beeen forced to ensure that non-mac gcc came before mac version in path. must verify...

PR #480, opened 2024-02-07
Duplicate entity name prevention

amandarichardsonn:
This PR prevents the launch of entities with names that already exist in the Experiment.

### DIFF
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -511,15 +511,15 @@ class Controller:
         :type entity: SmartSimEntity
         :raises SmartSimError: if launch fails
         """
-        try:
-            if (
-                entity.name in self._jobs.completed
-                or self._jobs.jobs
-                or self._jobs.db_jobs
+        if (
+            entity.name in self._jobs.completed
+            or entity.name in self._jobs.jobs
+            or entity.name in self._jobs.db_jobs
             ):
-                raise SSUnsupportedError(
-                    "SmartSim entities cannot have duplicate names."
-                )
+            raise SSUnsupportedError(
+                "SmartSim entities cannot have duplicate names."
+            )
+        try:
             job_id = self._launcher.run(job_step)
         except LauncherError as e:
             msg = f"An error occurred when launching {entity.name} \n"
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -511,22 +511,38 @@ class Controller:
         :type entity: SmartSimEntity
         :raises SmartSimError: if launch fails
         """
-        if (
-            entity.name in self._jobs.completed
-            or entity.name in self._jobs.jobs
-            or entity.name in self._jobs.db_jobs
-            ):
-            raise SSUnsupportedError(
-                "SmartSim entities cannot have duplicate names."
-            )
-        try:
-            job_id = self._launcher.run(job_step)
-        except LauncherError as e:
-            msg = f"An error occurred when launching {entity.name} \n"
-            msg += "Check error and output files for details.\n"
-            msg += f"{entity}"
-            logger.error(msg)
-            raise SmartSimError(f"Job step {entity.name} failed to launch") from e
+        # attempt to retrieve entity name in JobManager.completed
+        completed_job = self._jobs.completed.get(entity.name, None)
+
+        # if completed job DNE and is the entity name is not
+        # running in JobManager.jobs or JobManager.db_jobs,
+        # launch the job
+        if completed_job is None and (
+            entity.name not in self._jobs.jobs and entity.name not in self._jobs.db_jobs
+        ):
+            try:
+                job_id = self._launcher.run(job_step)
+            except LauncherError as e:
+                msg = f"An error occurred when launching {entity.name} \n"
+                msg += "Check error and output files for details.\n"
+                msg += f"{entity}"
+                logger.error(msg)
+                raise SmartSimError(f"Job step {entity.name} failed to launch") from e
+        # if the completed job does exist and the entity passed in is the same
+        # that has ran and completed, relaunch the entity.
+        elif completed_job is not None and completed_job.entity is entity:
+            try:
+                job_id = self._launcher.run(job_step)
+            except LauncherError as e:
+                msg = f"An error occurred when launching {entity.name} \n"
+                msg += "Check error and output files for details.\n"
+                msg += f"{entity}"
+                logger.error(msg)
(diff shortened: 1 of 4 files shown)


smartsim/_core/control/controller.py
```diff
@@ -512,6 +512,14 @@ def _launch_step(
...
         try:
+            if (
+                entity.name in self._jobs.completed
+                or self._jobs.jobs
+                or self._jobs.db_jobs
+            ):
+                raise SSUnsupportedError(
+                    "SmartSim entities cannot have duplicate names."
+                )
             job_id = self._launcher.run(job_step)
```
MattToast, line 523: Can we move this out the `try` block, just to keep the area from where we expect errors to be raised as small as possible?
```suggestion
        if (
            entity.name in self._jobs.completed
            or self._jobs.jobs
            or self._jobs.db_jobs
        ):
            raise SSUnsupportedError(
                "SmartSim entities cannot have duplicate names."
            )
        try:
            job_id = self._launcher.run(job_step)
```

smartsim/_core/control/controller.py
```diff
@@ -512,6 +512,14 @@ def _launch_step(
         :raises SmartSimError: if launch fails
         """
         try:
+            if (
+                entity.name in self._jobs.completed
+                or self._jobs.jobs
+                or self._jobs.db_jobs
```
MattToast, line 518: I'm guessing this is not going to do what you intended: This clause will always eval `True` if `self._jobs.jobs` or `self._jobs.db_jobs` is a non-empty `dict[str, Job]`.

Maybe
```suggestion
                entity.name in self._jobs.completed
                or entity.name in self._jobs.jobs
                or entity.name in self._jobs.db_jobs
```
?

smartsim/_core/control/controller.py
```diff
@@ -512,6 +512,14 @@ def _launch_step(
         :raises SmartSimError: if launch fails
         """
         try:
+            if (
+                entity.name in self._jobs.completed
```
MattToast, line 516: Are we sure about this check? Won't this prevent relaunching a completed model?
```py
exp = Experiment(...)
rs = exp.create_run_settings(...)
model = exp.create_model(run_settings=rs, ...)

exp.start(model, block=True)  # Very explicitly wait for the model to complete
exp.start(model, block=True)  # <-- I do not think we want this to error

```

tests/test_controller_errors.py
```diff
@@ -108,3 +114,41 @@ def test_bad_orc_checkpoint():
     cont = Controller(launcher="local")
     with pytest.raises(FileNotFoundError):
         cont.reload_saved_db(checkpoint)
+
+
+@pytest.mark.parametrize(
+    "collection,duplicate_name,completed",
```
MattToast, line 120: Very pedantic: can we change the name of `collection` to something a bit more representative? Maybe `entity` or `launchable`?

MattToast (changes requested):
Gave a quick skim: just a couple of small requests and some threads to pull on while you work out the CI

smartsim/_core/control/controller.py
ankona: I think I pushed this idea because running the same model would overwrite outputs from the first run. Do we care about the outputs?

MattToast: It's not super common for anything I write, but I know there have been use cases were users run a model, ignore the outputs or programmatically move them somewhere else, re-generate the model dir, and then re-run the same model.

Another more common case (at least for me) is, when a user is working in a jupyter notebook, they might start up an orchestrator, stop it for some reason, but then decide to relaunch the same orc object without restarting the kernel or creating a new experiment.

I'd be hesitant to make this change just because this is behavior that we have billed as a feature in the past (see `tests/on_wlm/test_restart.py`) and would constitute an pretty significant API break.

Which isn't to say that I think the API itself isn't somewhat incredibly convoluted, especially now that we are specifically trying to enforce a name uniqueness requirement that was in the past, at best, a suggestion. Unfortunately, that does mean that a user could start another model of the same name _if_ they let the first model finsish. e.g.
```py
exp = Experiment("multi-start")
rs = exp.create_run_settings("echo", ["this", "just", "works"])
model_1 = exp.create_model("my-model")
model_2 = exp.create_model("my-model")  # <-- note the same name

exp.start(model_1, block=True)
exp.start(model_2, block=True)  # <-- This, unfortunately, now just works
                                #     which, I would argue, is partially what
                                #     we are trying to discourage
```

MattToast: I suppose we could check to see if the user is trying to start the same object with a
```py
if not (entity.name in self._jobs.completed
        and entity is self._jobs.completed[entity.name]):
   ...  # TODO: block the launch
```
condition, but I think the same problem exists as a model's run settings is a writable attribute. So a user could effectively just launch the same model, with the same name, with a completely different set of run settings, e.g.
```py
exp = Experiment("multi-start")

rs_1 = exp.create_run_settings("echo", ["my", "model"])
model = exp.create_model("my-model")
exp.start(model, block=True)

rs_2 = exp.create_run_settings("echo", ["my", "completely", "different", "model"])
model.run_settings = rs_2
exp.start(model)  # <-- Effectively just ran two different models
                  #     with the same name
```
somewhat defeating the point of the check.

For that reason, I think we should scope this PR to "preventing duplicate _simultaneously running_ entity names", and decide how (or even _if_) we want to enforce a unique names constraint in a later spike. Does that seem like a good scope @amandarichardsonn @ankona?

amandarichardsonn: I have the current implementation below that will allow users to restart a completed entity but will not allow users to run an entity of the same name that is completed or running. I think this is a bandaid patch and should be revisited. 

```
completed_job = self._jobs.completed.get(entity_passed_in.name, None)

if completed_job is None and is not a running job:
    launch
elif completed_job is not None and completed_job.entity is entity_passed_in:
    launch
else:
    raise
```

However, with this change above, users are now **required** to specify a db_identifier var if launching multiple Orchestrators. The implementation caught a SS test that launched two orchs with no db_identifiers, and therefore threw on same name. Do we want this?

amandarichardsonn: I say if this is more complex than the implementation that I currently have - we scope down to what you mentioned @MattToast. Is this something we should discuss with the team as far as allowing users to overwrite entity info by relaunching before merging in? and therefore should rescope

MattToast (approved):
One last minor nit,  but nothing worth holding approval up over! LGTM!!

PR #489, opened 2024-02-15
Bug fix in model template

juliaputko:


smartsim/_core/control/controller.py
```diff
@@ -135,6 +135,10 @@ def start(
             # it may be called seperately
             self.poll(5, True, kill_on_interrupt=kill_on_interrupt)
 
+    @property
+    def active_orch_list(self) -> t.Dict[str, Job]:
+        return self._jobs.db_jobs
```
ankona, line 140: I think we might introduce a way for external code to mess with your orchestrator list here by returning `db_jobs` directly.

Can you write a test that uses this property, removes something from the list, and shows that `self._jobs.db_jobs` remains unchanged? 

Probably solvable w/something as simple as `return {}.update(self._jobs.db_jobs)`

smartsim/_core/control/previewrenderer.py
```diff
@@ -100,6 +106,17 @@ def as_toggle(_eval_ctx: u.F, value: bool) -> str:
     return "On" if value else "Off"
 
 
+@pass_eval_context
+def get_ifname(_eval_ctx: u.F, value: t.List[str]) -> str:
+    return next((item for item in value if "ifname" in item), str).split("=")[-1]
```
ankona, line 111: I think this will blow up if the filter clause fails. I think I either don't understand it, or the default `str` value could be a problem. Won't that pass the str _type_ and not an empty string?

I'd like to see tests for these methods added, w/edge case checks:
- `get_ifname`
- `get_dbtype `

smartsim/_core/control/previewrenderer.py
```diff
@@ -100,6 +106,17 @@ def as_toggle(_eval_ctx: u.F, value: bool) -> str:
...
+
+
+@pass_eval_context
+def get_dbtype(_eval_ctx: u.F, value: str) -> t.Any:
+    db_type, _ = value.split("/")[-1].split("-", 1)
```
ankona, line 116: ditto here. if `value == ""` the splits are going to blow up?

templates/templates/preview/model_plain_text.template
```diff
@@ -1,107 +1,97 @@
...
+        {% endif %}
+        {%- if model.files %}
+        {%- if model.files.tagged %}
+        Tagged Files for Model Configuration: 
+          {%- for tagged in model.files.tagged %}
+          {{ tagged }} 
+          -> {{ model.path }}
+          {%- endfor %}
+          {%- endif -%}
+          {%- if model.files.copy %}
```
ankona, line 27: super nitpicky, but if you have extra time (of course), consider fixing indentation so the `{%- if model.files.copy %}` is at the same level as the thing it's controlling (`Copy Files:...`). I think that might improve the readability.

Second, Consider the repetition in the tagged/copy/symlink sections here as a potential sub-template... You could pass the title and list source as args avoid having to change multiple places if this needs new formatting.

smartsim/_core/control/previewrenderer.py
ankona: Try something _direct_, like:

```
tpl = jinja.template("foo | get_ifname")
rendered_preview = tpl.render(
    foo=None
)
```

ankona (changes requested):
a couple minor issues here but otherwise good. only thing that looks potentially damaging is the db_jobs dict being exposed. 

The filters should definitely have tests that verify them with simple string templates that only verify their output...

smartsim/_core/control/controller.py
juliaputko: changed to return {**self._jobs.db_jobs}, and added test

smartsim/_core/control/previewrenderer.py
juliaputko: what edge cases should I check for?

smartsim/_core/control/previewrenderer.py
juliaputko: added check for empty string

templates/templates/preview/model_plain_text.template
juliaputko: fixed spacing

smartsim/_core/control/previewrenderer.py
ankona: strings missing an equal, strings with just equal, strings with LHS=, strings w/=RHS, None

PR #493, opened 2024-02-16
Hot Fix: bump dependency version

MattToast:
A critical performance concern was identified and addressed in SmartRedis. A patch fix was deployed, and SmartSim was updated to ensure users do not inadvertently pull the unpatched version of SmartRedis.


ashao (approved):
Looks good!

PR #496, opened 2024-02-16
Update tutorial containers

al-rigazzi:
Currently, the docker container created with `make tutorials-prod` does not work, because the `networkx` version which is pulled is not compatible with Python 3.8 (the default version in Ubuntu 20.04). This PR bumps up the Python version used when building the tutorial containers (both production and development).

A minor bug in the tutorial container version number is also addressed as part of this PR.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       - "8888:8888"
 
   tutorials-prod:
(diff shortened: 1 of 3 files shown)


MattToast (approved):
Good catch! thanks for making sure these stay up to date!!

PR #502, opened 2024-02-26
bump redis dep to 7.2.4

ankona:
A fix in the build scripts of redis 7.2.4 modifies build behavior on MacOS on Apple Silicon. The change fixes an issue where incorrect compiler flags are defined and result in build failures due to the `redis_fstat` macro


ashao (approved):
Looks good

PR #505, opened 2024-02-27
Add Verbosity Parameters to Preview method

juliaputko:
https://github.com/CrayLabs/Poseidon/issues/605


smartsim/_core/control/previewrenderer.py
```diff
@@ -74,14 +74,15 @@ def render(
...
-    verbosity_level = _check_verbosity_level(verbosity_level)
+    verbosity_level = _check_verbosity_level(Verbosity(verbosity_level))
 
     loader = jinja2.PackageLoader("templates")
     env = jinja2.Environment(loader=loader, autoescape=True)
 
     env.filters["as_toggle"] = as_toggle
     env.filters["get_ifname"] = get_ifname
     env.filters["get_dbtype"] = get_dbtype
+    env.globals["Verbosity"] = Verbosity
```
ankona, line 85: Are you sure you want to set this value to the _class_ Verbosity and not the parameter `verbosity_level`? I think this may not hold the value you want.

100% nitpick... can we make the global env name follow the same snake-case naming standard as `as_toggle`, `get_ifname`, and `get_dbtype`?

ankona: I take it back. I had to look @ the template to learn what was happening. I am wrong on both accounts here.

tests/test_preview.py
```diff
@@ -952,6 +1014,200 @@ def test_preview_colocated_db_script_ensemble(fileutils, test_dir, wlmutils, mlu
...
+    # Execute method for template rendering
+    output = previewrenderer.render(exp, preview_manifest, verbosity_level="info")
+
+    # Evaluate output
+    assert "TCP/IP Port(s)" not in output
+    assert "Executable" not in output
+    assert "Run Command" not in output
+
+
+def test_preview_verbosity_info_ensemble(test_dir, wlmutils):
```
ankona, line 1178: Don't forget that you can write real unit tests that test the feature only and don't have to include the whole stack... 

Here, you could test verbosity more directly by using `monkeypatch` to get a path to a smaller template that is simpler, such as:

```
{%- if verbosity_level == Verbosity.INFO %}
INFO!
{%endif%}
{%- if verbosity_level == Verbosity.WARN %}
WARN!
{%endif%}
{%- if verbosity_level == Verbosity.DEBUG %}
DEBUG!
{%endif%}
```

Using that would focus a test only on something that can verify that your use of verbosity was correct and that your env.globals works as expected. And, it's easy to parametrize w/all 3 all input and expected outputs.

Ditto on you

smartsim/_core/control/previewrenderer.py
```diff
@@ -90,6 +91,13 @@ def render(
 
     tpl = env.get_template(tpl_path)
 
+    if verbosity_level == Verbosity.INFO:
+        logger.warning(
+            "Only showing user set parameters. Some entity "
+            "fields are truncated. To view truncated fields: use verbosity_"
+            "level 'developer' or 'debug'"
```
ankona, line 98: super nitpick: can you move `fields are` to line 96 so we don't have to rely on concatenation to get `verbosity_level` put back together -> I worry about editing the error message and not paying attention to the concat, and having a nonsense error message.

ankona (approved):
Looks great.

tests/test_preview.py
juliaputko: I like this a lot better

smartsim/_core/control/previewrenderer.py
juliaputko: yup

templates/templates/preview/clientconfig_plain_text_info.template
```diff
@@ -0,0 +1,24 @@
...
+            Type: Standalone
+            {%- endif %}
+        {%- endfor %}
+          {%- if model.query_key_prefixing() %} 
+          Outgoing Key Collision Prevention (Key Prefixing): 
+            Tensors: {{ model.query_key_prefixing() | as_toggle }}
+            Datasets: {{ model.query_key_prefixing() | as_toggle }}
+            ML Models/Torch Scripts: {{ False | as_toggle }}
+            Aggregation Lists: {{ model.query_key_prefixing() | as_toggle }}
+        {%- endif %}
```
ankona, line 24: I really want you to try out `pre-commit` and just start w/a configuration that adds trailing newlines... it will be nice when you never have to hear me ask, won't it!?

```
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.5.0
    hooks:
    -   id: check-yaml
    -   id: end-of-file-fixer
    -   id: trailing-whitespace
```

juliaputko: woops! yep I will add

PR #512, opened 2024-03-06
ReadTheDocs Configuration File

amandarichardsonn:
Adding readthedocs yaml file to develop


doc/conf.py
```diff
@@ -82,6 +82,8 @@
 # a list of builtin themes.
 html_theme = "sphinx_book_theme"
 
+# Avoid indexing by search engines
+html_extra_path = ['../robots.txt']
```
ashao, line 86: This will create a robots.txt file for all documentation that we build (including the one that will be hosted at craylabs.org. I was digging into the documentation a little more (and verified) that readthedocs by default ships a robots.txt: https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html#seo-robots-txt

This seems to be the case for us, since this file exists: https://smartsim.readthedocs.io/en/latest/robots.txt

Could you dig into this a little more? It might also (for safety) be worthwhile to still figure out a way to make a robots.txt file but have a way to only create the robots.txt file if we know that it's being built via readthedocs

Could you dig into this and

ashao (changes requested):
One comment on the robots.txt work

Another item to take care of requires configuring ReadTheDocs:
- Make craylabs.org the canonical URL: https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html#canonical-urls
- Add me, Matt Ellis, and Al as owners of the SmartSim readthedocs account

doc/conf.py
amandarichardsonn: for sure!

amandarichardsonn:
canonical URL added! You can check it [here](https://readthedocs.org/dashboard/smartsim/domains/) (Details).

doc/changelog.rst
```diff
@@ -26,14 +26,14 @@ Description
 Detailed Notes
 
 - Implemented new structure of SmartSim documentation. Added examples
-  images and further detail of SmartSim components.
```
mellis13, line 29: Thanks for fixing the changelog.  Can you also fix the order in the detailed description section.  The last two items are out-of-order compared to the top short descriptions.

doc/changelog.rst
```diff
@@ -26,14 +26,14 @@ Description
 Detailed Notes
 
 - Implemented new structure of SmartSim documentation. Added examples
```
mellis13, line 28: Also, I think we need a changelog entry for this PR

amandarichardsonn: updated!

doc/changelog.rst
amandarichardsonn: fixed

.readthedocs.yaml
```diff
@@ -0,0 +1,41 @@
...
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.10"
+  jobs:
+    post_checkout:
+      # Cancel building pull requests when there aren't changed in the docs directory or YAML file.
```
mellis13, line 11: changed -> changes

.readthedocs.yaml
```diff
@@ -0,0 +1,41 @@
...
+      # You can add any other files or directories that you'd like here as well,
+      # like your docs requirements file, or other files that will change your docs build.
```
mellis13, line 13: like your docs requirements file, or other files that will change your docs build.

->

such as docs requirements file or other files that affect the docs build.

.readthedocs.yaml
```diff
@@ -0,0 +1,41 @@
...
+      #
+      # If there are no changes (git diff exits with 0) we force the command to return with 183.
```
mellis13, line 15: Great comment!  Thanks for explaining this.

mellis13 (approved):
Approving with only minor comments and pending @ashao review

.readthedocs.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - git clone --depth 1 https://github.com/CrayLabs/SmartRedis.git smartredis
+      - git clone --depth 1 https://github.com/CrayLabs/SmartDashboard.git smartdashboard
+    post_create_environment:
+      - python -m pip install .
+      - cd smartredis; python -m pip install .
+      - cd smartredis/doc; doxygen Doxyfile_c; doxygen Doxyfile_cpp; doxygen Doxyfile_fortran
+      - ln -s smartredis/examples ./examples
+      - cd smartdashboard; python -m pip install .
+    pre_build:
+      - python -m sphinx -b linkcheck doc/ $READTHEDOCS_OUTPUT/linkcheck
```
ashao, line 32: Could you make a ticket to update the `make docs` target so that we can eventually converge the two workflows?

doc/conf.py
```diff
@@ -84,6 +84,13 @@
 # a list of builtin themes.
 html_theme = "sphinx_book_theme"
 
+# Check if the environment variable is set to 'True'
+if os.environ.get('READTHEDOCS') == "True":
```
ashao, line 88: Nice solution!

doc/conf.py
```diff
@@ -84,6 +84,13 @@
...
+    # If it is, generate the robots.txt file
+    with open('../robots.txt', 'w') as f:
```
ashao, line 90: Why is this `../` and not `./`?

ashao (approved):
Thanks for your work on this! Some very minor comments that shouldn't hold up merging.

In addition to the `make docs` update ticket that I suggest, I'm going to open up another one about the canonical URL. What it looks like that means is actually that we could update craylabs.org to point to ReadTheDocs instead. That's not happening now, because we don't have our domain provider set up like that, but I think we should think through whether we want to do that.

.readthedocs.yaml
amandarichardsonn: yes!

PR #517, opened 2024-03-12
Correct ExecArgs Handling During RunSetting

amandarichardsonn:
Added `isinstance` check to RunSettings exe_args setter. Added additional tests.


smartsim/settings/base.py
```diff
@@ -448,12 +455,16 @@ def add_exe_args(self, args: t.Union[str, t.List[str]]) -> None:
         :type args: str | list[str]
         :raises TypeError: if exe args are not strings
         """
+
+        if not isinstance(args, (list, str)):
```
mellis13, line 459: Two comments here:

1.  This TypeError message is not really correct given this statement.  
2.  Given that the ``exe_args`` setter method is doing the same check as ``add_exe_args`` should we embed the type check in an internal class method like ``_check_exe_arg_type(args)`` or something similar?  Open to suggestions, but I think the checks in this routine need to be improved (at least group all of the checks at the top of the method and mimic the other method)

smartsim/settings/base.py
```diff
@@ -448,12 +455,16 @@ def add_exe_args(self, args: t.Union[str, t.List[str]]) -> None:
...
+            raise TypeError("All elements in the list should be of type str")
+
```
mellis13, line 461: Should we be using the _build_exe_args() function in this method instead of repeating the logic in that method?

tests/test_run_settings.py
```diff
@@ -208,18 +208,17 @@ def test_add_exe_args_list_of_mixed_lists_init():
...
 
 
 def test_add_exe_args_list_of_str_lists_init():
     """Ensure that list[list[str]] pass validation"""
     exe_args = [["1", "2", "3"], ["4", "5", "6"]]
 
-    settings = RunSettings("python", exe_args=exe_args)
-
-    assert settings.exe_args == exe_args
+    with pytest.raises(TypeError) as type_error:
```
mellis13, line 220: I'm not seeing a functional difference between this test and this previous.  What is it?

tests/test_run_settings.py
```diff
@@ -323,6 +322,13 @@ def test_bad_exe_args_2():
         _ = RunSettings("python", exe_args=exe_args)
 
 
+def test_bad_exe_args_3():
+    """test when user provides incorrect types to exe_args"""
+    exe_args = [["this", "should"], ["not", "work"]]
+    with pytest.raises(TypeError):
```
mellis13, line 328: How does this test of list of list[str] differ from the previous tests?

mellis13 (changes requested):

smartsim/settings/base.py
```diff
@@ -533,25 +526,24 @@ def set(
 
     @staticmethod
     def _build_exe_args(exe_args: t.Optional[t.Union[str, t.List[str]]]) -> t.List[str]:
-        """Convert exe_args input to a desired collection format"""
+        """Check and convert exe_args input to a desired collection format"""
+        if isinstance(exe_args, list):
+            exe_args = copy.deepcopy(exe_args)
+
+        if not (
```
mellis13, line 533: Thanks for the refactor.  I think this is much easier to read.  One minor critique is that I might just have at the beginning of the method:

```python
if not exe_args:
    return []
```

And that at the end of the function you can just return exe_args


Otherwise this is really good

mellis13 (changes requested):

smartsim/settings/base.py
amandarichardsonn: Might as well not go through the checks if empty!

mellis13 (approved):

PR #520, opened 2024-03-13
Remove duplicate launched model names from full test suite

MattToast:
Removes behavior deprecated in #480 from test suite.


mellis13 (approved):
LGTM!

PR #521, opened 2024-03-14
Disallow Uninitialized Variable Use

MattToast:
Configure mypy to error when a potentially uninitialized variable is used. Fix lingering errors found by mypy.


smartsim/_core/entrypoints/telemetrymonitor.py
```diff
@@ -460,6 +460,11 @@ def _to_completed(
 
         if hasattr(job.entity, "status_dir"):
             write_path = pathlib.Path(job.entity.status_dir)
+        else:
```
ankona, line 463: I can move this change into my working telemetry monitor branch if necessary. I don't want to deal with the merge conflict headache... Let me know if you want me to add it after you remove it

smartsim/_core/entrypoints/colocated.py
```diff
@@ -258,6 +254,15 @@ def launch_db_scripts(client: Client, db_scripts: t.List[t.List[str]]) -> None:
         raise SSInternalError("Colocated entrypoint raised an error") from e
 
 
+@contextlib.contextmanager
+def ephemeral_db_client(db_id: str) -> t.Generator[Client, None, None]:
+    """Create an context that will yield a ``smartredis.Client`` created from the
+    environment that will reclaim the client resources upon exit.
+    """
+    options = ConfigOptions.create_from_environment(db_id)
+    yield Client(options, logger_name="SmartSim")
```
ankona, line 263: None of the changes in this file feel appropriate for a PR titled "disallow unitialized variable use"

smartsim/_core/_cli/validate.py
```diff
@@ -112,10 +113,11 @@ def execute(
     except Exception as e:
         logger.error(
```
ankona, line 114: I look at these black reformats and wonder if anybody dislikes them as much as me. I'm a fan of:

```py
msg = f"Experiment failed due to the following exception:\n{e}"
logger.error(msg, exc_info=True)
```

just to avoid turning 2 lines into 4

smartsim/_core/entrypoints/telemetrymonitor.py
MattToast: Oh, good point! I don't want to remove this change as it will result in a failing CI, but I also agree that dealing with the merge conflict in #460 is probably a lot to add to an already big PR. Let's prioritize #460 getting merged and bring this in after it.

smartsim/_core/entrypoints/colocated.py
MattToast: These changes are explicitly necessary to remove the `del` of the potentially uninitialized `client` var. These changes are effectively the same as
```py
@contextlib.contextmanger
def ephemeral_db_client(db_id: str) -> t.Generator[Client, None, None]:
    options = ConfigOptions.create_from_environment(db_id)
    client = Client(options, logger_name="SmartSim")
    yield client
    del client
```
to match the intent and behavior of the original code. But the `del` is implied by exiting the stack frame and at that point the `client` var becomes vestigial, so I decided to remove it entirely.

I made these changes as it was the smallest footprint I could come up with to fix the error, but let me know if you would like to see more added/removed from this solution (to the best of my knowledge there is no `close` method on a `smartredis.Client` object).

MattToast: Looking back I guess since we don't actually recover in the `except` case we could just move the `del` to an `else` block instead of a `finally` block, which would be an even smaller footprint! Will revert to that!!

smartsim/_core/_cli/validate.py
MattToast: Hahaha a fair point, but in order to maintain the concatenation with the rest of the string and stay within the 88 char limit, we would have to result to a
```py
msg = "SmartSim failed to run a simple experiment!\n"
msg += f"Experiment failed due to the following exception:\n{e}"
logger.error(msg, exc_info=True)
```
And I hate the idea of creating a buffer and then immediately needing to create another second buffer and copy the contents of the original buffer into the new one to save ~1 line, which why I tend to prefer the black styling for long strings.

ankona (approved):
lgtm

PR #522, opened 2024-03-15
Application executes before colocated Orchestrator is created

amandarichardsonn:
Orchestrator launch moved to a blocking process. Application executes once Orchestrator is built.


smartsim/_core/launcher/colocated.py
```diff
@@ -56,6 +56,8 @@ def write_colocated_launch_script(
         script_file.write("set -e\n\n")
 
         script_file.write("Cleanup () {\n")
+        script_file.write("echo error detected cleanup triggered\n")
```
mellis13, line 59: I think this message is not strictly true because the ``cleanup()`` method is triggered on any exit (including non-error exits.  We should at least clean this message up but also I'm curious if the group thinks we should always be printing this with the process ID. @ashao @al-rigazzi

smartsim/_core/launcher/colocated.py
```diff
@@ -67,9 +69,12 @@ def write_colocated_launch_script(
...
         if colocated_settings["debug"]:
             script_file.write("export SMARTSIM_LOG_LEVEL=debug\n")
-
-        script_file.write(f"{colocated_cmd}\n")
-        script_file.write("DBPID=$!\n\n")
+        script_file.write(f"db_stdout=$({colocated_cmd})\n")
+        # pylint: disable=anomalous-backslash-in-string
+        sed_command = "sed -n 's/.*__PID__\([0-9]*\)__PID__.*/\\1/p'"
+        printable_command = repr(sed_command)
+        print(printable_command)
```
mellis13, line 76: I don't think we always want to be printing the command to stdout

mellis13 (changes requested):
A couple of comments on the first round but want to wait for @ashao and @al-rigazzi feedback

smartsim/_core/entrypoints/colocated.py
```diff
@@ -177,7 +177,8 @@ def main(
     db_scripts: t.List[t.List[str]],
     db_identifier: str,
 ) -> None:
-    global DBPID  # pylint: disable=global-statement
+    # pylint: disable=too-many-statements
+    global DBPID
```
al-rigazzi, line 181: I guess the `pylint` directive was accidentally deleted, but it is needed to make it accept `global` variables.

al-rigazzi (changes requested):
Just one change requested, otherwise looks good!

smartsim/_core/launcher/colocated.py
al-rigazzi: I think it would be good to add the process ID, to make post-mortem analysis of failed colocated orchestrators easier.

smartsim/_core/entrypoints/colocated.py
amandarichardsonn: yes oops!

smartsim/_core/launcher/colocated.py
```diff
@@ -67,9 +67,11 @@ def write_colocated_launch_script(
...
         if colocated_settings["debug"]:
             script_file.write("export SMARTSIM_LOG_LEVEL=debug\n")
-
-        script_file.write(f"{colocated_cmd}\n")
-        script_file.write("DBPID=$!\n\n")
+        script_file.write(f"db_stdout=$({colocated_cmd})\n")
+        # pylint: disable=anomalous-backslash-in-string
+        script_file.write(
+            "DBPID=$(echo $db_stdout | sed -n 's/.*__PID__\([0-9]*\)__PID__.*/\\1/p')\n"
+        )
```
MattToast, line 74: Small thing I saw during stand-up: Consider making this a r-string or explicitly escaping the anomalous backslashes rather than silencing the pylint error:
```suggestion
        script_file.write(
            "DBPID=$(echo $db_stdout | sed -n 's/.*__PID__\\([0-9]*\\)__PID__.*/\\1/p')\n"
        )
```

MattToast (commented):
One small pedantic thing!

doc/changelog.rst
```diff
@@ -30,6 +31,8 @@ Description
 
 Detailed Notes
 
+- Build shell script to execute application after colocated Orchestrator
```
ashao, line 34: Reword this for clarity

doc/changelog.rst
```diff
@@ -18,6 +18,7 @@ To be released at some future point in time
 
 Description
 
+- Colocated Model bug fix
```
ashao, line 21: Make this a little more descriptive

smartsim/_core/entrypoints/colocated.py
```diff
@@ -201,8 +202,10 @@ def main(
     # we generally want to catch all exceptions here as
     # if this process dies, the application will most likely fail
     try:
-        process = psutil.Popen(cmd, stdout=PIPE, stderr=STDOUT)
-        DBPID = process.pid
+        with open("colo_orch_output.txt", "w", encoding="utf-8") as file:
+            process = psutil.Popen(cmd, stdout=file.fileno(), stderr=STDOUT)
+            DBPID = process.pid
+        print(f"__PID__{DBPID}__PID__", flush=True)
```
ashao, line 208: Could you comment this line to explain why we have to print to standard out?

smartsim/_core/entrypoints/colocated.py
```diff
@@ -201,8 +202,10 @@ def main(
     # we generally want to catch all exceptions here as
     # if this process dies, the application will most likely fail
     try:
-        process = psutil.Popen(cmd, stdout=PIPE, stderr=STDOUT)
-        DBPID = process.pid
+        with open("colo_orch_output.txt", "w", encoding="utf-8") as file:
```
ashao, line 205: I kind of think we dump this to `/dev/null` unless we're in debug mode. Otherwise, to help with debugging, it would be useful to include the hostname so that we can separate out the logs from each node. Also, please rename the extension to .log.

ashao (changes requested):
Thanks for tracking this down. Some minor suggestions requested.

smartsim/_core/entrypoints/colocated.py
amandarichardsonn: I was not familiar with dumping to `/dev/null` but I agree with the suggestion!

smartsim/_core/entrypoints/colocated.py
```diff
@@ -201,8 +202,17 @@ def main(
     # we generally want to catch all exceptions here as
     # if this process dies, the application will most likely fail
     try:
-        process = psutil.Popen(cmd, stdout=PIPE, stderr=STDOUT)
-        DBPID = process.pid
+        hostname = socket.gethostname()
+        filename = (
+            f"{hostname}.log"
```
al-rigazzi, line 207: What about `colo_orc_{hostname}.log`? It would help understanding what the file represents.

al-rigazzi (changes requested):
I suggest a small change to the filename, requesting @ashao's opinion on that too, to avoid too many iterations

smartsim/_core/entrypoints/colocated.py
ashao: Agreed

smartsim/_core/entrypoints/colocated.py
```diff
@@ -201,8 +202,17 @@ def main(
...
+            if os.getenv("SMARTSIM_LOG_LEVEL") == "debug"
+            else "/dev/null"
```
al-rigazzi, line 209: I also suggest "/dev/null" -> `os.devnull` see [here](https://docs.python.org/3/library/os.html#os.devnull) as a more portable (and nice looking) solution.

al-rigazzi (approved):
That was quick! LGTM

smartsim/_core/entrypoints/colocated.py
amandarichardsonn: Good catch! would h8 for this to fail for others

PR #528, opened 2024-03-21
Addressing PR comments

juliaputko:
…ssing PR comments


smartsim/_core/control/previewrenderer.py
```diff
@@ -81,14 +86,30 @@ def get_dbtype(_eval_ctx: u.F, value: str) -> str:
...
+    otherwise, to determine how Jinja template should
+    render an item.
+    """
     return isinstance(value, list)
 
 
+def render_to_file(content: str, filename: str) -> None:
+    """
+    Output preview to a file if an output filename
+    is specified.
```
ankona, line 99: Don't forget to add your parameters to your docstring

templates/templates/preview/plain_text/base.template
```diff
@@ -1,5 +1,5 @@
 
-{% include "preview/experiment_plain_text.template" %}
+{% include "preview/plain_text/experiment.template" %}
```
ankona, line 2: Is it possible to use a relative path? If so, we could simplify down to "./experiment.template". As-is, a change in the directory structure will also cause the templates to change. That might be easy to miss...

tests/test_preview.py
```diff
@@ -162,9 +162,9 @@ def test_preview_to_file(test_dir, wlmutils):
     assert path.is_file()
 
 
-def test_active_orch_dict_property(wlmutils, test_dir, choose_host):
+def test_active_orchestrator_jobs_property(wlmutils, test_dir, choose_host):
     """Ensure db_jobs remaines unchanched after deletion
```
ankona, line 166: unchanged!

tests/test_preview.py
```diff
@@ -162,9 +162,9 @@ def test_preview_to_file(test_dir, wlmutils):
...
-    of active_orch_dict property stays intace when retrieving db_jobs"""
+    of active_orchestrator_jobs property stays intace when retrieving db_jobs"""
```
ankona, line 167: intace == ?

ankona (changes requested):
Looks good.

I'd really like if we could avoid changing the templates if the directory structure changes. If we can't do that, all that's left are a couple nitpicks on docstrings.

amandarichardsonn (commented):
Super nice work and incredible amount of code, I can definitely tell a lot of hard work went into this! Here is what I came up with:

Documentation:
I have some suggestions below of where to include information on preview.
1. Make sure the new Experiment.preview is added to doc/api/smartsim_api.rst
2. Add preview to this chart [here](https://www.craylabs.org/develop/experiment.html#id3)
3. Add preview as an Example step [here](https://www.craylabs.org/develop/experiment.html#example)

Preview templates:
1. I would add the path for Ensemble, Model and Orchestrator to the previews! It is possible that a user specifies their own path and therefore it is not located in the exp_path. The path arguments for Orch and Ensemble have been pushed to develop.
2. I notice that in an Experiment with multiple databases, the db_identifier for the standalone dbs are given, even if one is not provided. A colo db identifier is only shown if provided even though the identifier would be default: orchestrator -> should we show the colo db_identifier name always like for std?

**Possible Bug 1:**
I am running the two Model script below:

```
model = exp.create_model("hello_world", settings, path="./")
model_2 = exp.create_model("hello", settings, path="./")
model_2.colocate_db_tcp()
exp.preview(model, model_2)
exp.start(model, model_2)
```

The preview output is:

```
== Models ==

    = Model Name: hello_world =
        Colocated:
          Connection Type: TCP
          TCP/IP Port(s):
            6379
        Client Configuration:
            Database Backend: redis
            Connection Type: TCP
            TCP/IP Port(s):
              6379
            Type: Colocated


    = Model Name: hello =
        Colocated:
          Connection Type: TCP
          TCP/IP Port(s):
            6379
        Client Configuration:
            Database Backend: redis
            Connection Type: TCP
            TCP/IP Port(s):
              6379
            Type: Colocated
```

The preview suggests that both of the Models are colocated rather than the 1? Am I misunderstanding?

Document with more suggestions find [here](https://docs.google.com/document/d/1q9i1UqrggCDEYT8QCJ7ztNVWNEReafmleCby41KJMvU/edit?usp=sharing)

tests/test_preview.py
```diff
@@ -162,9 +162,9 @@ def test_preview_to_file(test_dir, wlmutils):
     assert path.is_file()
```
amandarichardsonn, line 162: Add a test for a model from memory

tests/test_preview.py
juliaputko: corrected to intact

smartsim/_core/control/previewrenderer.py
```diff
@@ -103,9 +128,11 @@ def render(
     :type output_format: _OutputFormatString
     """
 
-    verbosity_level = _check_verbosity_level(Verbosity(verbosity_level))
+    Verbosity(verbosity_level)
```
ankona, line 131: Is this statement doing anything? I think we want the whole thing below?

`env.globals["Verbosity"] = Verbosity(verbosity_level)`

ankona: or, potentially, just `env.globals["Verbosity"] = verbosity_level` since we are being passed the right type and have no need to construct a new object

tests/test_preview.py
ankona, line 162: I think we need to add a `pytest.mark.group_xxx` to this file to ensure it isn't skipped.

Based on that, I'd actually like to see this file split in two:
1. tests that require a WLM/allocation (perhaps `tests/on_wlm/test_preview_active_infra.py`)
2. tests that can complete without (current file)

- Items in (1) can be added to `pytestmark = pytest.mark.slow_tests`
- Items in (2) can be added to `pytestmark = pytest.mark.group_b`

Known "active" tests that fail for me are:
- FAILED tests/test_preview.py::test_active_orchestrator_jobs_property
- FAILED tests/test_preview.py::test_preview_active_infrastructure
- FAILED tests/test_preview.py::test_preview_orch_active_infrastructure
- FAILED tests/test_preview.py::test_preview_multidb_active_infrastructure
- FAILED tests/test_preview.py::test_preview_active_infrastructure_orchestrator_error

smartsim/experiment.py
```diff
@@ -868,49 +868,48 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
     def preview(
         self,
         *args: t.Any,
-        output_format: previewrenderer.Format = previewrenderer.Format.PLAINTEXT,
         verbosity_level: previewrenderer.Verbosity = previewrenderer.Verbosity.INFO,
+        output_format: previewrenderer.Format = previewrenderer.Format.PLAINTEXT,
```
ankona, line 872: Your additions, unfortunately, cause `experiment.py` to go over our 1k LoC limit per file. Let's not up the global limit, yet. Please add:

```py
# pylint: disable=too-many-lines
```

See [`_core/_install/builder.py`](https://github.com/CrayLabs/SmartSim/blob/develop/smartsim/_core/_install/builder.py) for example

tests/test_preview.py
ankona: note: there shouldn't be any reason to require an allocation to run `test_get_ifname_filter` - let's modify that test accordingly.

An option might be:

```py
value = ["_ifname=bad_if_key", "ifname=mock_if_name", "IFname=case_sensitive_key"]
template = "{{value | get_ifname}}"
FILTERS["get_ifname"] = previewrenderer.get_ifname

output = Template(template_str).render(value=value)

assert "mock_if_name" in output
```

We should also add tests establishing expectations for the other test cases shown above in `value`:
- line not starting with ifname
- IFname comparison case sensitivity).

smartsim/_core/control/previewrenderer.py
ankona: I think the intention was originally to do validation, but let's be sure it's required. I think receiving a `Verbosity` parameter should make it unnecessary.

juliaputko: For whatever reason, this is the way I got it to throw a value error when a verbosity isnt valid, but I will dig to make sure its necessary

smartsim/experiment.py
juliaputko: changed

smartsim/experiment.py
```diff
@@ -868,49 +868,48 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
-            developer: Display user defined field and entities, auto generated
+            - developer: Display user-defined field and entities, auto-generated
             fields, and run commands.
-            Defaults to info.
+
+        :param verbosity_level: verbosity level specified by user, defaults to info.
         :type verbosity_level: str
+        :param output_format: Set output format. The possible accepted
+            output formats are `json`, `xml`, `html`, `plain_text`, `color_text`.
+            Defaults to 'plain_text'.
```
AlyssaCote, line 892: Should `json`, `xml`, `html`, and `color_text` be listed as accepted output formats? It seems like right now only `plain_text` is supported.

smartsim/experiment.py
```diff
@@ -868,49 +868,48 @@ def reconnect_orchestrator(self, checkpoint: str) -> Orchestrator:
...
+        :type output_type: str
```
AlyssaCote, line 893: I think this is supposed to be `:type output_format:`

tests/test_preview.py
juliaputko: resolved on a call.

smartsim/experiment.py
juliaputko: good point. Removed others

smartsim/experiment.py
juliaputko: good catch, changed.

PR #532, opened 2024-03-25
Store SmartSim entity logs under the .smartsim directory

AlyssaCote:
This PR's goal is to keep historical entity output files and store them under the .smartsim directory. In order to maintain consistency with what users expect to see in their file structure, we symlink to the output files in the experiment directory. `Step.get_output_files()` is where the output files were changed so that they could be written to `.smartsim/telemetry/{experiment_name}/{run_id}/{entity_type}/{entity_name}/{entity_name}.out or .err` because `Step` holds a `meta` dictionary with the `status_dir` path where we want to write to. The symlinking logic was added to `Controller._launch()` to ensure that the output files are symlinked just because launching the steps.


smartsim/_core/control/controller.py
```diff
@@ -346,6 +347,28 @@ def init_launcher(self, launcher: str) -> None:
         else:
             raise TypeError("Must provide a 'launcher' argument")
 
+    @staticmethod
+    def symlink(
```
al-rigazzi, line 351: Should this function be called `symlink_output` or something similar? Otherwise it may seem like we are symlinking the `Step` and the `Entity`.

smartsim/_core/control/controller.py
```diff
@@ -527,6 +555,7 @@ def _launch_step(
             entity.name not in self._jobs.jobs and entity.name not in self._jobs.db_jobs
         ):
             try:
+                self.symlink(job_step, entity)
```
al-rigazzi, line 558: The call to `symlink` is in a `try`-`except` clause, but we don't actually catch any filesystem related exception. Should we catch such exceptions (possibly raising them with some informative output), or should we remove the call from the block?

smartsim/_core/control/controller.py
```diff
@@ -538,6 +567,7 @@ def _launch_step(
         # that has ran and completed, relaunch the entity.
         elif completed_job is not None and completed_job.entity is entity:
             try:
+                self.symlink(job_step, entity)
```
al-rigazzi, line 570: Same thoughts as in above comment.

al-rigazzi (changes requested):
Looks pretty good! Giving some quick feedback, while I dig a bit deeper into the tests you wrote. Good job so far!

smartsim/_core/control/controller.py
AlyssaCote: Great point. Fixing!

smartsim/_core/control/controller.py
AlyssaCote: I think there are definitely some exceptions that could be raised so I'll add them in with some informative output. Thanks for catching that!

AlyssaCote: I removed the call from the call block and handle the exception within `symlink_output_files`. Let me know what you think about it!

smartsim/_core/control/controller.py
AlyssaCote: Handled in the same way as mentioned above.

smartsim/_core/control/controller.py
```diff
@@ -346,6 +347,35 @@ def init_launcher(self, launcher: str) -> None:
...
+    ) -> None:
+        """Symlink output files under .smartsim to the entity output files
+
+        :param job_step: Job step instance
+        :type job_step: Step
+        :param entity: Entity instance
+        :type entity: SmartSimEntity | EntitySequence[SmartSimEntity]
+        """
+        historical_out, historical_err = job_step.get_output_files()
+        entity_out = osp.join(entity.path, f"{entity.name}.out")
```
ashao, line 362: I _think_ we're moving towards `pathlib` instead of `os.path`. If so and because these are always just local variables, could you change the `entity_out` and `entity_err` to Path objects and update the other `os` operations to use pathlib intrinsics intead.

smartsim/_core/launcher/step/step.py
```diff
@@ -67,9 +68,15 @@ def _create_unique_name(entity_name: str) -> str:
...
-        """Return two paths to error and output files based on cwd"""
-        output = self.get_step_file(ending=".out")
-        error = self.get_step_file(ending=".err")
+        """Return two paths to error and output files based on metadata directory"""
+        try:
+            output_dir = self.meta["status_dir"]
+        except KeyError as exc:
+            raise KeyError("Status directory for this step has not been set.") from exc
+        if not osp.exists(output_dir):
+            pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
```
ashao, line 77: It seems a little odd to have this routine have two effects (1) return the paths to the output and error files and (2) create an output directory. Is there another place where these two lines make more sense.

ashao (commented):
Almost looks ready to me. A couple of minor points to make, but overall a nice clean implementation

smartsim/_core/control/controller.py
AlyssaCote: You got it!

smartsim/_core/launcher/step/step.py
AlyssaCote: Definitely! I'll break it out into its own function and I could either call the new function from within `get_output_files` or I could call it within `_launch_step` right before I call `get_output_files`. Any preference?

smartsim/_core/control/controller.py
AlyssaCote: It looks like entities that call `get_output_files` expect to return strings. Should I start changing all those instances to expect pathlib.Path instead? Or leave it?

smartsim/_core/control/controller.py
```diff
@@ -346,6 +346,35 @@ def init_launcher(self, launcher: str) -> None:
...
+        entity_out = pathlib.Path(entity.path) / f"{entity.name}.out"
+        entity_err = pathlib.Path(entity.path) / f"{entity.name}.err"
+
+        # check if there is already a link to a previous run
+        if entity_out.is_symlink() or entity_err.is_symlink():
+            entity_out.unlink()
+            entity_err.unlink()
+
+        try:
+            entity_out.symlink_to(historical_out)
```
al-rigazzi, line 370: What guarantees that `entity_out` has already been created? Especially for a batch step (which could start after some time), do we create the files in advance? Or will they just be "broken" symlinks until the real files appear?

al-rigazzi (changes requested):
A couple of questions, but otherwise looks good!

smartsim/_core/control/controller.py
al-rigazzi: Well done!

smartsim/_core/control/controller.py
```diff
@@ -346,6 +346,35 @@ def init_launcher(self, launcher: str) -> None:
         else:
             raise TypeError("Must provide a 'launcher' argument")
 
+    @staticmethod
+    def symlink_output_files(
+        job_step: Step, entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]]
+    ) -> None:
+        """Symlink output files under .smartsim to the entity output files
```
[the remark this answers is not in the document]
al-rigazzi: (part of the question actually revolves around what it means "to symlink to", which I'm not totally sure of)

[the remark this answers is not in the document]
AlyssaCote: Nope you're correct! I'll change it to something like "Create symlinks for entity output files that point to the output files under .smartsim"

smartsim/_core/control/controller.py
AlyssaCote: `entity_out` is really just a reference that points to `historical_out`, so we don't need to create actual files on the filesystem for `entity_out` or `entity_err`. In the event that `historical_out` or `historical_err` aren't created yet, it'll throw the `FileNotFoundError` and a symlink will not be created.

doc/changelog.rst
```diff
@@ -34,6 +35,9 @@ Description
 
 Detailed Notes
 
+- The dashboard needs to display historical logs, so log files are written
+  out under the .smartsim directory and symlinked to the files in the
```
al-rigazzi, line 39: Pedantic, but... aren't we actually symlinking the ones in the experiment directory _to_ the ones under `.smartsim`?

al-rigazzi (approved):
Only the changelog entry to be made consistent with the docstrings for symlinking... the rest looks great! Thanks for the prompt responses to my questions and for this quality-of-life improvement!

smartsim/_core/control/controller.py
AlyssaCote: What I did instead was cast `historical_out` and `historical_err` once they're in the symlink function.

smartsim/_core/control/controller.py
```diff
@@ -355,6 +355,36 @@ def init_launcher(self, launcher: str) -> None:
...
+    ) -> None:
+        """Create symlinks for entity output files that point to the output files
+        under the .smartsim directory
+
+        :param job_step: Job step instance
+        :type job_step: Step
+        :param entity: Entity instance
+        :type entity: SmartSimEntity | EntitySequence[SmartSimEntity]
+        """
+        historical_out, historical_err = map(pathlib.Path, job_step.get_output_files())
```
ashao, line 370: Nice thank you!

smartsim/_core/launcher/step/step.py
```diff
@@ -66,10 +67,20 @@ def _create_unique_name(entity_name: str) -> str:
         step_name = entity_name + "-" + get_base_36_repr(time.time_ns())
         return step_name
 
+    @staticmethod
+    def ensure_output_directory_exists(output_dir: str) -> None:
```
ashao, line 71: - Add short docstring
- Pedantic: This function should live with a `utils`-like module. If we don't have one, then don't worry about putting in there, otherwise it would be worth doing for cleanliness

smartsim/_core/launcher/step/step.py
```diff
@@ -66,10 +67,20 @@ def _create_unique_name(entity_name: str) -> str:
...
+        if not osp.exists(output_dir):
+            pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
+
     def get_output_files(self) -> t.Tuple[str, str]:
```
ashao, line 75: Indicate that this a private function with a `_`

ashao (changes requested):
A couple of last very minor changes. Thanks for the work on this and a very nice clean implementation!

smartsim/_core/launcher/step/step.py
AlyssaCote: Docstring added! There is a `util` module but nothing in there related to `step`, so I'm leaving this function in `Step.py`.

smartsim/_core/launcher/step/step.py
AlyssaCote: done!

smartsim/_core/launcher/step/step.py
```diff
@@ -66,10 +67,21 @@ def _create_unique_name(entity_name: str) -> str:
         step_name = entity_name + "-" + get_base_36_repr(time.time_ns())
         return step_name
 
+    @staticmethod
+    def _ensure_output_directory_exists(output_dir: str) -> None:
+        """Create the directory for the step output if it doesn't exist already"""
+        if not osp.exists(output_dir):
+            pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
+
     def get_output_files(self) -> t.Tuple[str, str]:
```
ashao, line 76: One last underscore please

ashao (approved):
Good to go. Thanks @AlyssaCote!

smartsim/_core/launcher/step/step.py
AlyssaCote: Resolving as `get_output_files` isn't a private method.

PR #533, opened 2024-03-26
Change default path for entities

amandarichardsonn:
This PR makes changes to the default path for SS entities. New default path is `exp_path/entity_name/`.


smartsim/experiment.py
```diff
@@ -602,10 +603,12 @@ def create_model(
...
         """
-        path = init_default(getcwd(), path, str)
-
+        if name is None:
+            raise AttributeError("Entity has no name. Please set name attribute.")
+        path = init_default(osp.join(self.exp_path, name), path, str)
         if path is None:
-            path = getcwd()
+            path = osp.join(self.exp_path, name)
+
```
ashao, line 611: I think these lines were already redundant, since path will always be something not None

tests/test_experiment.py
```diff
@@ -229,3 +235,77 @@ def test_telemetry_default(
...
+def test_default_orch_path(test_dir: str, wlmutils: "conftest.WLMUtils") -> None:
+    """Ensure the default file structure is created for Orchestrator"""
+
+    exp_name = "default-orch-path"
+    exp = Experiment(exp_name, launcher="local", exp_path=test_dir)
+    db = exp.create_database(
+        port=wlmutils.get_test_port(), interface=wlmutils.get_test_interface()
+    )
+    exp.start(db)
+    exp.stop(db)
```
ashao, line 249: Instead of actually starting and stopping the database, you could try monkey patching out the launch of the database.

tests/test_experiment.py
```diff
@@ -229,3 +235,77 @@ def test_telemetry_default(
...
+    orch_path = pathlib.Path(test_dir) / db.name
+    err_file = os.path.join(orch_path, f"{db.name}_0.err")
+    out_file = os.path.join(orch_path, f"{db.name}_0.out")
```
ashao, line 252: Instead of using os.path.join use the `/` operator that you used in the above

tests/test_experiment.py
```diff
@@ -229,3 +235,77 @@ def test_telemetry_default(
...
+
+
+def test_default_model_path(test_dir: str) -> None:
+    """Ensure the default file structure is created for Model"""
+
+    exp_name = "default-model-path"
+    exp = Experiment(exp_name, launcher="local", exp_path=test_dir)
+    settings = exp.create_run_settings(exe="echo", exe_args="hello")
+    model = exp.create_model(name="nerp", run_settings=settings)
+    exp.start(model)
```
ashao, line 266: Same as before, instead of actually starting the model, monkeypatch the method so that it doesn't actually execute the model

ashao (changes requested):
Looking pretty good. See the above comments for specific places but overall here are the items to address:
- Try to avoid mixing using os.path and pathlib where possible
- See if you can figure out a way to monkeypatch exp.start so that you can avoid actually having to run the database and other entities

One task to address that's outside of the inline comments:
- Go through the rest of the tests and make sure that we're not unnecessarily setting their paths anymore

smartsim/experiment.py
```diff
@@ -923,6 +926,32 @@ def _launch_summary(self, manifest: Manifest) -> None:
...
+            ensemble_path = osp.join(self.exp_path, ensemble.name)
+            create_dir_and_set_path(ensemble, ensemble_path)
+
+            for member in ensemble.models:
+                # if member.path is None:
+                member_path = osp.join(self.exp_path, ensemble.name, member.name)
+                # else:
+                #     member_path = member.path
+                create_dir_and_set_path(member, member_path)
+
```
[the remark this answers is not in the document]
amandarichardsonn: Changing this function name to `create_entity_dir` and removing all `set_path` since paths are set in factory methods. The portion of the code that creates directories is in the `Generator` class - if not called no directories are created. If I do not create the directories here the experiment will throw an error that the path does not exists since before the path was defaulted to CWD. The telemetry monitor does create a experiment dir if DNE. I would suggest that we keep the dir generation here (unless you have another idea!). Telemetry should still create an Experiment directory bc if I create an Orchestrator using the constructor, the path is defaulted to CWD, no Exp will be created. I also do not believe I can use the Generator class since this uses set_paths rn to overwrite path arguments and also attached all generator files. But I might be convince to use the Generator function once the bug is fixed?

tests/test_experiment.py
```diff
@@ -229,3 +235,77 @@ def test_telemetry_default(
...
+    """Ensure the default file structure is created for Model"""
+
+    exp_name = "default-model-path"
+    exp = Experiment(exp_name, launcher="local", exp_path=test_dir)
+    settings = exp.create_run_settings(exe="echo", exe_args="hello")
+    model_path = os.path.join(test_dir, "test_folder")
+    model = exp.create_model(name="nerp", run_settings=settings, path=model_path)
+    exp.start(model)
+    err_file = os.path.join(model_path, f"{model.name}.err")
+    out_file = os.path.join(model_path, f"{model.name}.out")
```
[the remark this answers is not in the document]
amandarichardsonn: Since I am MonkeyPatching start, SS is no longer generating these files - removing

smartsim/experiment.py
```diff
@@ -491,10 +493,16 @@ def create_ensemble(
         :return: ``Ensemble`` instance
         :rtype: Ensemble
         """
+        if name is None:
+            raise AttributeError("Entity has no name. Please set name attribute.")
+        check_path = init_default(osp.join(self.exp_path, name), path, str)
+        entity_path: str = osp.abspath(check_path)
+
```
ashao, line 500: Let's try not to use the `init_default` function here. What's wrong with just doing

```
entity_path = osp.join(self.exp_path, name)
```

smartsim/experiment.py
```diff
@@ -813,9 +826,11 @@ def create_database(
         """
 
         self.append_to_db_identifier_list(db_identifier)
-
+        check_path = init_default(osp.join(self.exp_path, db_identifier), path, str)
+        entity_path: str = osp.abspath(check_path)
```
ashao, line 830: Same as above, what's wrong with just making the path directly if?

tests/test_experiment.py
```diff
@@ -60,6 +63,20 @@ def test_model_prefix(test_dir: str) -> None:
     assert model._key_prefixing_enabled == True
 
 
+def test_model_no_name():
+    exp = Experiment("test_model_no_name")
+    with pytest.raises(AttributeError):
+        _ = exp.create_model(name=None, run_settings=RunSettings("python"))
+
+
+def test_model_no_name():
```
ashao, line 72: This test needs to be renamed to `test_ensemble_no_name`; it's currently overriding the test above it.

tests/test_experiment.py
```diff
@@ -229,3 +246,126 @@ def test_telemetry_default(
...
+        port=wlmutils.get_test_port(),
+        interface=wlmutils.get_test_interface(),
+        path="./testing_folder1234",
+    )
+    exp.start(db)
+    orch_path = pathlib.Path(osp.abspath("./testing_folder1234"))
+    assert orch_path.exists()
+    assert db.path == str(orch_path)
+    shutil.rmtree(orch_path)
+    assert not orch_path.exists()
```
ashao, line 324: Extra assert is probably not needed

tests/test_experiment.py
```diff
@@ -229,3 +246,126 @@ def test_telemetry_default(
...
+    settings = exp.create_run_settings(exe="echo", exe_args="hello")
+    model = exp.create_model(
+        name="model_name", run_settings=settings, path="./testing_folder1234"
+    )
+    exp.start(model)
+    model_path = pathlib.Path(osp.abspath("./testing_folder1234"))
+    assert model_path.exists()
+    assert model.path == str(model_path)
+    shutil.rmtree(model_path)
+    assert not model_path.exists()
```
ashao, line 344: Same here

tests/test_experiment.py
```diff
@@ -229,3 +246,126 @@ def test_telemetry_default(
...
+    exp.start(ensemble)
+    ensemble_path = pathlib.Path(osp.abspath("./testing_folder1234"))
+    assert ensemble_path.exists()
+    assert ensemble.path == str(ensemble_path)
+    for member in ensemble.models:
+        member_path = ensemble_path / member.name
+        assert member_path.exists()
+        assert member.path == str(member_path)
+    shutil.rmtree(ensemble_path)
+    assert not ensemble_path.exists()
```
ashao, line 371: Same here

ashao (changes requested):
Thanks for addressing my previous comments. I had a handful of minor points this time around, but nothing substantive. Good job!

tests/test_experiment.py
amandarichardsonn: good catch! Thank you

smartsim/experiment.py
amandarichardsonn: yeah no need for init_default, Ill change

doc/experiment.rst
```diff
@@ -211,6 +211,103 @@ An ``Ensemble`` supports key features, including methods to:
...
+for more details.
+
+.. _default_folder:
+
+Default
+=======
+By default, an ``Experiment`` folder is created in your current working directory, using the
+specified `name` parameter during ``Experiment`` initialization. Each entity created by the
+``Experiment`` generates an output folder under the ``Experiment`` directory, named after the
+entity. These folders hold `.err` and `.out` files, containing execution-related information.
```
mellis13, line 230: Should we call out that files that were added to be symlinked, copied, and/or configured will not be placed in these directories unless experimernt.generate() is called with the entity?

mellis13 (changes requested):

doc/experiment.rst
amandarichardsonn: 100% Ill add

mellis13 (approved):

ashao (approved):
Looks good. Thanks for improvements!

PR #535, opened 2024-03-28
Enhanced Signal Management

MattToast:
Fixes unfalsifiable test that tests SmartSim's custom SIGINT signal handler. Adds infrastructure to make the test pass again.

### DIFF
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -121,7 +121,9 @@ class Controller:
         self._jobs.kill_on_interrupt = kill_on_interrupt
 
         # register custom signal handler for ^C (SIGINT)
-        SignalInterceptionStack.push(signal.SIGINT, self._jobs.signal_interrupt)
+        SignalInterceptionStack.get(signal.SIGINT).push_unique(
+            self._jobs.signal_interrupt
+        )
         launched = self._launch(exp_name, exp_path, manifest)
 
         # start the job manager thread if not already started
--- a/smartsim/_core/control/controller.py
+++ b/smartsim/_core/control/controller.py
@@ -71,6 +71,8 @@ from .jobmanager import JobManager
 from .manifest import LaunchedManifest, LaunchedManifestBuilder, Manifest
 
 if t.TYPE_CHECKING:
+    from types import FrameType
+
     from ..utils.serialize import TStepLaunchMetaData
 
 
@@ -113,8 +115,15 @@ class Controller:
         execution of all jobs.
         """
         self._jobs.kill_on_interrupt = kill_on_interrupt
+
         # register custom signal handler for ^C (SIGINT)
-        signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
+        handle = signal.getsignal(signal.SIGINT)
+        # XXX: THIS IS A DANGEROUS MEMORY LEAK
+        #      Every time this is called a new callback fn is registered
+        #      I need to find a way to make this not keep wrapping itself
+        #      for every new JM.
+        signal.signal(signal.SIGINT, _hacky_signal_handle(self._jobs, handle))
+
         launched = self._launch(exp_name, exp_path, manifest)
 
         # start the job manager thread if not already started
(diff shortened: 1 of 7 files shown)


smartsim/_core/utils/helpers.py
```diff
@@ -302,3 +306,65 @@ def decode_cmd(encoded_cmd: str) -> t.List[str]:
...
+        :type fn: t.Callable[[int, FrameType | None], t.Any]
+        :returns: The signal interception stack to which the callback was
+                  appended
+        :rtype: SignalInterceptionStack
+        """
+        handler = signal.getsignal(signalnum)
+        if isinstance(handler, cls):
+            # pylint: disable=protected-access
+            if fn not in handler._callbacks:
+                handler._callbacks.append(fn)
```
ashao, line 353: Same here, are we anticipating that there might actually be instances where we are regularly duplicating callbacks?

tests/test_helpers.py
```diff
@@ -68,3 +71,91 @@ def test_encode_raises_on_empty():
...
+
+
+def test_signal_intercept_stack_enforces_that_handlers_are_unique(mock_signal):
+    handler = lambda num, frame: ...
+    stack = helpers.SignalInterceptionStack.push(signal.NSIG, handler)
+    helpers.SignalInterceptionStack.push(signal.NSIG, handler)
+    assert list(stack._callbacks) == [handler]
+
+
+def test_singnal_intercept_stack_enforces_that_method_handlers_are_unique(mock_signal):
```
ashao, line 141: singnal -> signal

smartsim/_core/control/controller.py
```diff
@@ -113,8 +119,9 @@ def start(
         execution of all jobs.
         """
         self._jobs.kill_on_interrupt = kill_on_interrupt
+
         # register custom signal handler for ^C (SIGINT)
-        signal.signal(signal.SIGINT, self._jobs.signal_interrupt)
+        SignalInterceptionStack.push(signal.SIGINT, self._jobs.signal_interrupt)
```
ashao, line 124: Alternatively we could check to see if the signal_interrupt is already in the list of SignalInterceptionStack callback lists before adding.

ashao (changes requested):
Looks really good and thanks for fixing this up. It was both a developer and user annoyance. Two comments:

Could you change the name of this PR so that it reflects the effect that these changes have?

I'm not quite seeing the need to have enforce unique callables within the SignalInterceptionStack class itself. I can see that with multiple `start` calls, you're going to register multiple signals. I'm going back and forth on if it's 100% necessary that that the class enforces uniqueness (and add to some of its complexity) or whether there's a world where it might be useful to have multiple identical callbacks. For example, the current implementation prevents you from popping something multiple times adding it as a callback.

For the one place that this class is used, it's clearly incorrect for the signal handle to be registered more than once. I'm about 60% on the side of that it is the Controller's responsiblity to make sure it doesn't do wonky things when registering a callback and not the class' responsibility. Happy to be argued the other way though

smartsim/_core/utils/helpers.py
```diff
@@ -302,3 +306,65 @@ def decode_cmd(encoded_cmd: str) -> t.List[str]:
...
+        ] = None,
+    ) -> None:
+        """Set up a ``SignalInterceptionStack`` for particular signal number.
+
+        :param signalnum: The signal number to intercept
+        :type signalnum: int
+        :param callbacks: A iterable of functions to call upon receiving the signal
+        :type callbacks: t.Iterable[t.Callable[[int, FrameType | None], t.Any]]
+        """
+        self._callbacks = list(set(callbacks)) if callbacks else []
```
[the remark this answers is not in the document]
MattToast: Agreed! Per offline discussion, decided to revert this requirement

tests/test_helpers.py
MattToast: Spelling is hard :(

smartsim/_core/control/controller.py
MattToast: Good point! Per offline discussion, decided that we didn't want to exposed the `_callbacks` attr to users of the class, but did decided to have the controller know to call a new `push_unique` method

ashao (approved):
Looks great. Thanks for making hte changes

PR #537, opened 2024-04-02
Readthedocs import extension error

amandarichardsonn:
Readthedocs fails and is blocking existing PRs. Failure is: `Extension error: Could not import extension sphinx_tabs.tabs (exception: cannot import name 'TypeAliasType' from 'typing_extensions')`. The issue came from pydantic==2.6.4 and typing_extensions==4.5.0. Sphinx uses Open AI which requires "pydantic>=1.9.0, <3", "typing-extensions>=4.7, <5". The versions have been changed in `requirements-doc.txt`.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -29,8 +29,8 @@ build:
       - ln -s smartredis/examples ./examples
       - cd smartdashboard; python -m pip install .
     pre_build:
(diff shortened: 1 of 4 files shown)


al-rigazzi (approved):
LGTM, thanks for looking into this and fixing it so elegantly!

PR #538, opened 2024-04-02
Optionally skip building Torch with Intel MKL

ashao:
On systems that have the Intel Compilers and/or the Intel Math Kernel library installed, the Caffe2 package that comes with Torch will unconditionally try to link in the MKL during the Torch backend. This however can lead to two types of failures:
- Problems when compiling the Torch backend because the linker does not include the path to the MKL library path
- Loading the Torch backend into RedisAI fails because the user does not expect to need to have the MKL library loaded.

To alleviate this, a new option "--no_torch_with_mkl" has been added to the `smart build` command that modifies the mkl.cmake file to prevent the detection of MKL.

### DIFF
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -139,6 +139,7 @@ def build_redis_ai(
     torch_dir: t.Union[str, Path, None] = None,
     libtf_dir: t.Union[str, Path, None] = None,
     verbose: bool = False,
+    torch_with_mkl=True,
 ) -> None:
     # make sure user isn't trying to do something silly on MacOS
     if build_env.PLATFORM == "darwin" and device == Device.GPU:
@@ -186,6 +187,7 @@ def build_redis_ai(
         build_tf=use_tf,
         build_onnx=use_onnx,
         verbose=verbose,
+        torch_with_mkl=torch_with_mkl,
     )
 
     if rai_builder.is_built:
@@ -414,6 +416,7 @@ def execute(
                 args.torch_dir,
                 args.libtensorflow_dir,
                 verbose=verbose,
+                torch_with_mkl=args.torch_with_mkl,
             )
     except (SetupError, BuildError) as e:
         logger.error(str(e))
@@ -495,4 +498,11 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
         action="store_true",
         default=False,
         help="Build KeyDB instead of Redis",
+    )
+
+    parser.add_argument(
+        "--torch_with_mkl",
+        action="store_true",
+        default=True,
+        help="Build torch with Intel MKL support",
     )
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -506,4 +506,3 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
         action="store_false",
         help="Do not build Torch with Intel MKL",
     )
-    parser.set_defaults(torch_with_mkl=True)
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -139,7 +139,7 @@ def build_redis_ai(
     torch_dir: t.Union[str, Path, None] = None,
     libtf_dir: t.Union[str, Path, None] = None,
     verbose: bool = False,
-    torch_with_mkl=True,
+    torch_with_mkl: bool = True,
 ) -> None:
     # make sure user isn't trying to do something silly on MacOS
     if build_env.PLATFORM == "darwin" and device == Device.GPU:
--- a/smartsim/_core/_cli/build.py
+++ b/smartsim/_core/_cli/build.py
@@ -501,8 +501,23 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
     )
 
     parser.add_argument(
+<<<<<<< HEAD
         "--torch_with_mkl",
         action="store_true",
         default=True,
         help="Build torch with Intel MKL support",
     )
+=======
+        "--torch_with_mkl",
+        dest="torch_with_mkl",
+        action="store_true",
+        help="Build Torch with Intel MKL (if available)"
+    )
+    parser.add_argument(
+        "--no_torch_with_mkl",
+        dest="torch_with_mkl",
+        action="store_false",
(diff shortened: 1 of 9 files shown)


smartsim/_core/_cli/build.py
```diff
@@ -139,6 +139,7 @@ def build_redis_ai(
     torch_dir: t.Union[str, Path, None] = None,
     libtf_dir: t.Union[str, Path, None] = None,
     verbose: bool = False,
+    torch_with_mkl=True,
```
MattToast, line 142: Missing type
```suggestion
    torch_with_mkl: bool = True,
```

smartsim/_core/_cli/build.py
```diff
@@ -496,3 +499,17 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
...
+        action="store_true",
+        help="Build Torch with Intel MKL (if available)"
+    )
+    parser.add_argument(
+        "--no_torch_with_mkl",
+        dest="torch_with_mkl",
+        action="store_false",
+        help="Do not build Torch with Intel MKL"
+    )
+    parser.set_defaults(torch_with_mkl=True)
```
MattToast, line 515: Not wrong, just be warned: 

I'm not usually a fan of this style of args as this is legal input and will be parsed by `argparse` w/o error
```sh
$ smart build --torch_with_mkl --no_torch_with_mkl
# args.troch_with_mkl == False
$ smart build  --no_torch_with_mkl --torch_with_mkl
# args.torch_with_mkl == True
```
I would vote to return to a single `--no_torch_with_mkl` arg flag with a `'store_false'` directive, similar to how we do `--no_tf`/`--no_pt`/etc., but if we do decide to keep a this style, can we error out if a user were specify both flags just so that they don't have something unexpected happen?

smartsim/_core/_cli/build.py
```diff
@@ -496,3 +499,17 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
         default=False,
         help="Build KeyDB instead of Redis",
     )
+
```
MattToast, line 502: VERY petty style thing: This is the only new line in this function between the `add_argument` calls and my OCD cannot handle it. Can you please remove or add a banner comment or something, lol?

MattToast (changes requested):
Looks good! Just a couple of small things, but otherwise lgtm!!

smartsim/_core/_install/builder.py
```diff
@@ -1051,3 +1067,14 @@ def config_git_command(plat: Platform, cmd: t.Sequence[str]) -> t.List[str]:
             + cmd[where:]
         )
     return cmd
+
+
+def _modify_source_files(
+    files: t.Union[_PathLike, t.Iterable[_PathLike]], regex: str, replacement: str
+) -> None:
```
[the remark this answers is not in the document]
ashao: Good call.

smartsim/_core/_install/builder.py
```diff
@@ -1051,3 +1067,14 @@ def config_git_command(plat: Platform, cmd: t.Sequence[str]) -> t.List[str]:
...
+    compiled_regex = re.compile(regex)
+    patcher: t.Callable[[str], str] = lambda line: compiled_regex.sub(replacement, line)
+    with fileinput.input(files=files, encoding="utf-8", inplace=True) as f:
+        for line in f:
+            line = patcher(line)
+            print(line, end="")
```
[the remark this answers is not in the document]
ashao: Prefer Option 1 since I kind of want to comment the `print(line, end"")` line since it already feels like a weird construct.

smartsim/_core/_cli/build.py
ashao: I think black did that on me. Let me remove and double-check

smartsim/_core/_cli/build.py
```diff
@@ -496,3 +499,25 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+<<<<<<< HEAD
```
al-rigazzi, line 504: Merge artifact here?

al-rigazzi (commented):
Quick comment (I think it's quite superfluous but...)

smartsim/_core/_cli/build.py
ashao: Yah, missed a git add :facepalm:

smartsim/_core/_cli/build.py
ashao: Yep black cares little about your feelings. It's staying in

MattToast: How could black betray me like this 😢 . All well, seems reasonable to me

smartsim/_core/_cli/build.py
```diff
@@ -496,3 +499,11 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
...
         help="Build KeyDB instead of Redis",
     )
+
+    parser.add_argument(
+        "--no_torch_with_mkl",
+        dest="torch_with_mkl",
+        action="store_false",
+        help="Do not build Torch with Intel MKL",
+    )
+    parser.set_defaults(torch_with_mkl=True)
```
MattToast, line 509: No need for this line! the default should be set to true by `"store_false"`

smartsim/_core/_install/builder.py
```diff
@@ -854,9 +860,19 @@ def supported_platforms() -> t.Sequence[t.Tuple[OperatingSystem, Architecture]]:
...
+            libtorch_root / "share/cmake/Caffe2/public/mkl.cmake",
+            r"find_package\(MKL QUIET\)",
+            "# find_package(MKL QUIET)",
+        )
+
     def __place_for_rai__(self, target: t.Union[str, "os.PathLike[str]"]) -> Path:
         self.extract(target)
         target = Path(target) / "libtorch"
+        if not self.with_mkl:
+            self._patch_out_mkl(target)
```
MattToast, line 875: Should we do this patch out on every extract, not just on extraction for RAI, sense `with_mkl` is a class attr?
```py
def extract(target):
    super().extract(target)
    if not self.with_mkl:
        self._patch_out_mkl(target / "libtorch")
```
?

MattToast (approved):
Two small nits, and pending test suite, but otherwise LGTM!!

PR #540, opened 2024-04-04
Update watchdog dependency

ankona:
Updated watchdog dependency pin to next major version and removed `type: ignore` where possible due to new type hints added to watchdog


AlyssaCote (changes requested):
This looks great! The only thing holding up the approval for me is I'm curious why the `self._observer` starts and stops don't have their `# type: ignore` comments removed because it looks like `BaseObserver` os a `watchdog` class.

AlyssaCote (approved):
Thanks for answering my lingering question!

PR #542, opened 2024-04-09
Decouple authenticator and socket creation

ankona:
1. ZMQ authenticators appear to have clashing inproc addresses when using the `zmq.Context.instance()` factory method. Replaced as needed.
2. Updated underlying `Dragon` library version, which included a breaking changing causing the swap from `TemplateProcess` to `ProcessTemplate`
3. Fixed incomplete permission set on curve key files

### DIFF
--- a/smartsim/_core/launcher/dragon/dragonLauncher.py
+++ b/smartsim/_core/launcher/dragon/dragonLauncher.py
@@ -76,7 +76,6 @@ logger = get_logger(__name__)
 _SchemaT = t.TypeVar("_SchemaT", bound=t.Union[DragonRequest, DragonResponse])
 
 DRG_LOCK = RLock()
-DRG_CTX = zmq.Context()
 
 
 class DragonLauncher(WLMLauncher):
@@ -92,12 +91,10 @@ class DragonLauncher(WLMLauncher):
 
     def __init__(self) -> None:
         super().__init__()
-        self._context = DRG_CTX
+        self._context = zmq.Context()
         self._timeout = CONFIG.dragon_server_timeout
         self._reconnect_timeout = CONFIG.dragon_server_reconnect_timeout
         self._startup_timeout = CONFIG.dragon_server_startup_timeout
-        self._context.setsockopt(zmq.SNDTIMEO, value=self._timeout)
-        self._context.setsockopt(zmq.RCVTIMEO, value=self._timeout)
         self._dragon_head_socket: t.Optional[zmq.Socket[t.Any]] = None
         self._dragon_head_process: t.Optional[subprocess.Popen[bytes]] = None
         # Returned by dragon head, useful if shutdown is to be requested
@@ -105,13 +102,15 @@ class DragonLauncher(WLMLauncher):
         self._dragon_head_pid: t.Optional[int] = None
         self._authenticator: t.Optional[zmq.auth.thread.ThreadAuthenticator] = None
 
+        self._set_timeout(self._timeout)
+
     @property
     def is_connected(self) -> bool:
         return self._dragon_head_socket is not None
 
     def _handshake(self, address: str) -> None:
-        self._dragon_head_socket, self._authenticator = dragonSockets.get_secure_socket(
-            self._context, zmq.REQ, False, self._authenticator
+        self._dragon_head_socket = dragonSockets.get_secure_socket(
+            self._context, zmq.REQ, False
         )
         self._dragon_head_socket.connect(address)
         try:
@@ -149,6 +148,8 @@ class DragonLauncher(WLMLauncher):
             path = _resolve_dragon_path(path)
             dragon_config_log = path / CONFIG.dragon_log_filename
 
+            self._authenticator = dragonSockets.get_authenticator(self._context)
+
             if dragon_config_log.is_file():
                 dragon_confs = self._parse_launched_dragon_server_info_from_files(
                     [dragon_config_log]
@@ -180,14 +181,19 @@ class DragonLauncher(WLMLauncher):
                 "smartsim._core.entrypoints.dragon",
             ]
 
+            # Optionally configure the dragon host for debug output
+            if CONFIG.log_level in ["debug", "developer"]:
+                cmd.insert(1, "DEBUG")
+                cmd.insert(1, "-l")
+
             address = get_best_interface_and_address().address
             socket_addr = ""
             launcher_socket: t.Optional[zmq.Socket[t.Any]] = None
             if address is not None:
                 self._set_timeout(self._startup_timeout)
 
-                launcher_socket, self._authenticator = dragonSockets.get_secure_socket(
-                    self._context, zmq.REP, True, self._authenticator
+                launcher_socket = dragonSockets.get_secure_socket(
+                    self._context, zmq.REP, True
                 )
 
                 # find first available port >= 5995
@@ -206,6 +212,8 @@ class DragonLauncher(WLMLauncher):
             ) as dragon_err:
                 current_env = os.environ.copy()
                 current_env.update({"PYTHONUNBUFFERED": "1"})
+                logger.debug(f"Starting Dragon environment: {' '.join(cmd)}")
+
                 # pylint: disable-next=consider-using-with
                 self._dragon_head_process = subprocess.Popen(
                     args=cmd,
@@ -258,6 +266,7 @@ class DragonLauncher(WLMLauncher):
                 server_socket = self._dragon_head_socket
                 server_process_pid = self._dragon_head_process.pid
 
+                # avoid registering the same cleanup more than once
                 if server_socket is not None and self._dragon_head_process is not None:
                     atexit.register(
                         _dragon_cleanup,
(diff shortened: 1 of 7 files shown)


smartsim/_core/launcher/launcher.py
```diff
@@ -140,7 +143,13 @@ def get_step_update(
         s_names, step_ids = self.step_mapping.get_ids(step_names, managed=True)
         if len(step_ids) > 0:
             _step_ids = [str(sid) for sid in step_ids]
-            s_statuses = self._get_managed_step_update(_step_ids)
+            s_statuses: t.List[StepInfo] = []
+
+            try:
+                s_statuses = self._get_managed_step_update(_step_ids)
+            except Exception:
+                logger.warning(f"Step update retrieval failed for: {_step_ids}")
```
MattToast, line 151: Why are we silencing this exception? I'm not sure I want to commit to making this API break. Is it strictly necessary for the authenticator changes?

smartsim/_core/utils/telemetry/manifest.py
```diff
@@ -229,7 +245,7 @@ def load_manifest(file_path: str) -> t.Optional["RuntimeManifest"]:
             raise ValueError("Manifest missing required runs")
 
         exp_dir = pathlib.Path(exp["path"])
-        runs = [Run.load_run(raw_run, exp_dir) for raw_run in runs]
+        runs = [Run.load_run(raw_run, exp_dir, manifest_dict) for raw_run in runs]
```
MattToast, line 248: I'm not sure how I feel about shallow copying `runs` and `exp_dir` out of `manifest_dict`  but then passing `manifest_dict` around the everywhere alongside the copied data. Can we just pass around the `manifest_dict` instead??

smartsim/_core/utils/telemetry/telemetry.py
```diff
@@ -316,12 +318,24 @@ async def on_timestep(self, timestamp: int) -> None:
...
+                logger.warning(f"Telemetry step updates failed for {names.keys()}")
+
+            try:
+                for step_name, step_info in step_updates:
+                    if step_info and step_info.status in TERMINAL_STATUSES:
+                        completed_entity = names[step_name]
+                        await self._to_completed(timestamp, completed_entity, step_info)
+            except Exception as ex:
+                msg = f"An error occurred getting step updates on {names}"
+                logger.error(msg, exc_info=ex)
```
MattToast, line 338: This PR seems to have grown in scope some what. Are these telemetry monitor changes strictly related socket/authenticator changes? Is there any chance we could separate them into two or more PRs in case we need to roll one back?

If not consider updating the PR title/descriptions to encompass more of these TM related changes.

tests/test_dragon_launcher.py
```diff
@@ -108,19 +116,18 @@ def is_alive(self) -> bool:
...
     """Create a mock dragon environment that can talk to the launcher through ZMQ"""
+    logger = logging.getLogger(__name__)
+    logging.basicConfig(level=logging.DEBUG)
+
     try:
-        context = zmq.Context()
         addr = "127.0.0.1"
         callback_port = kwargs["port"]
         head_port = find_free_port(start=callback_port + 1)
+        context = zmq.Context.instance()
```
MattToast, line 126: Not wrong just checking: I thought I heard you mention we should try and stay away from the `instance` factory method. Is this intended

tests/test_dragon_launcher.py
```diff
@@ -202,15 +225,14 @@ def test_secure_socket_authenticator_setup(
...
-    context = zmq.Context()
 
     with monkeypatch.context() as ctx:
         # look at test dir for dragon config
         ctx.setenv("SMARTSIM_KEY_PATH", test_dir)
         # avoid starting a real authenticator thread
         ctx.setattr("zmq.auth.thread.ThreadAuthenticator", MockAuthenticator)
 
-        _, authenticator = get_secure_socket(context, socket_type, is_server=is_server)
+        authenticator = get_authenticator(zmq.Context.instance())
```
MattToast, line 235: Same "is this `instance` call intended" question here

tests/test_dragon_launcher.py
```diff
@@ -220,7 +242,7 @@ def test_secure_socket_authenticator_setup(
         assert authenticator.num_configure_curves > 0
         # ensure authenticator was started
         assert authenticator.num_starts > 0
-        assert authenticator.context == context
+        assert authenticator.context == zmq.Context.instance()
```
MattToast, line 245: and here

MattToast: Looks like there are a few more calls across the test suite. If it was intended, feel free to resolve all of these comments!

MattToast (changes requested):
Throwing some initial feedback your way! In general I think the authenticator changes look good, but I noticed that this PR was also making some changes along side the launchers and the telemetry monitor that didn't look strictly related. Is there anyway we can break them out into their own PRs in case we need to roll one back?

smartsim/_core/launcher/dragon/dragonLauncher.py
```diff
@@ -180,14 +182,19 @@ def _connect_to_dragon(self, path: t.Union[str, "os.PathLike[str]"]) -> None:
                 "smartsim._core.entrypoints.dragon",
             ]
 
+            # Optionally configure the dragon host for debug output
+            if CONFIG.log_level == "debug":
+                cmd.insert(1, "DEBUG")
+                cmd.insert(1, "-l")
```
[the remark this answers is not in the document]
ankona: Love it. Done.

smartsim/_core/launcher/dragon/dragonLauncher.py
```diff
@@ -258,24 +267,27 @@ def log_dragon_outputs() -> None:
                 server_socket = self._dragon_head_socket
                 server_process_pid = self._dragon_head_process.pid
 
-                if server_socket is not None and self._dragon_head_process is not None:
```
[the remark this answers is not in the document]
ankona: I don't like the old conditional. It needed to be updated each time a new resource was added...

I think the same launcher shouldn't be connected to multiple dragon environments. I don't think we need to support calling this more than once (and the null checks aren't discerning enough to recognize trying to register multiple unique cleanups).

How about:

```
                if atexit._ncallbacks() == 0:  # pylint: disable=protected-access
                    atexit.register(
                        _dragon_cleanup,
                        server_socket=server_socket,
                        server_process_pid=server_process_pid,
                        server_authenticator=self._authenticator,
                    )
```

smartsim/_core/launcher/launcher.py
ankona: I found that a bad step ID could cause the telemetry monitor / JobManager to go down in flames... But on reflection, that's usually due to not cleaning up old manifests.

Removed.

smartsim/_core/utils/telemetry/manifest.py
ankona: I wanted to give the deserialization the object it was working on but also let it see higher level context info when it needed to do something wacky, like change behavior of  if it's a dragon launcher. I didn't like sending `is_dragon` all the way down, nor do I like the idea of having to add a `launcher: str` param, then another later, etc...

I think I like switching to passing only the experiment. That achieves my goal but avoids passing the entire manifest around. What do you think of that?

tests/test_dragon_launcher.py
ankona: It's fine in tests (and I think I was probably wrong about the context when I discovered that the debug level dragon logs showed file permission issues).

tests/test_dragon_launcher.py
ankona: Ok for tests, resolving!

smartsim/_core/launcher/dragon/dragonLauncher.py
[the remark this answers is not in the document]
ankona: I don't like `_ncallbacks` - it feels like I'm just asking for a weird break later on when we use `atexit.register` somewhere new.

I can avoid dupes with this instead:

```
# avoid registering the same cleanup more than once
                atexit.unregister(_dragon_cleanup)
                atexit.register(
                    _dragon_cleanup,
                    server_socket=server_socket,
                    server_process_pid=server_process_pid,
                    server_authenticator=self._authenticator,
                )
```

Unregister [silently does nothing](https://docs.python.org/3/library/atexit.html#atexit.unregister) if a callback isn't registered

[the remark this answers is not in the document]
ankona: The more I considered it, the more I believed that allowing it to register multiple cleanups was the right choice. I reverted the change.

smartsim/_core/utils/telemetry/telemetry.py
ankona: Split in two

smartsim/_core/utils/telemetry/manifest.py
MattToast: That makes sense to me, and resolves my "multiple parameters pointing to same data" complaint! Totally in favor!!

MattToast (approved):
LGTM pending tests!! Thanks for cleaning this up!!

al-rigazzi (approved):
L
G
T
M

PR #544, opened 2024-04-09
Drop python 3.8

AlyssaCote:
Python 3.8 is no longer going to be supported because it's near end of life. I've gone through and removed/updated everything that I could find! I added `no-merge` because I am not sure if there's a specific order that we need to merge all the repos.


pyproject.toml
```diff
@@ -31,7 +31,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 88
-target-version = ['py38', 'py39', 'py310']
+target-version = ['py39', 'py310']
```
MattToast, line 34: Not strictly related to this PR, but can we add `'py311'` to this list while we are here?

MattToast (approved):
One small request, but nothing worth holding up approval over! LGTM!!

pyproject.toml
AlyssaCote: you got it!

mellis13 (approved):
Thank you!  LGTM

PR #547, opened 2024-04-12
Symlink batch ensembles and batch models

AlyssaCote:
After testing a bunch of batch ensembles and batch models, I found that I hadn't actually symlinked the substeps in the controller. This fix should properly symlink the substeps.


smartsim/_core/control/controller.py
```diff
@@ -385,6 +385,7 @@ def symlink_output_files(
                 "Symlinking files failed."
             ) from fnf
 
+    # pylint: disable-msg=too-many-locals
```
ankona, line 388: I know you didn't create the problem but it does look like we could probably break out a few chunks of this method into new methods and avoid the warning altogether. Can you give this a once over to see if anything is easy to pull out before we commit to disabling `too-many-locals`?

Maybe we could pull out the `for orchestrator in manifest.dbs:` loop into `_launch_orchestrators`, for example...

smartsim/_core/control/controller.py
```diff
@@ -572,7 +582,11 @@ def _launch_step(
         if completed_job is None and (
             entity.name not in self._jobs.jobs and entity.name not in self._jobs.db_jobs
         ):
-            self.symlink_output_files(job_step, entity)
+            # make sure it's not an _AnonymousBatchJob
```
ankona, line 585: I'm not sure this comment is accurate (since we're looking for `isinstance(Model)`). Consider tweaking your conditional to state your intent instead of relying solely on the comment?

```
not_a_model = not isinstance(entity, Model)
is_non_batch_model = not not_a_model and entity.batch_settings

if not_a_model or is_non_batch_model:
   self.symlink_output_files(job_step, entity)

```

ankona: if we're REALLY looking to see that it's an `_AnonymousBatchJob` what stops us from checking directly w/`if not isinstance(entity, _AnonymousBatchJob)`?

smartsim/_core/control/controller.py
```diff
@@ -584,7 +598,11 @@ def _launch_step(
         # if the completed job does exist and the entity passed in is the same
         # that has ran and completed, relaunch the entity.
         elif completed_job is not None and completed_job.entity is entity:
-            self.symlink_output_files(job_step, entity)
+            # make sure it's not an _AnonymousBatchJob
+            if not isinstance(entity, Model) or (
```
ankona, line 602: we're repeating the same check twice. can we do it once before the if/else starts and just use the once result?

smartsim/_core/control/controller.py
ankona, line 585: From a little further back, it seems that `_launch_step` is now completely in the business of doing symlinks. Can you consider extracting a method for the expanded symlinking needs instead of making `_launch_step` do it all?

AlyssaCote: What do you think about adding the if check to the `symlink_output_files` method?

AlyssaCote: Originally tried checking with `if not isinstance(entity, _AnonymousBatchJob)` and it didn't work, which I thought was interesting. So that's why it's broken up into an ugly conditional.

AlyssaCote: just kidding that doesn't work.

AlyssaCote: waaaaaait actually i think i tried `if not isinstance(job_step, _AnonymousBatchJob)`. oops! let me try your way!

AlyssaCote: That also doesn't work.

tests/test_output_files.py
```diff
@@ -225,3 +230,40 @@ def test_symlink_error(test_dir):
     bad_step = controller._create_job_step(bad_model, telem_dir)
     with pytest.raises(FileNotFoundError):
         controller.symlink_output_files(bad_step, bad_model)
+
+
+def test_batch_model_and_ensemble(test_dir):
```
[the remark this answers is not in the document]
AlyssaCote: Working on it!

smartsim/_core/control/controller.py
AlyssaCote: I pulled the symlinking calls out of `launch_step` because honestly they didn't need to be there. Because I did that, the AnonymousBatchJob doesn't overwrite its symlink anymore, so this logic is no longer needed!

smartsim/_core/control/controller.py
AlyssaCote: moved calls to symlink out of `launch_step` so this is no longer necessary (in this particular case, that is).

smartsim/_core/control/controller.py
AlyssaCote: `_launch_step` has retired from the business of doing symlinks

smartsim/_core/control/controller.py
AlyssaCote: `max-locals` got bumped to 25, so this is no longer an issue and I'll take out the comment!

PR #549, opened 2024-04-15
Fix telemetry monitor listener registration timeline

ankona:
Reorder experiment startup to ensure telemetry monitor registers event listeners prior to launching entities.


MattToast (approved):
LGTM!! Thanks for taking care of this!!

PR #550, opened 2024-04-15
Dragon status update fix

ankona:
## Fix a defect in retrieving status updates for the dragon launcher. 

Pre-dragon launchers used the task/step name to retrieve updates while the dragon launcher uses the `task_id`. This fix ensures that the name for dragon tasks is mapped appropriately.


ankona:
incorrect base

PR #553, opened 2024-04-16
set correct value for curve server public key

ankona:


al-rigazzi (approved):
LGTM thanks for this hot fix!

PR #554, opened 2024-04-17
checkpoint

ankona:


tests/on_wlm/test_dragon.py
```diff
@@ -1,6 +1,6 @@
 # BSD 2-Clause License
 #
-# Copyright (c) 2021-2023, Hewlett Packard Enterprise
+# Copyright (c) 2021-2024, Hewlett Packard Enterprise
```
al-rigazzi, line 3: I'm so old school.

al-rigazzi (approved):
LGTM, this will save us a lot of time in tests!

al-rigazzi (approved):
Still LGTM!

PR #557, opened 2024-04-18
Avoid setting `task_id` on a managed step in telemetry monitor

ankona:
Ensures that a managed step-mapping doesn't include a `task_id`.

The telemetry monitor exhibits a defect where it logs errors from a task manager, even with only managed tasks being monitored for updates:

image

This fix modifies the telemetry monitor to not set a `task_id` when adding items to the `step_mapping` collection. This avoids triggering lookups for unmanaged processes.

image

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -18,6 +18,7 @@ To be released at some future point in time
 
 Description
 
+- Fix telemetry monitor logging errrors for task history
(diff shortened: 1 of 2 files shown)


MattToast (approved):
Nice and clean fix! Thanks for the quick turn around and LGTM!!

PR #558, opened 2024-04-18
Upgrade ubuntu to 22.04

AlyssaCote:
After dropping python 3.8, there's been some issues with ubuntu versions  for the smartsim dev docs. Upgrading ubuntu to 22.04 is necessary now that we've dropped support for python 3.8 and should solve the problem.


ashao (approved):
Thanks for taking care of this!

PR #559, opened 2024-04-19
Fix failing and noisy WLM test

MattToast:
Fixes:
- `tests/backends/test_onnx.py::test_sklearn_onnx`
  - Correctly set number of tasks when not using the local launcher
  - Makes sure the DB is not left running on test failure
- `tests/full_wlm/test_generic_orc_launch_batch.py::test_launch_cluster_orc_reconnect`
  - Look for pickle file under the orchestrator path rather than the test dir
  - Makes sure that DB is cleaned up correctly on test failure

Quiets:
- `tests/on_wlm/test_het_job.py`
  - All experiments under the test module are given an explict test path


ashao (approved):
Thanks for getting these fixed up

PR #560, opened 2024-04-19
Remove defensive regexp in .gitignore

AlyssaCote:
In this PR I removed the defensive regexp in `.gitignore` and added `test_dir` to the tests that were writing to the `cwd` instead of the `test_output` directory.


ankona (approved):
LGTM!

PR #561, opened 2024-04-22
Auto generate typehints into documentation

amandarichardsonn:
This PR removes :type: and :rtype: driectives from function docstrings as well as implements the sphinx-autodoc-typehints extension.


AlyssaCote (approved):
The docs look great! Thanks for taking care of this! Don't forget to update the changelog, but other than that, LGTM!

PR #564, opened 2024-04-22
Force typing_extensions==4.6.1 in doc build

ashao:
Tensorflow requires that typing_extensions<=4.6.0, however this cuases the sphinx build process to fail due to an error importing sphinx_tabs. This is potentially a misleading error because sphinx_tabs itself does not use typing extensions, but the problem nevertheless exists even when testing other versions of Sphinx. To allow the deploy_dev_docs action to complete, this modifies the Dockerfile used to build the docs to ensure that
typing_extensions==4.6.1 (which is the lowest version that does not throw pip resolution errors) is present in the python environment prior to the build.


doc/changelog.rst
```diff
@@ -116,6 +119,7 @@ Detailed Notes
   handler. SmartSim will now attempt to kill any launched jobs before calling
   the previously registered signal handler. (SmartSim-PR535_)
 
+.. _SmartSim-PR561: https://github.com/CrayLabs/SmartSim/pull/561
```
amandarichardsonn, line 122: I believe this might be PR 564 instead of 561!

amandarichardsonn (approved):
LGTM! Just one small fix

PR #567, opened 2024-04-23
Adapt tests to reuse Orchestrator

ashao:
Tests which needed to launch an Orchestrator were spinning up and shutting down their own instances. This led to a number of cases where a single test failing would cascade into failures of other tests. Additionally, this also meant that a significant amount of time in the tests was spent waiting for Orchestrators to launch.

This PR adds a session-scoped fixture that returns an Orchestrator. Most tests which use an Orchestrator have been updated to use this fixture; the remaining for various reasons still need to spin up their own (for example the multiple database tests need to have a named Orchestrator).

This PR is not 100% complete with the following to be done
- Add another fixture that flushes the database after every test
- Undergo style checks
- Fix failing tests

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -671,11 +671,10 @@ class ColoUtils:
             colo_settings.set_tasks(1)
             colo_settings.set_nodes(1)
 
-        port = port if port is not None else _find_free_port(test_ports)
         colo_model = exp.create_model(colo_model_name, colo_settings)
 
         if db_type in ["tcp", "deprecated"]:
-            db_args["port"] = port
+            db_args["port"] = port if port is not None else _find_free_port(test_ports)
             db_args["ifname"] = "lo"
         if db_type == "uds" and colo_model_name is not None:
             tmp_dir = tempfile.gettempdir()
--- a/conftest.py
+++ b/conftest.py
@@ -926,9 +926,11 @@ def _create_and_launch_database(
     hostlist: t.Optional[t.List[str]],
     port: int,
 ):
+    exp_path = pathlib.Path(test_output_root, exp_name)
+    exp_path.mkdir()
     exp = Experiment(
         exp_name,
-        exp_path=str(pathlib.Path(test_output_root, exp_name)),
+        exp_path=str(exp_path),
         launcher=launcher,
     )
     orc = exp.create_database(
--- a/conftest.py
+++ b/conftest.py
@@ -875,6 +875,60 @@ class CountingCallable:
         return self._details
 
 
+# @pytest.fixture(scope="function")
+# def prepare_db_fixture():
+#     db_constructors = {
+#         "local_db":local_db,
+#         "single_db":single_db,
+#         "clustered_db":clustered_db
+#     }
+#     def _prepare_db_fixture(db_fixture, db_type, wlmutils):
+#         try:
+#             db_up = db_fixture.is_active()
+#         except:
+#             db_up = False
+#         if not db_up:
+#             return db_constructors[db_type](wlmutils)
+#         else:
+#             # TODO: Connect a client and flush the database
+#             return db_fixture
+#     return _prepare_db_fixture
+
+global_local_db = None
+
+@pytest.fixture(scope="function")
+def factory_experiment(test_dir):
+    # def _factory_experiment(*args, **kwargs):
+    return smartsim.Experiment("foo", exp_path=test_dir)
+    # return _factory_experiment
+
+@pytest.fixture(scope="function")
+#def prepare_db_fixture(local_db, construct_local_db, wlmutils):
+def prepare_db_fixture(factory_experiment):
+    global global_local_db
+    try:
+        db_up = global_local_db.is_active()
+    except:
+        db_up = False
+    if not db_up:
+        global_local_db = _create_and_launch_database2(
+            "local_db_fixture",
+            launcher="local",
+            num_nodes=1,
+            interface="lo",
+            hostlist=None,
+            port=_find_free_port(tuple(reversed(test_ports))),
+        )
+    # yield global_local_db
+    factory_experiment.reconnect_orchestrator(global_local_db.checkpoint_file)
+    return global_local_db
+    # return global_local_db
+
+@pytest.fixture(scope="session")
+def construct_local_db(wlmutils):
+    global global_local_db
+    yield global_local_db
+
 @pytest.fixture(scope="session")
 def local_db(wlmutils):
     with _create_and_launch_database(
@@ -943,3 +997,26 @@ def _create_and_launch_database(
         yield orc
     finally:
         exp.stop(orc)
+
+def _create_and_launch_database2(
+    exp_name: str,
+    launcher: str,
+    num_nodes: int,
+    interface: str,
+    hostlist: t.Optional[t.List[str]],
+    port: int,
+):
+    exp_path = pathlib.Path(test_output_root, exp_name)
+    exp_path.mkdir(exist_ok=True)
+    exp = Experiment(
+        exp_name,
+        exp_path=str(exp_path),
+        launcher=launcher,
+    )
+    orc = exp.create_database(
+        port, batch=False, interface=interface, hosts=hostlist, db_nodes=num_nodes
+    )
+    exp.generate(orc, overwrite=True)
+    exp.start(orc)
+    print("NEW ORCEHSTARAT")
+    return orc
--- a/conftest.py
+++ b/conftest.py
@@ -881,7 +881,7 @@ class CountingCallable:
 database_registry = defaultdict(lambda: None)
 
 @pytest.fixture(scope="function")
-def local_experiment(test_dir, wlmutils):
+def local_experiment(test_dir):
     """Create a default experiment that uses the requested launcher"""
     name = pathlib.Path(test_dir).stem
     return smartsim.Experiment(name, exp_path=test_dir, launcher="local")
@@ -890,7 +890,11 @@ def local_experiment(test_dir, wlmutils):
 def wlm_experiment(test_dir, wlmutils):
     """Create a default experiment that uses the requested launcher"""
     name = pathlib.Path(test_dir).stem
-    return smartsim.Experiment(name, exp_path=test_dir, launcher="local")
+    return smartsim.Experiment(
+        name,
+        exp_path=test_dir,
+        launcher=wlmutils.get_test_launcher()
+    )
 
 def _cleanup_db(name):
     global database_registry
--- a/conftest.py
+++ b/conftest.py
@@ -93,7 +93,6 @@ built_mpi_app = False
 test_hostlist = None
 has_aprun = shutil.which("aprun") is not None
 
-
 def get_account() -> str:
     return test_account
 
@@ -878,16 +877,16 @@ class CountingCallable:
 
 ## Reuse database across tests
 
-database_registry = defaultdict(lambda: None)
+database_registry: t.DefaultDict[str, t.Optional[Orchestrator]] = defaultdict(lambda: None)
 
 @pytest.fixture(scope="function")
-def local_experiment(test_dir):
+def local_experiment(test_dir: str) -> smartsim.Experiment:
     """Create a default experiment that uses the requested launcher"""
     name = pathlib.Path(test_dir).stem
     return smartsim.Experiment(name, exp_path=test_dir, launcher="local")
 
 @pytest.fixture(scope="function")
-def wlm_experiment(test_dir, wlmutils):
+def wlm_experiment(test_dir: str, wlmutils: WLMUtils) -> smartsim.Experiment:
     """Create a default experiment that uses the requested launcher"""
     name = pathlib.Path(test_dir).stem
     return smartsim.Experiment(
@@ -896,7 +895,7 @@ def wlm_experiment(test_dir, wlmutils):
         launcher=wlmutils.get_test_launcher()
     )
 
-def _cleanup_db(name):
+def _cleanup_db(name: str) -> None:
     global database_registry
     db = database_registry[name]
     if db and db.is_active():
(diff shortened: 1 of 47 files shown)


conftest.py
```diff
@@ -227,7 +229,27 @@ def _reset():
 )
 
 
-@pytest.fixture
+def _reset_signal(signalnum: int):
```
ashao, line 232: I think this must have come in as a weird merge commit. Will delete

conftest.py
```diff
@@ -244,7 +266,12 @@ def get_test_launcher() -> str:
 
     @staticmethod
     def get_test_port() -> int:
-        return test_port
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
```
ankona, line 269: Depending on when this gets merged, you might be able to swap to using [`smartsim._core.utils.network.find_free_port`](https://github.com/CrayLabs/SmartSim/blob/dragon_launcher/smartsim/_core/utils/network.py#L107-L125)

conftest.py
```diff
@@ -859,3 +780,44 @@ def num_calls(self) -> int:
     @property
     def details(self) -> t.List[t.Tuple[t.Tuple[t.Any, ...], t.Dict[str, t.Any]]]:
         return self._details
+
+@pytest.fixture(scope="session")
```
ankona, line 784: 👍 for use of scope.

smartsim/_core/config/config.py
```diff
@@ -180,6 +180,10 @@ def test_num_gpus(self) -> int:  # pragma: no cover
     def test_port(self) -> int:  # pragma: no cover
         return int(os.environ.get("SMARTSIM_TEST_PORT", 6780))
 
+    @property
+    def test_num_ports(self) -> int:  # pragma: no cover
```
ankona, line 184: 1. I'm really disliking the mixing of test and production configuration in `config.py`. It might be better to set test values in `conftest.py` - or in this case, right in the fixture. 

2. Why do we limit ourselves here to 10 available ports? It could be 100 or 1000 if you're trying to avoid an infinite loop. If we actually need 10 databases and there is a port or two unavailable, the tests just start failing.

smartsim/database/orchestrator.py
```diff
@@ -356,10 +357,13 @@ def is_active(self) -> bool:
         :return: True if database is active, False otherwise
         :rtype: bool
         """
-        if not self._hosts:
+        if not self.hosts:
             return False
```
MattToast, line 361: Is the try catch needed here? It looks like the hosts attr (for some reason) can throw this `SSDBFilesNotParseable` error

MattToast (commented):
Two small questions about intent that I wanted to write down so that I could cover them later while I pick up this PR

smartsim/_core/control/controller.py
```diff
@@ -773,10 +787,12 @@ def _orchestrator_launch_wait(self, orchestrator: Orchestrator) -> None:
...
                 # _jobs.get_status acquires JM lock for main thread, no need for locking
                 statuses = self.get_entity_list_status(orchestrator)
-                if all(stat == SmartSimStatus.STATUS_RUNNING for stat in statuses):
+                if (
+                    all(stat == SmartSimStatus.STATUS_RUNNING for stat in statuses)
+                    and orchestrator.is_active()
+                ):
                     ready = True
                     # TODO remove in favor of by node status check
-                    time.sleep(CONFIG.jm_interval)
```
MattToast, line 779: Why was this sleep removed?

tests/test_experiment.py
```diff
@@ -77,11 +77,6 @@ def test_ensemble_no_name():
         )
 
 
-def test_bad_exp_path() -> None:
-    with pytest.raises(NotADirectoryError):
-        exp = Experiment("test", "not-a-directory")
```
MattToast, line 82: Why was this test removed?

smartsim/experiment.py
```diff
@@ -148,10 +148,6 @@ def __init__(
         """
         self.name = name
         if exp_path:
-            if not isinstance(exp_path, str):
-                raise TypeError("exp_path argument was not of type str")
-            if not osp.isdir(osp.abspath(exp_path)):
-                raise NotADirectoryError("Experiment path provided does not exist")
```
MattToast, line 154: Similar question here: why did this PR touch this code path?

conftest.py
```diff
@@ -244,7 +247,12 @@ def get_test_launcher() -> str:
 
     @staticmethod
     def get_test_port() -> int:
-        return test_port
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            for port in range(test_port, test_port + test_num_ports):
+                result = sock.connect_ex(("127.0.0.1", port))
+                if result:
+                    return port
```
MattToast, line 254: I'm a little confused on the logic here: It looks like the idea is that this socket is going to try to connect to this addr/port, and if there is some error we assume that it is because there is no socket bound there to accept the connection, and therefore the port must be free.

However, errors could come from a variety of problems (e.g. there is a socket bound but it is not listening, there is a socket bound and it is listening but it ran out of buffer space for pending new unaccepted connections, etc.) where the port may not actually be available.

Should this be trying to bind instead??

MattToast: Switched to `bind` to avoid port conflicts

smartsim/database/orchestrator.py
```diff
@@ -356,10 +357,13 @@ def is_active(self) -> bool:
...
 
-        return db_is_active(self._hosts, self.ports, self.num_shards)
+        try:
+            return db_is_active(self._hosts, self.ports, self.num_shards)
+        except SSDBFilesNotParseable:
+            return False
```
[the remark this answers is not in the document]
MattToast: Leftover for a previous refactor attempt that was reverted. The try catch block was correctly moved to wrap `self.hosts`

tests/test_experiment.py
MattToast: This is something we want to support, but was deemed out of scope for this PR. Reverting this change

smartsim/experiment.py
MattToast: Re-adding this test per reverted change

conftest.py
MattToast: Current plan is this PR will likely go in before the dragon work, so this fn is not yet defined, but we should consider using it as soon as the dragon work is merged!

smartsim/_core/config/config.py
MattToast: 1. Totally agree, we are already pushing the number of public methods on the `Config` class! I'll write up a ticket to make multiple configuration class for different areas of configuration!!
1. After talking with Andrew, it seems like you are right! The 10 port limit was deliberately chosen to prevent an infinite loop, and 10 was chosen slightly arbitrarily just to keep the iteration to an order of 10s. Sounds like there is some wiggle room to be debated, but I actually kinda like that it doesn't check an infinite number of ports as if orcs start leaking from tests we have a way to back track to the offending test. Willing to be debated on this point though!!

AlyssaCote (changes requested):
This is awesome! And I really like how it was implemented. I have a couple questions and there's a test file that needs some love, but that's it!

smartsim/_core/config/config.py
ankona: I'm ok w/any number that feels less likely to randomly break than 10 - anything between 25 and 1000 is still arbitrary but also feels safer!

ankona (approved):
This PR is like a dream come true. I can't wait for my speedy test results.

conftest.py
```diff
@@ -396,65 +367,6 @@ def choose_host(rs: RunSettings) -> t.Optional[str]:
...
-    exp_name = request.function.__name__
-    exp = Experiment(exp_name, launcher="local", exp_path=test_dir)
-    db = Orchestrator(port=wlmutils.get_test_port(), interface="lo")
-    db.set_path(test_dir)
-    exp.start(db)
-
-    yield db
-    # pass or fail, the teardown code below is ran after the
-    # completion of a test case that uses this fixture
-    exp.stop(db)
```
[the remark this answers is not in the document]
ankona: @AlyssaCote The fixtures [have scopes applied](https://github.com/CrayLabs/SmartSim/pull/567/files#diff-a31c7ed5d35f5ed8233994868c54d625b18e6bacb6794344c4531e62bd9dde59R246) that will ensure they only get called once per run. That lets pytest return the yielded result to every test until the scope expires!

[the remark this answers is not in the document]
AlyssaCote: @ankona Thanks for the explanation!! Makes sense!

smartsim/_core/config/config.py
MattToast: That makes sense, and I like the reasoning for a higher number! I'll bump it to 25, and we can increase it in future as we see fit!

conftest.py
```diff
@@ -653,7 +565,7 @@ def setup_test_colo(
         db_args: t.Dict[str, t.Any],
         colo_settings: t.Optional[RunSettings] = None,
         colo_model_name: str = "colocated_model",
-        port: int = test_port,
+        port: int = test_ports[0],
```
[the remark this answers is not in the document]
MattToast: We probably should, this was 100% me being lazy and keeping the existing impl. Will change the type of this param to `int | None` where `None` is a sentinel value to find a free port!

tests/test_fixtures.py
```diff
@@ -0,0 +1,42 @@
...
+from smartsim.error import SmartSimError
+from smartsim.error.errors import SSUnsupportedError
+
+# The tests in this file belong to the group_a group
+pytestmark = pytest.mark.group_a
+
+
+def test_single_db_fixture(single_db):
+    experiment = Experiment("test_single_fixture")
+    experiment.reconnect_orchestrator(single_db.checkpoint_file)
```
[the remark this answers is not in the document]
MattToast: Uhhh... good question! I've never written testing code for testing code before. I'll defer to @ashao on this one for intention.

smartsim/database/orchestrator.py
```diff
@@ -388,6 +396,15 @@ def _redis_exe(self) -> str:
     def _redis_conf(self) -> str:
         return CONFIG.database_conf
 
+    @property
+    def checkpoint_file(self) -> str:
+        """Get the path to the checkpoint file for this Orchestrator
+
+        :return: Path to the checkpoint file if it exists, otherwise a None
+        :rtype: str or None
```
amandarichardsonn, line 404: No need to add the rtype here I believe!

smartsim/database/orchestrator.py
```diff
@@ -388,6 +396,15 @@ def _redis_exe(self) -> str:
     def _redis_conf(self) -> str:
         return CONFIG.database_conf
 
+    @property
+    def checkpoint_file(self) -> str:
```
amandarichardsonn, line 400: I would add this property in doc/api/smartsim_api around line 389 unless this function shouldnt be user facing, then I would add a _ infront or else it will show up on readthedocs

amandarichardsonn: do you mind adding the batch property as well?

smartsim/error/errors.py
```diff
@@ -87,6 +87,12 @@ class SSDBIDConflictError(SmartSimError):
     """
 
 
+class SSDBFilesNotParseable(SmartSimError):
```
amandarichardsonn, line 90: Thanks for adding a description! I am lost without them fr

conftest.py
```diff
@@ -244,7 +260,9 @@ def get_test_launcher() -> str:
 
     @staticmethod
     def get_test_port() -> int:
-        return test_port
+        # TODO: Ideally this should find a free port on the correct host(s),
```
amandarichardsonn, line 263: Is a ticket created for this? If not, possibly might forget!

doc/changelog.md
```diff
@@ -102,7 +102,7 @@ Detailed Notes
     Torch will unconditionally try to link in this library, however
     fails because the linking flags are incorrect.
     ([SmartSim-PR538](https://github.com/CrayLabs/SmartSim/pull/538))
--   Change type_extension and pydantic versions in readthedocs
+-   Change typing\_extensions and pydantic versions in readthedocs
```
amandarichardsonn, line 105: I notice two updates to the Detailed Notes section but one update to the Description!

smartsim/_core/control/controller.py
```diff
@@ -775,8 +787,7 @@ def _orchestrator_launch_wait(self, orchestrator: Orchestrator) -> None:
                 statuses = self.get_entity_list_status(orchestrator)
                 if all(stat == SmartSimStatus.STATUS_RUNNING for stat in statuses):
                     ready = True
-                    # TODO remove in favor of by node status check
-                    time.sleep(CONFIG.jm_interval)
+                    # TODO: Add a node status check
```
amandarichardsonn, line 790: ticket created 👀

amandarichardsonn (approved):
LGTM, will take a second pass in the morning! Just some super small comments

smartsim/database/orchestrator.py
MattToast: `:rtype:` is gone!

doc/changelog.md
MattToast: Yeah the change to "Description" and one of the "Detailed Notes" edits were just minor typo/grammar things.

The actual "Detailed Note" I added I figured fell under the category of "Minor enhancements to test suite" which is already listed in the "Description" section, so I just added something the details. LMK if you think this warrants its on entry in the "Description" and I can add something there!!

smartsim/database/orchestrator.py
MattToast: Added!

smartsim/_core/control/controller.py
MattToast: [Added](https://github.com/CrayLabs/Poseidon/issues/690)!!

conftest.py
MattToast: [Added](https://github.com/CrayLabs/Poseidon/issues/691)!!

tests/test_fixtures.py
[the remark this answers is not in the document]
ashao: How meta do we want to go? But yes the intention of this was to make sure that fixture works as expected. I'm imagining us bashing our head against something related to the fixture and just being able to do:
```
pytest test_fixtures.py::test_single_db_fixture
```
and save us some time down the line. Happy to be argued down, but in general I think it's not crazy to have a some sanity checking of testing code since it helps to debug tests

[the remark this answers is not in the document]
AlyssaCote: I definitely don't hate the idea of being able to sanity check that the fixtures are working as expected.

ashao:
Note that Al noted a defect which causes cascading failures if the database was stopped on purpose or because of a database error. A fix is coming in

conftest.py
```diff
@@ -856,3 +771,132 @@ def num_calls(self) -> int:
...
+            exp_path=str(exp_path),
+            launcher=launcher,
+        )
+        orc = exp.create_database(
+            port, batch=False, interface=interface, hosts=hostlist, db_nodes=num_nodes
+        )
+        exp.generate(orc, overwrite=True)
+        exp.start(orc)
+        global database_registry
+        database_registry[name] = orc
```
AlyssaCote, line 874: Does the `database_registry` hold the name of the experiment instead of the database?

AlyssaCote (commented):
This looks really great! I have a few general questions rather than specific fix requests, but once those are cleared up this should be good to go!

conftest.py
```diff
@@ -856,3 +771,132 @@ def num_calls(self) -> int:
...
+def local_experiment(test_dir, wlmutils):
+    """Create a default experiment that uses the requested launcher"""
+    name = pathlib.Path(test_dir).stem
+    return smartsim.Experiment(name, exp_path=test_dir, launcher="local")
+
+@pytest.fixture(scope="function")
+def wlm_experiment(test_dir, wlmutils):
+    """Create a default experiment that uses the requested launcher"""
+    name = pathlib.Path(test_dir).stem
+    return smartsim.Experiment(name, exp_path=test_dir, launcher="local")
```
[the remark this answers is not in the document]
ashao: Ahhh nice catch! Thanks for that

tests/test_collectors.py
```diff
@@ -42,6 +42,8 @@
 # The tests in this file belong to the group_a group
 pytestmark = pytest.mark.group_a
 
+PrepareDB = t.Callable[[dict], smartsim.experiment.Orchestrator]
```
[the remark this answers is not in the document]
ashao: Mainly because @ankona typehinted everything else in this file. We should probably do the same for the other tests at some point

conftest.py
ashao: It's supposed to be the identifier for a specific "type" of orchestrator that a test requests. Since we can't launch a database outside of an experiment, the two are synonymous right now.

AlyssaCote: Makes sense!

conftest.py
```diff
@@ -856,3 +771,136 @@ def num_calls(self) -> int:
     @property
     def details(self) -> t.List[t.Tuple[t.Tuple[t.Any, ...], t.Dict[str, t.Any]]]:
         return self._details
+
+## Reuse database across tests
+
+database_registry = defaultdict(lambda: None)
+
+@pytest.fixture(scope="function")
+def local_experiment(test_dir):
```
AlyssaCote, line 780: Totally missed this on the first pass through, but can we add type annotations to the new functions throughout this file?

AlyssaCote (approved):
LGTM!!

PR #570, opened 2024-04-25
Set GH_TOKEN environment variable to use Github CLI in workflow

amandarichardsonn:
This PR updates the authetication used in the release workflow from a developer created token to the GH_TOKEN environment variable.


MattToast (approved):
LGTM!!

PR #589, opened 2024-05-16
Update tutorials and tutorial containers

al-rigazzi:
This PR udpates tutorials to the new SmartSim and SmartRedis APIs.

In addition, the tutorial containers' Docker files were also updated, to prevent an error caused by IPython.


amandarichardsonn (approved):
LGTM, compiles with no errors

PR #598, opened 2024-05-22
Fix build error caused by use of deprecated pkg_resources

ankona:
Fixes a build error encountered after the latest update of `setuptools`.

### Additional Included Changes

- Fixed non-working links causing documentation build to fail validation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -185,6 +185,7 @@ deps.append("smartredis>={}".format(versions.SMARTREDIS))
 
 extras_require = {
     "dev": [
+        "packaging>=24.0",
         "black==24.1a1",
         "isort>=5.6.4",
         "pylint>=2.10.0,<3",
--- a/setup.py
+++ b/setup.py
@@ -166,6 +166,7 @@ class BinaryDistribution(Distribution):
 
 # Define needed dependencies for the installation
 deps = [
+    "packaging>=24.0",
     "psutil>=5.7.2",
     "coloredlogs>=10.0",
     "tabulate>=0.8.9",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -185,6 +185,7 @@ def has_ext_modules(_placeholder):
 
 extras_require = {
     "dev": [
+        "packaging>=24.0",
```
MattToast, line 188: Since we actually import this `smartsim._core._install.buildenv` (and thus `packaging`) when a user runs `smart build` can we actually make this a regular `install_requires` dependency in the `deps` list? Otherwise I think there might be a super weird corner case when a user runs `smart build` when installing without the optional `ml` dependencies.

MattToast: Can confirm that this very weird sequence of actions will lead to an error:
```sh
$ pip install .  # Very intentionally installing smartsim w/o `dev` or `ml` deps
# Abridged
Successfully built smartsim
# Abrigdged
Installing collected packages:  # Abridged

$ pip list | grep packaging
# No output is returned
# This is not a problem when installing with `dev` because we require it there
# This is not a problem when installing with `ml` because TF (and maybe scipy?) have a dependency on `packaging`

$ smart build
# Abridged
  File "/Users/mrdro/repos/ssimdev/ss/venv/lib/python3.11/site-packages/smartsim/_core/_install/buildenv.py", line 38, in <module>
    from packaging.version import InvalidVersion, Version, parse
ModuleNotFoundError: No module named 'packaging'

MattToast (changes requested):
I ran into one small problem when installing without the `ml` or `dev` optional dependencies, but otherwise LGTM!

setup.py
ankona: Thank you, sir. Fixed!

MattToast (approved):
LGTM pending tests!! Thanks for tracking this down!

PR #605, opened 2024-05-29
Compound Entity and Ensemble

MattToast:
Creates a basic implementation for a `CompundEntity` base class that can be used to build a launchable containing multiple processes (`Job`s, `JobGroup`, etc) each with their own settings from a single `LaunchSettings` instance.

This class is then subclassed to create an `Ensemble` class that mimics the original implementation: I takes in an `Application` (`Model`) and then maps settings/params/input files over the collection before creating the launchable jobs.


smartsim/entity/entity.py
```diff
@@ -118,3 +121,10 @@ def set_path(self, path: str) -> None:
 
     def __repr__(self) -> str:
         return self.name
+
+
+class CompoundEntity(abc.ABC):
```
amandarichardsonn, line 126: Add docstrings here :P

smartsim/entity/strategies.py
```diff
@@ -25,44 +25,97 @@
...
+
+import functools
+import itertools
 import random
 import typing as t
-from itertools import product
+
+from smartsim.error import errors
+
+TPermutationStrategy = t.Callable[
```
amandarichardsonn, line 38: Do you mind commenting what this is for?

smartsim/entity/strategies.py
```diff
@@ -25,44 +25,97 @@
...
+    [t.Mapping[str, t.Sequence[str]], int], list[dict[str, str]]
+]
+
+_REGISTERED_STRATEGIES: t.Final[dict[str, TPermutationStrategy]] = {}
```
amandarichardsonn, line 42: as well as here?

smartsim/entity/strategies.py
```diff
@@ -25,44 +25,97 @@
...
+
+
+def _register(name: str) -> t.Callable[
```
amandarichardsonn, line 45: Do you mind writing docstrings for all functions youve added in this file?

smartsim/entity/ensemble.py
```diff
@@ -117,19 +112,17 @@ def _initialize_entities(self, **kwargs: t.Any) -> None:
 
         :raises UserStrategyError: if user generation strategy fails
         """
-        strategy = self._set_strategy(kwargs.pop("perm_strat"))
+        strategy = resolve_strategy(kwargs.pop("perm_strat"))
```
amandarichardsonn, line 115: We are retrieving the perm_strat from kwargs passed in but we have a perm_strat pass to super.__init__, can we retrieve it via the attribute? This is probably more of a miss from my understanding

smartsim/entity/_new_ensemble.py
```diff
@@ -0,0 +1,93 @@
...
+# https://github.com/CrayLabs/SmartSim/pull/603
+from smartsim.entity import _mock, entity, strategies
+from smartsim.entity.files import EntityFiles
+from smartsim.entity.model import Application
+
+
+# TODO: If we like this design, we need to:
+#         1) Move this to the `smartsim._core.entity.ensemble` module
+#         2) Decide what to do witht the original `Ensemble` impl
+class Ensemble(entity.CompoundEntity):
```
amandarichardsonn, line 47: Is this class replacing the original Ensemble class?

smartsim/entity/_new_ensemble.py
```diff
@@ -0,0 +1,93 @@
...
+        self.name = name
+        self.exe = os.fspath(exe)
+        self.exe_args = list(exe_args) if exe_args else []
+        self.files = copy.deepcopy(files) if files else EntityFiles()
+        self.parameters = dict(parameters) if parameters else {}
+        self.permutation_strategy = permutation_strategy
+        self.max_permutations = max_permutations
+        self.replicas = replicas
+
+    def _create_applications(self) -> tuple[Application, ...]:
```
amandarichardsonn, line 68: do you mind writing a doc string here?

amandarichardsonn (changes requested):
Nice work!!

smartsim/entity/_new_ensemble.py
MattToast: That's the plan! I just didn't want to remove the original yet in case anyone was using it for reference. I'd be more than willing to move this implementation to `smartsim.entity.ensemble`, delete the `# TODO` comment and remove the original if that's okay with you!

smartsim/entity/entity.py
MattToast: Can do!

smartsim/entity/strategies.py
MattToast: Absolutely!

smartsim/entity/strategies.py
MattToast: Of course!!

smartsim/entity/strategies.py
MattToast: Can do!

smartsim/entity/ensemble.py
MattToast: Not a miss of your understanding, it is just very confusingly implemented!

This is a relic from the days before SmartSim was not rigidly using `abc.ABC`s so ensembles inherit from this "psuedo-abstract" `EntityList` class and the idea is that users would implement this `_initialize_entities` function to finish initialization of "abstract" properties.

More than willing move this to named kwarg in this method! Hopefully we will end up trashing this implementation entierly, but if for some reason we do decide to keep it, the "correct" thing to do would be to make `EntityList` a true `abc.ABC`, mark our abstract attrs, and remove the constructor from the interface.

smartsim/entity/_new_ensemble.py
MattToast: One day, I will learn to explain my work without being coerced by a reviewer, but, alas, that day is not today, hahahaha

Can do!!

smartsim/entity/_new_ensemble.py
```diff
@@ -0,0 +1,109 @@
...
+from smartsim.entity.strategies import ParamSet
+from smartsim.launchable.job import Job
+
+if t.TYPE_CHECKING:
+    from smartsim.settings.launchSettings import LaunchSettings
+
+
+# TODO: If we like this design, we need to:
+#         1) Move this to the `smartsim._core.entity.ensemble` module
+#         2) Decide what to do witht the original `Ensemble` impl
```
MattToast, line 46: Is this something I should address in this PR? Is anyone still using the old impl? Can I safely delete it?

smartsim/entity/ensemble.py
```diff
@@ -200,7 +191,8 @@ def add_application(self, application: Application) -> None:
         """
         if not isinstance(application, Application):
             raise TypeError(
-                f"Argument to add_application was of type {type(application)}, not Application"
+                f"Argument to add_application was of type {type(application)}, "
+                " not Application"
```
MattToast, line 195: 🤦 
```suggestion
                f"Argument to add_application was of type {type(application)}, "
                "not Application"
```

smartsim/entity/_new_ensemble.py
amandarichardsonn: I am not using it, I do not think Julia is either, I say remove

smartsim/entity/_new_ensemble.py
```diff
@@ -0,0 +1,109 @@
...
+        replicas: int = 1,
+    ) -> None:
+        self.name = name
+        self.exe = os.fspath(exe)
+        self.exe_args = list(exe_args) if exe_args else []
+        self.exe_arg_parameters = (
+            copy.deepcopy(exe_arg_parameters) if exe_arg_parameters else {}
+        )
+        self.files = copy.deepcopy(files) if files else EntityFiles()
+        self.file_parameters = dict(file_parameters) if file_parameters else {}
```
amandarichardsonn, line 71: do you mind adding a deep_copy for the file_parameters, just to be consistent with exe_arg_parameters? That is a my bad MERP

amandarichardsonn: What are your thoughts on this, I am thinking that the attributes that do not need to be deepcopied are replicas, max_permutations and strategy, just bc they do not live on to Application and are not reused

amandarichardsonn: or hmm maybe deepcopy is just something to worry about for Application?

smartsim/entity/ensemble.py
```diff
@@ -32,27 +32,18 @@
 from tabulate import tabulate
 
 from .._core._install.builder import Device
```
amandarichardsonn, line 34: This file is about to be completely removed correct?

smartsim/_core/utils/helpers.py
```diff
@@ -122,7 +122,6 @@ def expand_exe_path(exe: str) -> str:
 
     # which returns none if not found
     in_path = which(exe)
-    print(f"hmm what is this: {in_path}")
```
amandarichardsonn, line 125: merp thank you this was def me

smartsim/launchable/job.py
```diff
@@ -24,13 +24,18 @@
...
+
+import typing as t
 from copy import deepcopy
 
 from smartsim._core.commands.launchCommands import LaunchCommands
-from smartsim.entity.entity import SmartSimEntity
 from smartsim.launchable.basejob import BaseJob
 from smartsim.settings import LaunchSettings
 
+if t.TYPE_CHECKING:
```
amandarichardsonn, line 36: Why is SmartSimEntity conditional now?

tests/test_ensemble.py
```diff
@@ -0,0 +1,242 @@
...
+
+import pytest
+
+from smartsim.entity import _mock
+from smartsim.entity._new_ensemble import Ensemble
+from smartsim.entity.strategies import ParamSet
+from smartsim.settings.launchCommand import LauncherType
+from smartsim.settings.launchSettings import LaunchSettings
+
+pytestmark = pytest.mark.group_a
```
amandarichardsonn, line 38: What is this: `pytestmark = pytest.mark.group_a`?

tests/test_ensemble.py
```diff
@@ -0,0 +1,242 @@
...
+def user_created_function(
+    file_params: t.Mapping[str, t.Sequence[str]],
+    exe_arg_params: t.Mapping[str, t.Sequence[t.Sequence[str]]],
+    n_permutations: int = 0,
+) -> list[ParamSet]:
+    return [ParamSet({}, {})]
+
+
+@pytest.fixture
+def mock_launcher_settings():
```
amandarichardsonn, line 53: This is probably something I should have done previously... BUT what if we used the test_launcher global in conftest? Instead of just Local

tests/test_ensemble.py
```diff
@@ -0,0 +1,242 @@
...
+    return LaunchSettings(LauncherType.Local, {}, {})
+
+
+def test_ensemble_user_created_strategy(mock_launcher_settings):
```
amandarichardsonn, line 57: Do you mind adding `test_path` to Ensemble.path? Im thinking this is needed but maybe not

amandarichardsonn (commented):
Lets go! Awesome work and thank you for fixing those type errors, some suggestions but feel free to say nah

smartsim/launchable/job.py
MattToast: It's there to prevent a circular import from occurring!

<details>
<summary> Info dump if you're curious about details</summary>

One of the problems with python is that it does not have a "true" compile step, and therefore doesn't have header files or really anyway to separate declaration from implementation. Usually this is a good thing IMO, but it becomes kind of a pain when type hinting python specifically.

For example consider the following modules:
```py
# my_module_a.py
from my_module_b import B

class A:
    def do_something_with_a_b(self, b: B) -> None: ...
```
```py
# my_module_b.py
from my_module_a import A

class B:
    def do_something_with_an_a(self, a: A) -> None: ...
```
and
```py
# main.py
from my_module_a import A
from my_module_b import B

def main():
    a = A()
    b = B()
    a.do_something_with_a_b(b)
    b.do_something_with_an_a(a)

if __name__ == "__main__":
    raise SystemExit(main())
```
We can see that this _should_ work, but when we actually try to run `python main.py` we get the following error:
```py
  File "/private/tmp/d/main.py", line 2, in <module>
    from my_module_a import A
  File "/private/tmp/d/my_module_a.py", line 2, in <module>
    from my_module_b import B
  File "/private/tmp/d/my_module_b.py", line 2, in <module>
    from my_module_a import A
ImportError: cannot import name 'A' from partially initialized module 'my_module_a' (most likely due to a circular import) (/private/tmp/d/my_module_a.py)
```
we can see that `my_module_a` requires `my_module_b` which requires `my_module_a` which leads to a circular dependency at run time.

BUT! We don't _actually_ need the dependency at runtime, we just need it for type checking, so by moving to imports into a `t.TYPE_CHECKING` block and changing the type hints to forward references(i.e. strings (which is what `from __future__ import annotations` does behind the scenes (for now ([see here if you're curious](https://docs.python.org/3/library/__future__.html#id2))))), we can use their declaration for static analysis, but ignore their implementation at run time.
```py
# my_module_a.py (revised)
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from my_module_b import B

class A:
    def do_something_with_a_b(self, b: "B") -> None: ...
    #             Note the forward ref ^^^
```
```py
# my_module_b.py (revised)
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from my_module_a import A

class B:
    def do_something_with_an_a(self, a: "A") -> None: ...
    #              Note the forward ref ^^^
```

and now running `python main.py` works as expected!
</details>

smartsim/entity/ensemble.py
MattToast: That's the plan! I'll delete this file and rename `_new_ensemeble.py` -> `ensemble.py` when I get rid of the old impl!

tests/test_ensemble.py
MattToast: It's a shortcut that `pytest` offers to turn this:
```py
@pytest.mark.group_a
def test_a():
    ...

@pytest.mark.group_a
def test_b():
    ...

@pytest.mark.group_a
def test_c()
```
into
```py
# all tests in this module are now a part of `group_a`

def test_a(): ...
def test_b(): ...
def test_c(): ...
```

Our CI got split up into `group_a`, `group_b`, `slow_tests` and `backends` a while back in an attempt to reduce the time it takes to evaluate.

Unfortunately, it means that in order for a test to get automatically run by the CI it needs to be manually added to one of those groups (something we haven't enforced super well in the past and at least part of the reason I would be okay with removing the groups from the CI).

For now, since all of our tests are going pretty fast, I have just been adding everything into `group_a` so that the CI will continuously run the checks on each commit.

tests/test_ensemble.py
MattToast: Great idea!!

tests/test_ensemble.py
MattToast: Makes sense! Will do!!

smartsim/launchable/job.py
amandarichardsonn: thanks!

smartsim/entity/ensemble.py
MattToast: Resolving as we have decided to remove the old impl of `Ensemble`

smartsim/entity/_new_ensemble.py
MattToast: I think the general thought process is that we should make copies of anything that is mutable when it is past to the ensemble constructor.

Because strings are immutable in python, a deep-copy of a `dict[str, str]` (or really any `MutableMapping[str, str]`) is functionally equivalent to making a shallow copy! I chose `dict` over `copy.copy` or `copy.deepcopy` simply so that any helper methods know the actual type of the container, and so that user know they could modify the collection after construction (should they choose)
```py
original_params = {"SOME": "PARAM"}
ens = Ensemeble(..., file_parameters=original_params, ...)

# Wait I need to change those params for some reason!!
# But know that I can because the attr is typed as a dict!
ens.file_parameters["SOME"] = "OTHER_PARAMETER"  # passes type check!
assert ens.file_parameters != original_params    # passes!
```

MattToast: To that end, I think we need:
- a deep copy of `files` as its a complex user defined class
- a deep copy of `exe_args_parameters` as its a container of containers of immutable
- shallow copies of `file_parameters` and `exe_args` as they are containers of immutables
- No copies of any other parameters as they are all immuatable

What do you think?

smartsim/entity/_new_ensemble.py
MattToast: Removed!

smartsim/entity/ensemble.py
MattToast: Old impl is gone!

smartsim/entity/model.py
```diff
@@ -73,19 +83,20 @@ def __init__(
...
         :param batch_settings: Launcher settings for running the individual
                                application as a batch job
+        :param files: Files to have available to the application
         """
         super().__init__(name, str(path), run_settings)
         self.exe = [expand_exe_path(exe)]
         # self.exe = [exe] if run_settings.container else [expand_exe_path(exe)]
         self.exe_args = exe_args or []
-        self.params = params or {}
+        self.params = params.copy() if params else {}
```
amandarichardsonn, line 92: why shallow copy of params here? and not deep? but deep for files?

tests/test_permutation_strategies.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def test_strategy_registration(monkeypatch):
+    monkeypatch.setattr(strategies, "_REGISTERED_STRATEGIES", {})
+    assert not strategies._REGISTERED_STRATEGIES
+
+    new_strat = lambda params, exe_args, nmax: []
+    strategies._register("new_strat")(new_strat)
+    assert strategies._REGISTERED_STRATEGIES == {"new_strat": new_strat}
+
```
amandarichardsonn, line 45: Just to check my understanding! 

Here you first make sure that _REGISTERED_STRAT is empty. next you define a lambda function (easier than actually writing a user function) then you register it. But it seems kinda funny to do `._register("new_strat")(new_strat)` - why the two ()() when _register just accpets (str)

amandarichardsonn (approved):
LGTM just 2 Qs

smartsim/entity/model.py
MattToast: Strings are immutable in python, a shallow copying is container of immutable objects is functionally equivalent to making deep copy because in order to change any of elements in the container it must be replaced wholesale!

 See [here](https://github.com/CrayLabs/SmartSim/pull/605#discussion_r1664601477) for my decision process for what I chose to deep/shallow/not copy, but I could be very easily convinced to make this a deep copy just for consistencies stake if you'd like!

tests/test_permutation_strategies.py
MattToast: ^^ Exactly correct!!

The reason for the two `()` calls is because `register` is a callable that returns a callable that actually does the registration. The registration is actually done during the second call. You can think of it as actually doing something like this

```py
_tmp_fn_to_add_a_new_strategy_under_the_name_new_strategy = strategies._register("new_strat")

reveal_type(_tmp_fn_to_add_a_new_strategy_under_the_name_new_strategy)
# reveals: `t.Callable[[PermutationStrategyType], PermutationStrategyType]`

assert strategies._REGISTERED_STRATEGIES == {}
# Passes! Notice that nothing has been added yet

new_strat = lambda p, e, n: []
return_from_tmp_fn = _tmp_fn_to_add_a_new_strategy_under_the_name_new_strategy(new_strat)

assert return_from_tmp_fn is new_strat
# Passes! Notice that the function is returned unchanged

assert strategies._REGISTERED_STRATEGIES == {"new_strat": new_strat}
# Passes! Notice that the key was only added after
# `_tmp_fn_to_add_a_new_strategy_under_the_name_new_strategy` was called
```

All of this is for the weird magic that let's us use this function like a decorator. This test is functionally testing this use case
```py
stategies._register("new_strat")
def new_strat(p, e, n):
    return []
```

PR #609, opened 2024-06-07
Fine Grain GPU Build Options

MattToast:
Updates to the SmartSim build process:
- SmartSim can now be built using Cuda version 11.8 or Cuda 12.1 by specify `smart build --device=cuda118` or `smart build --device=cuda121`. The original `smart build --device=gpu` will default to using Cuda 11.8.
- As a result of the previous change, SmartSim now requires C++17 and a minimum Cuda version of 11.8 in order to build Torch 2.1.0.


smartsim/_core/_install/builder.py
```diff
@@ -97,7 +100,34 @@ def from_str(cls, string: str, /) -> "Architecture":
...
+    def is_gpu(self) -> bool:
+        return self != type(self).CPU
+
+    def is_cuda(self) -> bool:
+        cls = type(self)
+        return self in (cls.CUDA118, cls.CUDA121)
+
+    def torch_suffix(self) -> str:
+        cls = type(self)
+        if self == cls.CPU:
```
ashao, line 124: Thoughts on changing this into a dictionary definition instead of multiple `if`/`return`

ashao (changes requested):
Looking good so far. Update the Tensorflow version when you can and let's make a cuda-12 feature branch

smartsim/_core/_install/builder.py
MattToast: Totally reasonable to me! I originally had code like this in a couple of places wanted exhaustiveness checking from `_assert_never`, and I just kinda naively moved it over.

Now these values are sitting right next to their definition I don't see a huge need for the check anymore. Will remove!

MattToast:
This PR is now targeting a [cuda-12-support](https://github.com/CrayLabs/SmartSim/tree/cuda-12-support) branch!

PR #611, opened 2024-06-11
Add Dockerfiles with CUDA support

ashao:
The existing dockerfiles did not have the necessary environments to enable GPU support for users wanting to run the tutorials. Two Dockerfiles (for CUDA 11.8 and 12.1) are provided. Note that this has only been tested on a workstation and no HPC support should be expected at this time.


MattToast (approved):
LGTM! Thanks for adding these!!

PR #614, opened 2024-06-12
Fix util-tests outputs appearing in root directory

ankona:
Fix unit tests that omit an experiment path and write to the root directory of the project.

## Secondary Changes

- Added typehints to enable successfully passing `mypy test_manifest.py`
- Refactored test using global variables in pytest parameterization to use a fixture

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -15,11 +15,13 @@ To be released at some future point in time
 
 Description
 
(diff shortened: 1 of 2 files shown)


MattToast (approved):
LGTM, pending tests! Thanks for handling this!!

PR #618, opened 2024-06-14
Add 'f' to f-strs

MattToast:
Add 'f' to intended f-strings for better error messages


ashao (approved):
Where's the PEP for automatic f-strings?! Surely that wouldn't cause any problems.

PR #619, opened 2024-06-20
Remove device attribute from schemas

AlyssaCote:
This PR removes `device` from the schemas, MessageHandler, and tests.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -15,6 +15,7 @@ Description
 
 - Added schemas and MessageHandler class for de/serialization of
   inference requests and response messages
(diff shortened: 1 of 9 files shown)


ankona:
There is a test [here](https://github.com/AlyssaCote/SmartSim/blob/remove_device/tests/mli/test_worker_manager.py#L147) that is likely silently failing because it provides an unexpected keyword arg. At minimum, you should just have to remove `device=expected_device,` but could also remove from the commented out tests in `test_integrated_torch_worker.py`, too

ankona (approved):
LGTM!

PR #621, opened 2024-06-24
Enable environment variable based configuration for ML Worker Manager

AlyssaCote:
This PR adds a `EnvironmentConfigLoader`, which loads and deserializes an `SSFeatureStore` and an `SSQueue` from environment variables for the ML Worker Manager.


.github/workflows/run_tests.yml
```diff
@@ -143,6 +143,13 @@ jobs:
           echo "SMARTSIM_LOG_LEVEL=debug" >> $GITHUB_ENV
           py.test -s --import-mode=importlib -o log_cli=true --cov=$(smart site) --cov-report=xml --cov-config=./tests/test_configs/cov/local_cov.cfg --ignore=tests/full_wlm/ ./tests/backends
 
+      # Run pytest (backends subdirectory)
```
ankona, line 146: Don't forget to update the copypasta'd comment to be correct for this new block (it is not the backends subdirectory).

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -28,10 +28,12 @@
 import typing as t
 
 import numpy as np
+from dragon.fli import FLInterface
```
ankona, line 31: We want to avoid having a dragon dependency loaded here so we can test worker manager features independent of if dragon is installed. Consider a test that only needs to verify a message is mapped from `capnproto.Request` to the `InferenceRequest` - it cannot be run without having dragon installed.

Review `DragonCommChannel` for an example of wrapping and injecting a dragon implementation of a feature. Notice the separate module containing that.

**WARNING**: the import `from smartsim._core.mli.comm.channel.dragonchannel import DragonCommChannel` is also bad form (my fault). It should be removed also, though this is not part of your issue.

AlyssaCote: Do you think in this case it might be better to do `FLInterface.attach` in the `EnvironmentConfigLoader` then? So that `FLInterface | None` is return with `get_queue()`? Or would that not make a difference?

smartsim/_core/mli/infrastructure/environmentloader.py
```diff
@@ -0,0 +1,60 @@
...
+import typing as t
+
+import dragon.utils as du
+
+from smartsim._core.mli.infrastructure.storage.featurestore import FeatureStore
+
+
+class EnvironmentConfigLoader:
+    """
+    Class to help facilitate the loading of a FeatureStore and Queue
```
ankona, line 39: Let's shorten to "Facilitates the loading..." and skip telling ourselves that a class is a class!

ankona: I think this description is going to be annoying and easy to become outdated. Consider to a more general description that won't require a change each time new config is loaded, e.g. `Facilitate the loading of environment variables holding configuration values and instantiating high level objects that use them`

smartsim/_core/mli/infrastructure/environmentloader.py
```diff
@@ -0,0 +1,60 @@
...
+    into the WorkerManager.
+    """
+
+    def __init__(self) -> None:
+        self._feature_store_bytes = os.getenv("SSFeatureStore", None)
```
ankona, line 44: Can you confirm we want `SSXxx` here? Is this a new convention to add to the `SMARTSIM_XXX` and `SS_XXX`? Do we want these to be camel- or snake-case?

smartsim/_core/mli/infrastructure/environmentloader.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_feature_store(self) -> t.Optional[FeatureStore]:
+        """Loads the Feature Store previously set in SSFeatureStore"""
+        if self._feature_store_bytes is not None:
+            self.feature_store = pickle.loads(
+                base64.b64decode(self._feature_store_bytes)
+            )
+        return self.feature_store
+
+    def get_queue(self) -> t.Optional[t.ByteString]:
```
ankona, line 54: The name is inconsistent w/behavior. `get_feature_store` returns a feature store. `get_queue` returns a descriptor. 

- Consider renaming to `get_queue_descriptor`

tests/dragon/test_environment_loader.py
```diff
@@ -0,0 +1,110 @@
...
+    "content",
+    [
+        pytest.param(b"a"),
+        pytest.param(b"new byte string"),
+    ],
+)
+def test_environment_loader_attach_FLI(content):
+    chan = Channel.make_process_local()
+    queue = FLInterface(main_ch=chan)
+    os.environ["SSQueue"] = du.B64.bytes_to_str(queue.serialize())
```
ankona, line 54: consider replacing direct environment manipulation with monkeypatch to avoid bleeding into other tests:

`monkeypatch.setenv("SSQueue", value)`

tests/dragon/test_environment_loader.py
```diff
@@ -0,0 +1,110 @@
...
+    new_sender = new_queue.sendh(use_main_as_stream_channel=True)
+    new_sender.send_bytes(content)
+
+    old_recv = queue.recvh(use_main_as_stream_channel=True)
+    result, _ = old_recv.recv_bytes()
+    assert result == content
+
+
+def test_environment_loader_serialize_FLI():
+    chan = Channel.make_process_local()
```
ankona, line 69: Consider adding docstrings to describe the purpose of each test in more detail than the method name.

tests/mli/test_worker_manager.py
```diff
@@ -184,7 +185,12 @@ def test_worker_manager(prepare_environment: pathlib.Path) -> None:
     # create a mock client application to populate the request queue
     msg_pump = mp.Process(
         target=mock_messages,
-        args=(work_queue, file_system_store, fs_path, comm_path),
+        args=(
+            config_loader.get_queue(),
```
ankona, line 189: Does this work? It seems like we would need to put a valid value in env vars for the queue and feature store, but never do so.

This test isn't a test of config loader, so consider monkeypatching:

```py
with monkeypach.context() as ctx:
  ctx.setattr(config_loader, "get_queue", work_queue)
  ctx.setattr(config_loader, "get_feature_store", file_system_store)
```

AlyssaCote: I think it likely works because if the environment variable aren't set, they just return `None`. But I'll fix the test!

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: I `FLInterface.attach` in `EnvironmentConfigLoader ` now and `get_queue()` returns `FLInterface | None`, which should solve this problem.

smartsim/_core/mli/infrastructure/environmentloader.py
AlyssaCote: I'm returning the FLInterface now, so I think this should stay `get_queue `.

smartsim/_core/mli/infrastructure/environmentloader.py
AlyssaCote: We decided these were fine in an offline discussion

ankona: I am incorrect and SS is existing standard.

pyproject.toml
```diff
@@ -62,6 +62,7 @@ markers = [
   "group_a: fast test subset a",
   "group_b: fast test subset b",
   "slow_tests: tests that take a long duration to complete",
+  "dragon: dragon tests",
```
ankona, line 65: pedantic/optional request - could we replace defining something with itself by saying "tests that must be executed in a dragon runtime"

smartsim/_core/mli/infrastructure/environmentloader.py
```diff
@@ -0,0 +1,61 @@
...
+        if self._feature_store_descriptor is not None:
+            self.feature_store = pickle.loads(
+                base64.b64decode(self._feature_store_descriptor)
+            )
+        return self.feature_store
+
+    def get_queue(self) -> t.Optional["FLInterface"]:
+        """Returns the Queue  previously set in SSQueue"""
+        if self._queue_descriptor is not None:
+            return FLInterface.attach(base64.b64decode(self._queue_descriptor))
```
ankona, line 60: it appears that we're making caching of some value into an external clients problem.

we have:

```py
if A:
  return Thing
return B
```

shouldn't we be setting the `A` to ensure cache behavior? for example:

```py
def get_queue(self) -> t.Optional["FLInterface"]:
    """Returns the Queue  previously set in SSQueue"""
    if self.queue is not None:
        # load it one time
        self.queue = FLInterface.attach(base64.b64decode(self._queue_descriptor))
    # return the cached item 
    return self.queue
```

AlyssaCote: Yes you're right! I had it this way before and must have changed it when I reworked it. I'll fix asap.

ankona (approved):
LGTM!

PR #623, opened 2024-06-25
Restrict to numpy 1.x

ashao:
The new major version release of Numpy is incompatible with modules compiled against Numpy 1.x. For both SmartSim and SmartRedis we request a 1.x version of numpy. This is needed in SmartSim because some of the downstream dependencies request NumPy.


al-rigazzi (approved):
LGTM, thanks for fixing this bug so quickly!

PR #624, opened 2024-06-27
Add model metadata to request schema

AlyssaCote:
This PR adds a `Model` schema in the `Request` schema with metadata about the model included so that it's not just a data blob. I also tweaked some of the attribute names. For example, what was `request.input.inputKeys` is now `request.input.keys`.


smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -65,12 +66,12 @@ def deserialize_message(
...
     # return request
     model_key: t.Optional[str] = None
-    model_bytes: t.Optional[bytes] = None
+    model_bytes: t.Optional[Model] = None
 
-    if request.model.which() == "modelKey":
-        model_key = request.model.modelKey.key
-    elif request.model.which() == "modelData":
-        model_bytes = request.model.modelData
+    if request.model.which() == "key":
```
ankona, line 71: ❤️ loving the simplified naming

smartsim/_core/mli/mli_schemas/model/model.capnp
```diff
@@ -0,0 +1,33 @@
...
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+@0xaefb9301e14ba4bd;
+
+struct Model {
+  data @0 :Data;
+  name @1 :Text;
+  version @2 :Text;
+}
```
ankona, line 33: trailing newline to make linters happy?

AlyssaCote: Added a newline to all `capnp` schema files!

smartsim/_core/mli/mli_schemas/model/model_capnp.pyi
```diff
@@ -0,0 +1,46 @@
+"""This is an automatically generated stub for `model.capnp`."""
```
ankona, line 1: BSD license

smartsim/_core/mli/mli_schemas/model/model_capnp.py
```diff
@@ -0,0 +1,12 @@
+"""This is an automatically generated stub for `model.capnp`."""
```
ankona, line 1: license

ankona (changes requested):
I see a couple minor license issues and a potential breaking bug to fix. Otherwise, looks good.

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -91,19 +92,19 @@ def deserialize_message(
...
+        input_meta = [data.tensorDescriptor for data in request.input.data]
 
     inference_request = InferenceRequest(
         model_key=model_key,
         callback=comm_channel,
         raw_inputs=input_bytes,
         input_meta=input_meta,
         input_keys=input_keys,
-        raw_model=model_bytes,
+        raw_model=model_bytes.data if model_bytes is not None else None,
```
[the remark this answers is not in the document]
AlyssaCote: The new `Model` schema has a `data` attribute that holds the bytes. `request.model.data` accesses the `Model` object, and the one more `.data` accesses the bytes within the model object. If we just want to pass around the raw bytes, then I think this is correct, but if we want to pass the `Model` schema object around (which I think we do, now that I'm seeing it), then I think some renaming may need to happen?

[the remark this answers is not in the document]
AlyssaCote: What I've done is pass in `model_bytes` which is now a `Model` schema object, and then later on when bytes need to be passed into `FetchModelResults` I pass in the `data` attribute, which is bytes. Let me know what you think!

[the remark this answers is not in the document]
ankona: that makes sense, now. it looked like repeating the same method call but was not. resolving!

ankona (approved):
LGTM

PR #626, opened 2024-07-01
[nomerge] ENV loader

al-rigazzi:


al-rigazzi:
The env loader was implemented as part of #621

PR #629, opened 2024-07-02
Improve error handling in worker manager

AlyssaCote:
This PR aims to allow the `WorkerManager` to continue if a `worker` throws an error. The `WorkerManager` needs  to return a `Response` without blowing up in the process.


smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -71,11 +74,15 @@ def __init__(
         outputs: t.Optional[t.Collection[t.Any]] = None,
         output_keys: t.Optional[t.Collection[str]] = None,
         failed: bool = False,
+        status_enum: "StatusEnum" = "complete",
```
ankona, line 77: Should probably _use_ the enum if we declare the type as such? e.g.

`status: StatusEnum = StatusEnum.complete`

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -71,11 +74,15 @@ def __init__(
         outputs: t.Optional[t.Collection[t.Any]] = None,
         output_keys: t.Optional[t.Collection[str]] = None,
         failed: bool = False,
```
ankona, line 76: Should we dump `failed` and just use the value of `status` instead?

tests/mli/test_error_handling.py
```diff
@@ -0,0 +1,164 @@
...
+    work_queue: "mp.Queue[bytes]" = mp.Queue()
+    integrated_worker = IntegratedTorchWorker()
+    file_system_store = FileSystemFeatureStore(test_dir)
+
+    worker_manager = WorkerManager(
+        work_queue,
+        integrated_worker,
+        file_system_store,
+        as_service=False,
+        cooldown=10,
```
ankona, line 57: - `cooldown=10` tells the service to run 10s after it satisfies shutdown conditions

this _might_ cause tests to run for an extra 10s, even if it's not necessary. If these tests are taking a long time, consider adjusting cooldown

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: It's a `StatusEnum` on the `capnproto` schema side of things, so we need to set it to be one of its accepted strings. `StatusEnum.complete` doesn't do the trick but I really wish it did.

from `pycapnp` quickstart guide:
```
struct PhoneNumber {
    number @0 :Text;
    type @1 :Type;

    enum Type {
      mobile @0;
      home @1;
      work @2;
    }
  }


Enums are treated like strings, and you assign to them like they were a Text field:

alicePhone.type = 'mobile'
```

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: I won't lie, keeping  a `failed` parameter makes the if checks I make throughout a lot easier.

AlyssaCote: This makes me want to add a n `in progress` enum so I don't immediately set it to `complete`. I think I'll do that.

tests/mli/test_error_handling.py
```diff
@@ -0,0 +1,163 @@
...
+    request = MessageHandler.build_request(
+        b"channel", model, [tensor_key], [tensor_key], [], None
+    )
+    ser_request = MessageHandler.serialize_request(request)
+
+    return worker_manager, work_queue, integrated_worker, ser_request
+
+
+def test_execute_errors_handled(setup_worker_manager, monkeypatch):
+    worker_manager, work_queue, integrated_worker, ser_request = setup_worker_manager
```
ankona, line 71: Let's add some comments to tests that describe the test in a little more verbosity than the title alone, e.g. `"""Ensures that the worker manager continues to handling incoming requests after a failure in the `load_model` pipeline stage"""`

tests/mli/test_error_handling.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def mock_execute():
+        raise ValueError("Simulated error in execute")
+
+    work_queue.put(ser_request)
+
+    monkeypatch.setattr(integrated_worker, "execute", mock_execute)
+
+    worker_manager._on_iteration()
```
ankona, line 80: we don't have any verification of the results here. we probably could benefit by making sure that `worker_manager` really does respond.

you could make use of the knowledge that you're using a file system feature store and verify the channel output was written:

```py
assert test_path = pathlib.Path(test_dir)
assert list(test_path.glob("*"))
```

- or - to avoid file system (probably preferable)

````py
mock_reply_fn = MagicMock()
monkeypatch.setattr("smartsim._core.mli.infrastructure.workermanager", "build_failure_reply, mock_reply_fn)

...

assert mock_reply_fn.called_once()
  
...

# now, in the test you can use magic mock features to verify call count

assert ser_request.reply_channel

```

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -240,41 +260,85 @@ def _on_iteration(self) -> None:
...
+        transformed_input = None
+        execute_result = None
+        transformed_output = None
+
         try:
-            execute_result = self._worker.execute(
-                request, model_result, transformed_input
-            )
+            fetch_model_result = self._worker.fetch_model(request, self._feature_store)
+        except Exception as e:
```
ankona, line 274: can we reduce noise and repetition here by passing the action to the exception handler?

```py

def fail_handler(
    func: t.Callable[..., t.Any],  # <-----
    func_args: t.Tuple[t.Any, ...],  # <-----
    exc: Exception,
    func_descriptor: str,
    reply: InferenceReply,
) -> None:
    ...
    try:
        return func(*func_args)  # <-----
    except:
        logger.exception(
            f"An error occurred while {func_descriptor}.\n"
            f"Exception type: {type(exc).__name__}.\n"
            f"Exception message: {str(exc)}"
        )
        reply.status_enum = "fail"
        reply.message = f"Failed while {func_descriptor}."
```

and modifying the calls like:

```py
# fetch_model_result = self._worker.fetch_model(request, self._feature_store)
fetch_model_result = fail_handler(
  self._worker.fetch_model,
  (self._feature_store, )
  "fetching the model",
  reply
)
```

ankona: another idea here for reducing verbosity is to avoid all the extra checks....

```py
# send reply on fail
def exception_handler(
    exc: Exception, func_descriptor: str, reply: InferenceReply,
    reply_channel: CommChannelBase #<------
) -> None:
  ...
  reply.statu_enum = "failed"
  reply.message = f"failed while..."
  reply_channel.send(reply)
```

and return on exceptions

```py
try:
            fetch_model_result = self._worker.fetch_model(request, self._feature_store)
        except Exception as e:
            exception_handler(e, "fetching the model", reply)
            return. # <-----
```


```

Comment (AlyssaCote) [reply smartsim/_core/mli/infrastructure/control/workermanager.py:279]: Oh I like the idea of sending the failed response asap!

File: tests/mli/test_error_handling.py (hunk 5 above)
Comment (AlyssaCote) [reply tests/mli/test_error_handling.py]: Currently working on this! Now that I return early if there's an exception I need to make sure that I can properly get all the way down to each stage.

File: smartsim/_core/mli/infrastructure/worker/worker.py (hunk 1 above)
Comment (ankona) [reply smartsim/_core/mli/infrastructure/worker/worker.py]: Understood. Any issues with avoiding `XxxEnum` naming and just do `Status` as the field name here?

Comment (AlyssaCote) [reply smartsim/_core/mli/infrastructure/worker/worker.py]: `StatusEnum` -> `Status` is complete!

File: smartsim/_core/mli/infrastructure/control/workermanager.py [hunk 7]
```diff
@@ -152,13 +157,41 @@ def build_reply(reply: InferenceReply) -> Response:
...
         result=results,
         custom_attributes=None,
     )
 
 
+def exception_handler(
+    exc: Exception,
+    reply_channel: t.Optional[CommChannelBase],
+    func_descriptor: str,
+    reply: InferenceReply,
```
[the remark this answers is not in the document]
AlyssaCote: I think you're right! I can totally get rid of the `reply` being passed into the `exception_handler`.

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: I went with the last option!

tests/mli/test_error_handling.py
AlyssaCote: Tests are fixed.

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -107,19 +108,23 @@ def deserialize_message(
...
         raw_model=model_bytes,
         batch_size=0,
     )
     return inference_request
 
 
-def build_failure_reply(status: "StatusEnum", message: str) -> Response:
+def build_failure_reply(status: "Status", message: str) -> Response:
     return MessageHandler.build_response(
         status=status,  # todo: need to indicate correct status
```
mellis13, line 129: Are these TODO comments still applicable?  Can they be addressed or are they follow-up tickets?

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -258,95 +295,137 @@ def _on_iteration(self) -> None:
         timings.append(time.perf_counter() - interm)  # timing
         interm = time.perf_counter()  # timing
 
+        reply = InferenceReply()
+
         if not request.raw_model:
             if request.model_key is None:
-                # A valid request should never get here.
-                raise ValueError("Could not read model key")
+                send_failure(request.callback, "Could not find model key or model.")
```
mellis13, line 302: Comment for discussion in MLI sync: I just noticed this, but there seems to be two paths for handling errors, one is a call to ``send_failure`` and the other is to call ``exception_handler`` which then calls ``send_failure``.  Is there a way we can align with one method for all cases so that it is consistent what should be done for errors and exceptions?

smartsim/_core/mli/infrastructure/control/workermanager.py
```diff
@@ -258,95 +295,137 @@ def _on_iteration(self) -> None:
...
-                        )
-                    except KeyError:
-                        time.sleep(0.1)
-                    else:
-                        break
-
-                if fetch_model_result is None:
-                    raise SmartSimError("Could not retrieve model from feature store")
+                try:
+                    interm = time.perf_counter()  # timing
```
mellis13: it seems funny that this time.perf_counter is inside the try block but the next one isn't. Should it be?

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: I just added this yesterday. Basically I split it out because with Al's new code there are situations where there is no exception, but we need to return because it's at a state where we can no longer continue. Rather than have the Worker Manager raise an error, we want to just send a response back that there was a failure. Happy to discuss it more in the meeting because I was also unsure what to do in these situations!

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: These are old comments that I just didn't delete. I think these particular ones can be deleted but I wasn't sure if I should be the one to delete them or if others wanted them for reference.

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: Actually, I thought I needed to use `send_failure` in more than one spot, but since it's only here I can use `exception_handler` and pass in the `ValueError` so there is only one path for handling error. Fixing this now!

AlyssaCote: `send_failure` has been removed

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: I'll ask Al if these should all be inside the try blocks or outside of them so there's consistency.

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: Comments removed.

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: Marking as resolved per our group discussion.

smartsim/_core/mli/infrastructure/control/workermanager.py
AlyssaCote: Asked Al and I made sure they're all outside of the try blocks.

mellis13 (approved):

ankona (approved):
LGTM!

PR #630, opened 2024-07-03
Add Type Check Back to Refactor CI

MattToast:
Addresses type errors found in the SmartSim core refactor branch. Re-adds a static type check back into the CI to allow for gradual typing as more parts of the API become more concrete


smartsim/experiment.py
```diff
@@ -477,7 +477,7 @@ def preview(
         """
 
         # Retrieve any active feature store jobs
-        active_fsjobs = self._control.active_active_feature_store_jobs
```
amandarichardsonn, line 480: nice catch!

smartsim/launchable/baseJobGroup.py
```diff
@@ -8,14 +10,11 @@
 from .basejob import BaseJob
 
 
-class BaseJobGroup(Launchable, MutableSequence):
```
amandarichardsonn, line 11: That was my bad, nice catch!

amandarichardsonn: How did the tests still pass even though I did not inherit the ABC class?

smartsim/launchable/baseJobGroup.py
```diff
@@ -8,14 +10,11 @@
...
+class BaseJobGroup(Launchable, MutableSequence[BaseJob], ABC):
     """Highest level ABC of a group of jobs that can be
     launched
     """
 
-    def __init__(self) -> None:
```
amandarichardsonn, line 16: noted!

smartsim/launchable/baseJobGroup.py
```diff
@@ -36,23 +35,31 @@ def __iter__(self) -> t.Iterator[BaseJob]:
...
+        self, idx: int | slice, value: BaseJob | t.Iterable[BaseJob]
+    ) -> None:
         """Sets the job at the specified index (idx) to the given value."""
-        self.jobs[idx] = deepcopy(value)
+        if isinstance(idx, int):
+            if not isinstance(value, BaseJob):
+                raise TypeError("Can only assign a `BaseJob`")
+            self.jobs[idx] = deepcopy(value)
+        else:
+            if not isinstance(value, t.Iterable):
```
amandarichardsonn, line 51: Do you need to check here that its an iterable of BaseJobs?

smartsim/launchable/baseJobGroup.py
```diff
@@ -36,23 +35,31 @@ def __iter__(self) -> t.Iterator[BaseJob]:
         """Allows iteration over the jobs in the collection."""
         return iter(self.jobs)
 
-    def __getitem__(self, idx: int) -> BaseJob:
-        """Retrieves the job at the specified index (idx)."""
-        return self.jobs[idx]
-
-    def __setitem__(self, idx: int, value: BaseJob) -> None:
+    @t.overload
+    def __setitem__(self, idx: int, value: BaseJob) -> None: ...
```
amandarichardsonn, line 39: why is setitem repeated twice and what is t.overload?

smartsim/launchable/jobGroup.py
```diff
@@ -26,9 +31,19 @@ def jobs(self) -> t.List[BaseJob]:
         """
         return self._jobs
 
-    def __str__(self):  # pragma: no-cover
+    @t.overload
+    def __getitem__(self, idx: int) -> BaseJob: ...
```
amandarichardsonn, line 35: repeated dunder methods again, hmmmm

smartsim/settings/__init__.py
```diff
@@ -32,69 +34,55 @@
 
 
 # TODO Mock imports for compiling tests
-class DragonRunSettings:
-    pass
+class SettingsBase:
+    def __init__(self, *_: t.Any, **__: t.Any) -> None: ...
```
amandarichardsonn, line 38: why did u define dunders for base?

tests/temp_tests/test_colocatedJobGroup.py
```diff
@@ -10,8 +10,13 @@
 app_3 = Application("app_3", "python", run_settings=LaunchSettings("slurm"))
 
 
+class MockJob(BaseJob):
```
amandarichardsonn, line 13: why did you switch out BaseJob for MockJob?

amandarichardsonn (approved):

smartsim/launchable/baseJobGroup.py
MattToast: Without going _too_ far into the weeds of how `ABC` and `abstractmethod` work, the `@abstractmethod` decorator just sets a `__isabstractmethod__ = True` attr to the method.

```py
class C:
    @abstractmethod
    def some_method(self):  print("Totally works!")

C().some_method()  # prints: "Totally works!"
C().some_method.__isabstractmethod__  # returns: `True`
```

The `ABC` class is what actually enforces that no methods have a `__isabstractmethod__` set to true when the class is created:
```py
class C(ABC):
    @abstractmethod
    def some_method(self): ...

C()  # Now cannot be instanced, raises a `TypeError`
```
So basically, it got past the tests precisely _because_ it did not inherit from `ABC`!

smartsim/launchable/baseJobGroup.py
MattToast: `__setitem__` is actually defined here 3 times! The first two definitions give the different signatures to the type checker to provide more information about how a method should be used. Only the last one is the one that matters for runtime though.

Python does not have proper method overloading (each method has one, and exactly one, definition), but the [`@t.overload`](https://docs.python.org/3/library/typing.html#typing.overload) decorator informs the type checker that depending on how the method is called there are really many (in this case two) discrete signatures. This allows for python to "spoof" true method overloading at type check time.

Now if you index into the job group with an `int` mypy knows that you should be trying to assign a `BaseJob`, but if you index into the job group with a `slice`, mypy knows that it should expect an `Iterable[BaseJob]`

smartsim/launchable/baseJobGroup.py
MattToast: If we want to keep it type safe at runtime, we unfortunately do :(.

From the overloads, we can tell that when `idx` is a `slice` that `value` should be a `Iterable`, but unfortunately that assumed constraint is not actually guaranteed at runtime, and thus the type checker will complain for the definition of this method if we do not explicitly handle it.

smartsim/launchable/jobGroup.py
MattToast: Similar to the [`__setitem__`](https://github.com/CrayLabs/SmartSim/pull/630/files#r1669426045) conversation, only the last definition matters at runtime, but these other methods provide more information to the type checker.

Now if a user calls this method with different indexes, mypy is able to infer what type you _should_ get back
```py
reveal_type(job_group[0])    # reveals: `int`
reveal_type(job_group[1:3])  # reveals: `JobGroup`
```
but do note that it is still up to the author to make sure that the actual definition conforms to the method overloads. In this case, it's basically a fancy way to the the type checker "trust me: when you get this input you can expect this output".

smartsim/settings/__init__.py
MattToast: Just to get past the type check without ignoring all of `_core`. There are still a few places where settings are instanced using one of these stubclasses. If we do not define a constructor that takes any number of args and any number of kwargs, the type checker will complain anytime a settings is instanced that does not conform to the signature

MattToast: Similarly, the `__getattr__` is there so that the type checker does not complain when try to dot-access an attr that does not directly exist on these stub classes.

tests/temp_tests/test_colocatedJobGroup.py
MattToast: We need to provide a concrete implementation of the `BaseJob` abstract class. Otherwise we cannot instance the class we are testing!

smartsim/launchable/baseJobGroup.py
MattToast: This is 100% me forcing my style on the project, btw! These lines are perfectly valid, I just prefer to not redefine methods if all they are going to do is defer to the super implementation. I find it helps keep down the amount of clutter in a file.

If we decide that we actually prefer this style (I can't think of a good reason to prefer it off the top of my head, but that doesn't necessarily mean it doesn't exist) I could be swayed to revert this change!!

PR #642, opened 2024-07-18
Fix-hostname

al-rigazzi:
This PR addresses an inconstent internal host name representation in the Dragon backend.

### DIFF
--- a/smartsim/_core/launcher/dragon/dragonBackend.py
+++ b/smartsim/_core/launcher/dragon/dragonBackend.py
@@ -210,10 +210,8 @@ class DragonBackend:
 
     def _initialize_hosts(self) -> None:
         with self._queue_lock:
-            self._hosts: t.List[str] = sorted(
-                node for node in dragon_machine.System().nodes
-            )
-            self._nodes = [dragon_machine.Node(node) for node in self._hosts]
+            self._nodes = [dragon_machine.Node(node) for node in dragon_machine.System().nodes]
+            self._hosts: t.List[str]  = sorted(node.hostname for node in self._nodes)
             self._cpus = [node.num_cpus for node in self._nodes]
             self._gpus = [node.num_gpus for node in self._nodes]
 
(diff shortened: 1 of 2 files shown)


smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -452,7 +452,7 @@ def create_run_policy(
                 if run_request.policy.gpu_affinity:
                     affinity = dragon_policy.Policy.Affinity.SPECIFIC
                     gpu_affinity = run_request.policy.gpu_affinity
-
+            logger.debug(f"Affinity: {affinity}, {cpu_affinity}, {gpu_affinity}")
```
mellis13, line 455: If we are going to leave this in, can we clarify the message so we don't rely on remembering the order of the debug output:

```python
logger.debug(f"Affinity: {affinity}")
logger.debug(f"CPU affinity: {cpu_affinity}")
logger.debug(f"GPU affinity: {gpu_affinity}")
```

or do the equivalent in a single line

al-rigazzi: Sure, let me do this!

ankona: Per meeting w/dragon team, the affinity parameter is also being dropped and is not used. I should removed affinity and leave only cpu/cpu affinity lists behind. I think that will clear up your ordering comment?

mellis13: Agreed they are dropping the ``affinity`` parameter.  We can still specify CPU and GPU affinity at the same time (and probably want to know which is which) right?

al-rigazzi: Yeah, we'll update it as soon as the new version is out.

ankona (approved):
LGTM! Thanks for fixing my bug

PR #646, opened 2024-07-23
Link up SmartSim Refactor CI

MattToast:
Add headers/marks/boilerplate to test files for `smartsim-refactor`

### DIFF
--- a/tests/temp_tests/test_colocatedJobGroup.py
+++ b/tests/temp_tests/test_colocatedJobGroup.py
@@ -1,8 +1,38 @@
+# BSD 2-Clause License
+#
+# Copyright (c) 2021-2024, Hewlett Packard Enterprise
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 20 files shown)


tests/temp_tests/test_colocatedJobGroup.py
```diff
@@ -1,9 +1,39 @@
...
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import pytest
+
 from smartsim.entity.model import Application
 from smartsim.launchable.basejob import BaseJob
 from smartsim.launchable.colocatedJobGroup import ColocatedJobGroup
 from smartsim.launchable.job import Job
 from smartsim.settings import LaunchSettings
```
amandarichardsonn, line 34: I notice in the `tests/test_init.py` file, we have `[pytest.mark.group_a, pytest.mark.group_b, pytest.mark.slow_tests]` - what are each for?

amandarichardsonn (approved):
Nice catch! Really support your idea of a check, mayhaps could add it in this PR

tests/temp_tests/test_colocatedJobGroup.py
MattToast: The old CI was arbitrarily split up into three groups (`group_a` (no real meaning), `group_b` (no real meaning), and `slow_tests` (basically our integration tests)) to try and reduce the amount of time it took to expose an error. Otherwise the old CI would run for a _really_ long time (like 30-45 min)

I'm hoping now that we are making the API a bit more modular and we are writing more "true" unit tests, we can remove the need for this arbitrary split all together and keep our test suite running fast because it leads to weird inconsistencies like this where developers write tests but then the CI doesn't run them. But that's more work on the CI than I am willing to address in the `smartsim-refactor` branch. So for now we'll just add a check that looks over the test files and makes sure there is  a line that looks like `pytestmark = .*[group_a,group_b,slow_tests].*`

As for what
```py
pytestmark = [pytest.mark.group_a, pytest.mark.group_b, pytest.mark.slow_tests]
```
in `tests/test_init.py` means specifically; it just means that test file should be run in all of the test groups!

PR #653, opened 2024-07-31
Mitigate dragon/numpy, mypy/typing_extension dependency issues

ankona:
This PR mitigates two issues encountered during installation on build agents

## mypy/typing_extensions

Installation of mypy or dragon in separate build actions caused some dependencies (typing_extensions, numpy) to be upgraded. Those upgrades result in runtime failures. The build actions were tweaked to allow pip to consider all optional dependencies during resolution. 

## dragon/numpy

Additionally, the numpy version was capped on dragon installations.

### DIFF
--- a/smartsim/_core/_cli/scripts/dragon_install.py
+++ b/smartsim/_core/_cli/scripts/dragon_install.py
@@ -182,7 +182,7 @@ def install_package(asset_dir: pathlib.Path) -> int:
         logger.info(f"Installing package: {wheel_path.absolute()}")
 
         try:
-            pip("install", "--force-reinstall", str(wheel_path))
+            pip("install", "--force-reinstall", str(wheel_path), "numpy<2")
             wheel_path = next(wheels, None)
         except Exception:
             logger.error(f"Unable to install from {asset_dir}")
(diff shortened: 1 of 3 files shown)


ashao (approved):
Thanks for tracking this one down so quickly

MattToast (approved):
LGTM!!

smartsim/_core/_cli/scripts/dragon_install.py
```diff
@@ -182,7 +182,7 @@ def install_package(asset_dir: pathlib.Path) -> int:
         logger.info(f"Installing package: {wheel_path.absolute()}")
 
         try:
-            pip("install", "--force-reinstall", str(wheel_path))
+            pip("install", "--force-reinstall", str(wheel_path), "numpy<2")
```
MattToast, line 185: Actually one small note: It looks like the version of numpy that comes in when doing a "normal"
```
pip install --no-cache-dir smartsim[dev,mypy,ml] && smart build --onnx
```
is version 1.24.3 (at least for me, from pypi, on osx).

This change will presumably pin to [version 1.26.4](https://pypi.org/project/numpy/#history). I'm not sure if that matters, but if you want to be _SUPER_ safe, you may want to pin to that number.

MattToast (commented):
One last thing actually! Not worth holding up approval over, though!!

MattToast:
Whoops, just missed the merge, hahaha

PR #654, opened 2024-08-01
Remove builder from setup.py

ashao:
The builder module was included in `setup.py` to allow us to ship the main Redis binaries (not RedisAI) with installs from PyPI. The changes in this PR remove our ability to do this and requires users to build Redis as part of the `smart build`. This change in behaviour is justified by the following:
- The additional cost/complexity to the user for building Redis is minimal due to the pre-existing need to compile RedisAI.
- Removes the need for different platform-dependent wheels
- Frees the builder module from being a single file to allow convenient import into `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,12 +95,6 @@ buildenv_spec = importlib.util.spec_from_file_location("buildenv", str(buildenv_
 buildenv = importlib.util.module_from_spec(buildenv_spec)
 buildenv_spec.loader.exec_module(buildenv)
 
-# import builder module
-builder_path = _install_dir.joinpath("builder.py")
-builder_spec = importlib.util.spec_from_file_location("builder", str(builder_path))
-builder = importlib.util.module_from_spec(builder_spec)
-builder_spec.loader.exec_module(builder)
-
 # helper classes for building dependencies that are
 # also utilized by the Smart CLI
 build_env = buildenv.BuildEnv(checks=False)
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -128,47 +122,8 @@
...
-    """
-
-    def has_ext_modules(_placeholder):
-        return True
-
-
 # Define needed dependencies for the installation
 
-# Add SmartRedis at specific version
-# install_requires.append("smartredis>={}".format(versions.SMARTREDIS))
```
ashao, line 170: Note that this was a holdover from a previous PR

setup.py
```diff
@@ -128,47 +122,8 @@
 class BuildError(Exception):
     pass
 
-
-# Hacky workaround for solving CI build "purelib" issue
-# see https://github.com/google/or-tools/issues/616
-class InstallPlatlib(install):
```
MattToast, line 134: Minor nit: can you remove the `install` import from the top of the module?

setup.py
```diff
@@ -128,47 +122,8 @@
...
-    def finalize_options(self):
-        super().finalize_options()
-        if self.distribution.has_ext_modules():
-            self.install_lib = self.install_platlib
-
-
-class SmartSimBuild(build_py):
```
MattToast, line 141: ditto for `build_py`

MattToast (approved):
A two of small requests, but nothing worth holding approval up over.

LGTM, pending CI!!

PR #655, opened 2024-08-02
Simple Status Checking

MattToast:
Add ability for the `Experiment` to fetch the status a launched job that it started given a `LaunchedJobID`. Teach the `ShellLauncher` and `DragonLauncher` to get statuses of jobs they have launched.

### DIFF
--- a/smartsim/_core/control/jobmanager.py
+++ b/smartsim/_core/control/jobmanager.py
@@ -228,7 +228,7 @@ class JobManager:
     def get_status(
         self,
         entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]],
-    ) -> SmartSimStatus | FailedToFetchStatus:
+    ) -> t.Union[SmartSimStatus, FailedToFetchStatus]:
         """Return the status of a job.
 
         :param entity: SmartSimEntity or EntitySequence instance
--- a/smartsim/_core/control/jobmanager.py
+++ b/smartsim/_core/control/jobmanager.py
@@ -36,7 +36,7 @@ from ..._core.launcher.step import Step
 from ...database import FeatureStore
 from ...entity import EntitySequence, FSNode, SmartSimEntity
 from ...log import ContextThread, get_logger
-from ...status import TERMINAL_STATUSES, SmartSimStatus
+from ...status import TERMINAL_STATUSES, FailedToFetchStatus, SmartSimStatus
 from ..config import CONFIG
 from ..launcher import Launcher, LocalLauncher
 from ..utils.network import get_ip_from_host
@@ -228,7 +228,7 @@ class JobManager:
     def get_status(
         self,
         entity: t.Union[SmartSimEntity, EntitySequence[SmartSimEntity]],
-    ) -> SmartSimStatus:
+    ) -> SmartSimStatus | FailedToFetchStatus:
         """Return the status of a job.
 
         :param entity: SmartSimEntity or EntitySequence instance
@@ -242,7 +242,7 @@ class JobManager:
                 job: Job = self[entity.name]  # locked
                 return job.status
 
-            return SmartSimStatus.STATUS_NEVER_STARTED
+            return FailedToFetchStatus.STATUS_NEVER_STARTED
 
     def set_launcher(self, launcher: Launcher) -> None:
         """Set the launcher of the job manager to a specific launcher instance
--- a/smartsim/_core/launcher/dragon/dragonLauncher.py
+++ b/smartsim/_core/launcher/dragon/dragonLauncher.py
@@ -145,9 +145,11 @@ class DragonLauncher(WLMLauncher):
         res = _assert_schema_type(self._connector.send_request(req), DragonRunResponse)
         return LaunchedJobID(res.step_id)
 
-    def get_status(self, *launched_ids: LaunchedJobID) -> tuple[SmartSimStatus, ...]:
+    def get_status(
+        self, *launched_ids: LaunchedJobID
+    ) -> t.Mapping[LaunchedJobID, SmartSimStatus]:
         infos = self._get_managed_step_update(list(launched_ids))
-        return tuple(info.status for info in infos)
+        return {id_: info.status for id_, info in zip(launched_ids, infos)}
 
     def run(self, step: Step) -> t.Optional[str]:
         """Run a job step through Slurm
@@ -341,7 +343,7 @@ class DragonLauncher(WLMLauncher):
             return [step_id_updates[step_id] for step_id in step_ids]
         except KeyError:
             msg = "A step info could not be found for one or more of the requested ids"
-            raise errors.UnrecognizedLaunchedJobError(msg) from None
+            raise errors.LauncherJobNotFound(msg) from None
 
     def __str__(self) -> str:
         return "Dragon"
--- a/smartsim/_core/launcher/dragon/dragonLauncher.py
+++ b/smartsim/_core/launcher/dragon/dragonLauncher.py
@@ -30,6 +30,7 @@ import os
 import typing as t
 
 from smartsim._core.schemas.dragonRequests import DragonRunPolicy
+from smartsim.error import errors
 from smartsim.types import LaunchedJobID
 
 from ...._core.launcher.stepMapping import StepMap
@@ -143,6 +144,10 @@ class DragonLauncher(WLMLauncher):
         req = DragonRunRequest(**dict(req_args), current_env=merged_env, policy=policy)
         res = _assert_schema_type(self._connector.send_request(req), DragonRunResponse)
         return LaunchedJobID(res.step_id)
+
+    def get_status(self, *launched_ids: LaunchedJobID) -> tuple[SmartSimStatus, ...]:
+        infos = self._get_managed_step_update(list(launched_ids))
+        return tuple(info.status for info in infos)
 
     def run(self, step: Step) -> t.Optional[str]:
         """Run a job step through Slurm
@@ -331,8 +336,12 @@ class DragonLauncher(WLMLauncher):
 
                 step_id_updates[step_id] = info
 
-        # Order matters as we return an ordered list of StepInfo objects
-        return [step_id_updates[step_id] for step_id in step_ids]
+        try:
+            # Order matters as we return an ordered list of StepInfo objects
+            return [step_id_updates[step_id] for step_id in step_ids]
+        except KeyError:
+            msg = "A step info could not be found for one or more of the requested ids"
+            raise errors.UnrecognizedLaunchedJobError(msg) from None
 
     def __str__(self) -> str:
         return "Dragon"
--- a/smartsim/_core/utils/helpers.py
+++ b/smartsim/_core/utils/helpers.py
@@ -48,6 +48,7 @@ if t.TYPE_CHECKING:
 
 
 _T = t.TypeVar("_T")
+_HashableT = t.TypeVar("_HashableT", bound=t.Hashable)
 _TSignalHandlerFn = t.Callable[[int, t.Optional["FrameType"]], object]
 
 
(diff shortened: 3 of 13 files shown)


smartsim/settings/dispatch.py
```diff
@@ -378,10 +382,16 @@ class ExecutableProtocol(t.Protocol):
...
+class LauncherProtocol(collections.abc.Hashable, t.Protocol[_T_contra]):
     @classmethod
+    @abc.abstractmethod
     def create(cls, exp: Experiment, /) -> Self: ...
+    @abc.abstractmethod
+    def start(self, launchable: _T_contra, /) -> LaunchedJobID: ...
+    @abc.abstractmethod
+    def get_status(
+        self, *launched_ids: LaunchedJobID
+    ) -> tuple[SmartSimStatus, ...]: ...
```
MattToast, line 394: Not sure how I feel about `start` acting on scalars and `get_status` on collections of `LaunchedJobID`s. I think this may need to be a necessary evil though, especially if we want to enable "batch" status fetching akin to SSv0.7 `SlurmLauncher`.

Probably a good point to discuss with the group!

smartsim/_core/launcher/dragon/dragonLauncher.py
```diff
@@ -144,6 +145,10 @@ def start(
         res = _assert_schema_type(self._connector.send_request(req), DragonRunResponse)
         return LaunchedJobID(res.step_id)
 
+    def get_status(self, *launched_ids: LaunchedJobID) -> tuple[SmartSimStatus, ...]:
+        infos = self._get_managed_step_update(list(launched_ids))
```
juliaputko, line 149: infos plural intentional?

juliaputko (approved):

smartsim/settings/dispatch.py
MattToast, line 394: Currently, this assumes with this call:
```py
stat_1, stat_2, stat_3, ... = launcher.get_status(id_1, id_2, id_3, ...)
```
`stat_1` is the status of the job with id `id_1`, `stat_2` is the status of the job with id `id_2`, etc. and is behavior that is relied upon by the `Experiment`.

Do we like this or is this too much of an "implied constraint" from the protocol?

Should the return type here actually be a `Mapping[LaunchJobID, SmartSimStatus]` or maybe a `Collection[tuple[LaunchJobID, SmartSimStatus]]`? That way its more obvious to users looking to write and register their own launchers the intention of what this method should return.

smartsim/_core/launcher/dragon/dragonLauncher.py
MattToast: `infos`, plural intentional! The return type of `_get_managed_step_update` is `list[StepInfo]`, so I pluralized as a reminder that this is a "collection of many `StepInfo` instances".

If its unclear or confusing to read, I'm more than willing to rename to `info_list` or similar!

smartsim/settings/dispatch.py
amandarichardsonn: I feel like it might not be implied for all SS users that stat_1 and stat_2 will map to id_1 and id_2 but I might be wrong. I would rather get back a Mapping or Iterable -> Im wondering what will be easier for the user to inspect, maybe an iterable since you can use a for loop? But then again I can search via key in a Mapping - am I on to something here or is this so far off?

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+            # individual job, and the experiment can monitor job statuses.
             # pylint: disable-next=protected-access
-            self._active_launchers.add(launch_config._adapted_launcher)
-            return launch_config.start(exe, env)
+            self._active_launchers[launch_config._adapted_launcher].add(id_)
+            return id_
 
         return execute_dispatch(job), *map(execute_dispatch, jobs)
 
+    def get_status(self, *ids: LaunchedJobID) -> tuple[SmartSimStatus, ...]:
```
mellis13, line 236: This function is good, but I wonder how performant it will be if it is called very often with a large amount of jobs present.  

Would it be easier / more efficient if the initial mapping was job_id -> [launcher]?  Or maybe it is better if we put the data structures and logic for mananging this mapping in a class so implmeentation details are hidden a bit.  Maybe something like ``class JobInventory`` (better class names welcome)

smartsim/error/errors.py
```diff
@@ -116,6 +116,10 @@ class LauncherNotFoundError(LauncherError):
     """A requested launcher could not be found"""
 
 
+class UnrecognizedLaunchedJobError(LauncherError):
```
mellis13, line 119: The other error classes start with ``Launcher``.  Would ``LauncherJobNotFound`` be better / more consistent?

smartsim/settings/dispatch.py
```diff
@@ -378,10 +382,16 @@ class ExecutableProtocol(t.Protocol):
     def as_program_arguments(self) -> t.Sequence[str]: ...
 
 
-class LauncherProtocol(t.Protocol[_T_contra]):
-    def start(self, launchable: _T_contra, /) -> LaunchedJobID: ...
+class LauncherProtocol(collections.abc.Hashable, t.Protocol[_T_contra]):
     @classmethod
```
mellis13, line 386: Can we add doc strings to this class?

tests/temp_tests/test_settings/conftest.py
```diff
@@ -52,11 +52,16 @@ def launcher_str(self):
 @pytest.fixture
 def mock_launcher():
     class _MockLauncher(dispatch.LauncherProtocol):
+        __hash__ = object.__hash__
```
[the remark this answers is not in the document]
MattToast: I added `t.Hashable` to the `LauncherProtocol` protocol, to make sure that we could catch some more potential type errors at static analysis time. Most objects in python are hashable, but you can technically make them unhashable by setting `__hash__` to `None` (common examples are `list`, `dict`, `set`, etc.) so I wanted to make sure the type checker could catch this corner case.

BUT! Unfortunately, by adding `t.Hashable` to the protocol it also made `LauncherProtocol.__hash__` an `abstractmethod` that needs to be implemented if the protocol is directly subclassed, like it is here. If you don't the class cannot be instantiated.

This line is basically just say "yes! I'm _really_ sure that I want to use the basic hash algorithm for a normal python object" which is basically just hashing on the object's `id`. If we think this too confusing or too much of a pain to write, I could very easily be talked into taking `t.Hashable` out of the protocol and letting it be a "soft" requirement, or adding `__hash__` to the protocol ourselves, but not marking the method as abstract!

smartsim/settings/dispatch.py
MattToast: Agreed offline that working over collections is necessary.

MattToast: Agreed offline to make this protocol method return a `Mapping`.

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+        associated with the launched job id, then a
+        `SmartSimStatus.STATUS_NEVER_STARTED` status is returned for that id.
+
+        If the experiment maps the launched job id to multiple launchers, then
+        a `ValueError` is raised. This should only happen in the case when
+        launched job ids issued by user defined launcher are not sufficiently
+        unique.
+
+        :param ids: A sequence of launched job ids issued by the experiment.
+        :returns: A tuple of statuses with order respective of the order of the
```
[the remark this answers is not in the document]
MattToast: Yes I do!

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+            calling arguments.
+        """
+        ids_ = set(ids)
```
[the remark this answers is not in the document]
MattToast: Absolutely!

> Sorry I always ask that MERP

Don't be sorry! If its not immediately obvious what's happening, that's a good sign comments are needed. I don't want to over comment, so I only tend to comment sections that reviewers ask for more details on!!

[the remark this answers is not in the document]
MattToast: Basically I'm just offloading the work of figuring out what needs comments onto you, the reviewer, lol.

smartsim/error/errors.py
MattToast: Somehow I totally missed that pattern! I'll rename to conform!

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+        """Get the status of jobs launched through the `Experiment` from their
+        launched job id returned when calling `Experiment.start`.
+
+        The `Experiment` will map the launched ID back to the launcher that
+        started the job and request a status update. The order of the returned
+        statuses exactly matches the order of the launched job ids.
+
+        If the `Experiment` cannot find any launcher that started the job
+        associated with the launched job id, then a
+        `SmartSimStatus.STATUS_NEVER_STARTED` status is returned for that id.
```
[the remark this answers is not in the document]
MattToast: Agreed offline to split up the `SmartSimStatus` enum into a `SmartSimStatus` enum and an `SmartSimStatusError` (name pending) enum. Launchers will be able to report a `SmartSimStatus`, but any invalid input would return a `SmartSimStatusError` enum.

The reason for this split is because it was requested that this method does not raise!

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+
+        If the experiment maps the launched job id to multiple launchers, then
```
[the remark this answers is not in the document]
MattToast: Moved to using a new data structure for tracking launch history. This error can now no longer be raised!

smartsim/experiment.py
MattToast: Moved to a new `LaunchHistory` data structure that stores a mapping of `job_id` -> `launcher` that issued the id

smartsim/experiment.py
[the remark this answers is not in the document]
MattToast: Removed the need to raise from this method!

smartsim/settings/dispatch.py
MattToast: Added!

smartsim/_core/control/jobmanager.py
```diff
@@ -242,7 +242,7 @@ def get_status(
                 job: Job = self[entity.name]  # locked
                 return job.status
 
-            return SmartSimStatus.STATUS_NEVER_STARTED
+            return FailedToFetchStatus.STATUS_NEVER_STARTED
```
mellis13, line 245: I like the change to FiledToFetchStatus.  Is "never started" most correct, or is it just something more like "status not found"

smartsim/experiment.py
```diff
@@ -218,15 +223,73 @@ def execute_dispatch(job: Job) -> LaunchedJobID:
...
+                If you are not using a custom launcher, and this error
+                persists, please file an issue at the following link, and we'll
+                be sure to help address the problem as soon as possible:
+
+                https://github.com/CrayLabs/SmartSim/issues
+                """))
+
+        ids_as_sequence = ((l, tuple(ids)) for l, ids in to_query)
+        stats_zips = (zip(ids, l.get_status(*ids)) for l, ids in ids_as_sequence)
+        stats_map = dict(itertools.chain.from_iterable(stats_zips))
```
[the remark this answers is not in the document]
MattToast: Right now this is just returning the enum value, but if/when we start polling and printing job statuses, we can absolutely print out the enum value and not the enum instance itself!

smartsim/_core/utils/helpers.py
```diff
@@ -436,6 +437,22 @@ def first(predicate: t.Callable[[_T], bool], iterable: t.Iterable[_T]) -> _T | N
...
 
 
+def unique(iterable: t.Iterable[_HashableT]) -> t.Iterable[_HashableT]:
+    seen = set()
+    for item in filter(lambda x: x not in seen, iterable):
+        seen.add(item)
+        yield item
+
+
+def group_by(
```
mellis13, line 447: Doct strings?

smartsim/_core/utils/helpers.py
```diff
@@ -436,6 +437,22 @@ def first(predicate: t.Callable[[_T], bool], iterable: t.Iterable[_T]) -> _T | N
     return next((item for item in iterable if predicate(item)), None)
 
 
+def unique(iterable: t.Iterable[_HashableT]) -> t.Iterable[_HashableT]:
```
mellis13, line 440: Doc strings?

mellis13 (approved):
Very small comments otherwise LGTM

smartsim/_core/utils/helpers.py
MattToast: 🤦 docs are hard

MattToast: Added!

smartsim/_core/utils/helpers.py
MattToast: Added!

PR #658, opened 2024-08-07
Move `ShellLauncher`

juliaputko:


smartsim/_core/shell/__init__.py
```diff
@@ -0,0 +1,25 @@
+# BSD 2-Clause License
+#
+# Copyright (c) 2021-2024, Hewlett Packard Enterprise
```
mellis13, line 3: Wondering if it would be more consistent for the ``ShellLauncher`` to be in ``_core/launcher/shellLauncher.py`` since DragonLauncher is in ``_core/launcher/DragonLauncher.py``.  Thoughts?

mellis13 (changes requested):
One quick question before approving

smartsim/_core/shell/__init__.py
juliaputko: Having it under _core/launcher was causing a circular import error!

smartsim/_core/dispatch.py
```diff
@@ -433,22 +433,4 @@ def impl(
     return impl
 
 
-class ShellLauncher:
```
MattToast, line 436: Could we actually move the `make_shell_format_fn` to the shell launcher module as well? It's really just meant as a short hand to format `LaunchArguments` instances to input for the `ShellLauncher`

smartsim/_core/shell/shellLauncher.py
```diff
@@ -0,0 +1,65 @@
...
+
+import typing as t
+
+from smartsim.types import LaunchedJobID
+
+from smartsim.log import get_logger
+
+if t.TYPE_CHECKING:
+    from typing_extensions import Self
+    from smartsim.experiment import Experiment
```
MattToast, line 38: Super minor style nit: I tend to like have any `if t.TYPE_CHECKING:` blocks below any "real" imports just so `isort` can do a better job of grouping together common imports

smartsim/_core/dispatch.py
MattToast: Follow up question: did we want to move `create_job_id`, `ExecutableProtocol` and `LauncherProtocol` under some other different namespace? They're not strictly wrong here because they are mostly only used here so far, but I tend to think all three have more to do with "launching" rather than "dispatching".

smartsim/_core/shell/shellLauncher.py
```diff
@@ -0,0 +1,65 @@
...
+        # pylint: disable-next=consider-using-with
+        self._launched[id_] = sp.Popen((helpers.expand_exe_path(exe), *rest))
+        return id_
+
+    @classmethod
+    def create(cls, _: Experiment) -> Self:
+        return cls()
+    
+
+# <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```
MattToast, line 65: Nit: Feel free to drop this comment since it's no longer delimiting the end of a block!

MattToast (changes requested):
A couple of small nits and a couple of philosophy questions on where we think some of the protocols should "live", but otherwise looks about ready to go on my end!

smartsim/_core/dispatch.py
juliaputko: I am down! do you have any suggestions for where to move them to?

smartsim/_core/shell/shellLauncher.py
juliaputko: changed

smartsim/_core/shell/shellLauncher.py
juliaputko: removed

smartsim/_core/dispatch.py
MattToast: moved to `smartsim._core.utils.launcher`

smartsim/_core/shell/shellLauncher.py
```diff
@@ -0,0 +1,155 @@
...
+    @classmethod
+    def create(cls, _: Experiment) -> Self:
+        return cls()
+
+    @staticmethod
+    def make_shell_format_fn(
+        run_command: str | None,
+    ) -> _FormatterType[
+        LaunchArguments, tuple[str | os.PathLike[str], t.Sequence[str]]
+    ]:
```
MattToast, line 108: Not wrong, but I thought I would ask the group for a preference (@juliaputko @amandarichardsonn @mellis13):

Methods like these (that aren't strictly used as helper functions within a class) I tend to prefer to leave as top level functions of a namespace rather than static methods on a class. I think we originally had this as a top level function, but was moved to a `staticmethod` in this PR.

Is there a group/team wide consensus on where methods like these live? Like I mentioned, its mostly just a question of philosophy, but probably best to be consistent!

MattToast (approved):
Looks great! I have one small philosophy question, but nothing worth holding up approval over.

LGTM!! Thanks for helping to keep us organized!

mellis13 (approved):

PR #661, opened 2024-08-08
Use `torch.from_numpy` instead of `torch.tensor` to reduce a copy

AlyssaCote:
During memory profiling we found that `torch.tensor` makes a copy and `torch.from_numpy` does not. We only use `torch.tensor` in `torch_worker.py`, so this is a very small change.

I also added in a change to `mock_app.py` after some discussion that we may not want to return tensors, but we can have more discussion surrounding this if necessary.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -13,6 +13,7 @@ Jump to:
 
 Description
 
+- Reduce a copy by using torch.from_numpy instead of torch.tensor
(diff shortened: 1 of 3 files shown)


mellis13 (approved):
LGTM -- thanks!

al-rigazzi (approved):
LGTM! Thanks for doing this!

PR #664, opened 2024-08-12
sample

ankona:


ankona:
closing PR w/sample implementation of signal handler registration in `class Service`

PR #667, opened 2024-08-14
Remove pydantic dependency from MLI code

AlyssaCote:
We only used pydantic BaseModels in `FeatureStoreKey`, so I converted that into a dataclass and used `_post_init_` to validate that the key and descriptor are not empty strings. I also made the dataclass frozen.

### DIFF
--- a/smartsim/_core/mli/infrastructure/storage/featurestore.py
+++ b/smartsim/_core/mli/infrastructure/storage/featurestore.py
@@ -27,19 +27,20 @@
 import typing as t
 from abc import ABC, abstractmethod
 
-from pydantic import BaseModel, Field
+from dataclasses import dataclass
+
 
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
-
-class FeatureStoreKey(BaseModel):
+@dataclass
+class FeatureStoreKey:
     """A key,descriptor pair enabling retrieval of an item from a feature store"""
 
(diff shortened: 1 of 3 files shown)


smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -26,22 +26,32 @@
...
-from pydantic import BaseModel, Field
+from dataclasses import dataclass
 
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
 
-class FeatureStoreKey(BaseModel):
+@dataclass(frozen=True)
```
mellis13, line 36: Do we want ``frozen=True``?  I'm not familiar with the ``BaseModel`` -- is that frozen by default?

If frozen is not used, then I would assume that we need to add setter methods to check that any updates adhere to non-empty string

mellis13 (commented):
Just one question/comment.  Thanks for the update!

smartsim/_core/mli/infrastructure/storage/featurestore.py
AlyssaCote: `BaseModels` are not frozen by default, but I think we definitely want a frozen dataclass because we don't want `FeatureStoreKey` instances to be updated.

ankona (approved):
LGTM

PR #668, opened 2024-08-14
Merge develop into smartsim-refactor

juliaputko:


MattToast (changes requested):
Looks like this is still failing to build on the CI. It looks like you are using undefined symbols `install`, `build_py`, and `Distribution` in the `setup.py` (they came from removed imports).

It looks like these were removed from `develop`, do you still need them here??

MattToast (approved):
Looks great on my end! Just one last obligatory reminder to please put this in with a merge commit to prevent future conflicts, hahaha.

LGTM!! Thanks for handling this, it'll be nice to have a passing CI again!

PR #669, opened 2024-08-15
Refactor RedisAI Build

ashao:
- The RedisAIBuilder class was completely overhauled to allow users to
  express a wider range of support for hardware/software stacks. This 
  will be extended to support ROCm, CUDA-11, and CUDA-12.
- Versions for each of these packages are no longer specified in an
  internal class. Instead a default set of JSON files specifies the
  sources and versions. Users can specify their own custom specifications
  at smart build time


smartsim/_core/_install/utils.py
```diff
@@ -0,0 +1,161 @@
...
+        extract
+
+        :param source: URL or path to find the package
+        :param destination: where to place the package
+        :raises UnsupportedArchive: Unknown archive type
+        :raises PathNotFound: Path to archive does not exist
+        """
+        url_scheme = urlparse(str(source)).scheme
+        if str(source).endswith(".git"):
+            cls._from_git(str(source), destination, **retrieve_kwargs)
```
MattToast, line 140: Consider using the `ParseResult` path directly and stripping off any trailing `/` when making this comparison:
```suggestion
        parse_res = urlparse(str(source))
        url_scheme = parse_res.scheme
        if parse_res.path.rstrip('/').endswith(".git"):
            cls._from_git(str(source), destination, **retrieve_kwargs)
```

smartsim/_core/_install/utils.py
```diff
@@ -0,0 +1,161 @@
...
+                    ".zip",
+                    ".tgz",
+                ):
+                    cls._from_local_archive(source, destination, **retrieve_kwargs)
+                else:
+                    raise UnsupportedArchive(
+                        f"Source ({source}) is not a supported archive or directory "
+                    )
+            else:
+                raise PathNotFound(f"Package path or file does not exist: {source}")
```
MattToast, line 161: super minor style thing: Consider inverting your error checks to avoid deeply nesting:
```suggestion
            if not source_path.exists():
                raise PathNotFound(f"Package path or file does not exist: {source}")
            if source_path.is_dir():
                cls._from_local_directory(source, destination, **retrieve_kwargs)
            elif source_path.is_file() and source_path.suffix in (
                ".gz",
                ".zip",
                ".tgz",
            ):
                cls._from_local_archive(source, destination, **retrieve_kwargs)
            else:
                raise UnsupportedArchive(
                    f"Source ({source}) is not a supported archive or directory "
                )
```

MattToast: if you want, I would also consider making the `elif` blocks just `if` blocks with early returns to avoid the need to nest on the `else` clause here

tests/install/test_package_retriever.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    assert pathlib.Path(f"{test_dir}/smartsim_git").is_dir()
+
+
+def test_https(test_dir):
+    output_dir = pathlib.Path(test_dir) / "output"
+    PackageRetriever.retrieve(
+        "https://github.com/CrayLabs/SmartSim/archive/refs/tags/v0.6.0.zip", output_dir
+    )
+    assert False
```
MattToast, line 106: Test looks like it will always fail?

tests/test_builder.py
```diff
@@ -0,0 +1,39 @@
...
+
+import smartsim._core._install.platform as platform
+
+# The tests in this
+#  file belong to the group_a group
+pytestmark = pytest.mark.group_a
+
+
+def test_default_configs():
+    print(platform.DEFAULT_PLATFORMS)
```
MattToast, line 39: Test, in theory, is unfalsifiable

MattToast (changes requested):
Looks great! Such a nicer build system than was there previously!!

I just left some initial feedback while we resort out the CI and get everything building and passing. As always, feel free to lmk what you think!!

smartsim/_core/_cli/build.py
```diff
@@ -412,50 +284,30 @@ def execute(
...
-            )
-    except (SetupError, BuildError) as e:
-        logger.error(str(e))
-        return os.EX_SOFTWARE
+    # REDIS/KeyDB
+    build_database(build_env, versions, keydb, verbose)
 
     backends = installed_redisai_backends()
+    if backends:
+        logger.warning("RedisAI was previously built, run 'smart clean' to rebuild")
```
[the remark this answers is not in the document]
ashao: Good catch yes because you can install DLPack which at least gives you the ability to send/retrieve tensors.

smartsim/_core/_cli/build.py
```diff
@@ -412,50 +284,30 @@ def execute(
...
-            check_py_tf_version(versions)
-        if "onnxruntime" in backends:
-            check_py_onnx_version(versions)
-    except (SetupError, BuildError) as e:
-        logger.error(str(e))
-        return os.EX_SOFTWARE
+    if args.install_python_packages:
+        for package in mlpackages.values():
+            logger.info(f"Installing python packages for {package.name}")
+            package.pip_install(quiet=not verbose)
```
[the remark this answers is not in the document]
ashao: I think we might have to live with a little bit of oddities like this. We're going to be supporting different versions of the packages for the different versions. I'm actually not entirely sure why we depend on protobuf itself. Let me play around

smartsim/_core/_cli/build.py
```diff
@@ -125,212 +112,72 @@ def build_database(
...
-    validate_installed_version = validate_installed_version or BuildEnv.check_installed
-    missing, conflicts = _assess_python_env(
-        packages,
-        package_pinning,
-        validate_installed_version,
-    )
+    logger.info("Building RedisAI and backends...")
+    RAIBuilder = RedisAIBuilder(platform, mlpackages, build_env, verbose)
+    RAIBuilder.build()
+    RAIBuilder.cleanup_build()
```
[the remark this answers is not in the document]
ashao: I'm a little bit wary of doing either of those two things, because if there was an Exception thrown during the build, we need to have the build directory around to debug the problem. I suppose we could go the context manager route and only cleanup on success. What's your preffered path?

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,266 @@
...
+        self.build_env = build_env
+        self.verbose = verbose
+        self.source = source
+        self.version = version
+        self.patches: t.List[_RedisAIPatch] = []
+        self._define_patches_by_version()
+
+        self.src_path = CONFIG.build_path / "RedisAI" / "src"
+        self.build_path = CONFIG.build_path / "RedisAI" / "build"
+        self.package_path = CONFIG.build_path / "RedisAI" / "mlpackages"
```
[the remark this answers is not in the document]
ashao: I'll use my discretion to decline this suggestion. One of the things that I would love to do in the future is freeze all these variables related to fundamental paths at pip install time and/or smart build time. It's difficult otherwise to ensure that the world hasn't changed underneath the install.

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,266 @@
...
+        self.build_path.mkdir(parents=True)
+        self.package_path.mkdir(parents=True)
+
+        # Create the build directory structure
+        git_kwargs = {
+            "depth": 1,
+            "branch": self.version,
+        }
+
+        PackageRetriever.retrieve(self.source, self.src_path, **git_kwargs)
```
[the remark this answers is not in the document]
ashao: Doh yes, I left this in thinking that we might need to change `git_kwargs` dynamically, but life turned out easier than expected.

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,266 @@
...
+        version: str = "v1.2.7",
+    ) -> None:
+        self.platform = platform
+        self.mlpackages = mlpackages
+        self.build_env = build_env
+        self.verbose = verbose
+        self.source = source
+        self.version = version
+        self.patches: t.List[_RedisAIPatch] = []
+        self._define_patches_by_version()
```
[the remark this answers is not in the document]
ashao: I was kind of anticipating a future where a user might need to add their own patch. I'm half debating ripping this part out and trying to chuck it into the mlpackages JSON somehow. Thoughts?

smartsim/_core/_install/mlpackages.py
```diff
@@ -0,0 +1,182 @@
...
+    configs = {}
+    for config_file in config_file_path.glob("*.json"):
+        dependencies = MLPackageCollection.from_json_file(config_file)
+        configs[dependencies.platform] = dependencies
+    return configs
+
+
+DEFAULT_MLPACKAGE_PATH: pathlib.Path = pathlib.Path(
+    str(resources.files("smartsim._core._install.configs.mlpackages").joinpath(""))
+)
```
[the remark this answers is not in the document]
ashao: Ok, so this one I fought with for a very long time. What gets returned is a `MultiplexedPath` and the only way that I can figure out how to get it to return something meaningful is by using the `joinpath` method on that object. I agree though that we can probably just keep it either a string or a `pathlib.Path`. Do you know of a better way to get the path to this directory?

smartsim/_core/_install/configs/mlpackages/LinuxX64CPU.json
```diff
@@ -0,0 +1,50 @@
...
+                "skl2onnx",
+                "scikit-learn",
+                "onnxmltools"
+            ],
+            "lib_source": "https://github.com/microsoft/onnxruntime/releases/download/v1.18.1/onnxruntime-linux-x64-1.18.1.tgz"
+        }
+    ]
+
+
+}
```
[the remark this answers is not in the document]
ashao: SUFFER

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
ashao: Could also just make it the responsibility of the builder to cleanup the build directory for everything

smartsim/_core/_install/mlpackages.py
```diff
@@ -0,0 +1,215 @@
...
+    def __len__(self) -> int:
+        return len(self.ml_packages)
+
+    def __delitem__(self, key: str) -> None:
+        del self.ml_packages[key]
+
+    def __setitem__(self, key: str, value: MLPackage) -> None:
+        self.ml_packages[key] = value
+
+    # def values(self) -> t.Iterable[MLPackage]:
```
juliaputko, line 156: intentionally leaving these commented?

smartsim/_core/_install/mlpackages.py
```diff
@@ -0,0 +1,215 @@
...
+
+    #     :return: The names of the MLPackages
+    #     """
+    #     return self.ml_packages.keys()
+
+    def tabulate_versions(self, tablefmt: str = "github") -> str:
+        """Display package names and versions as a table
+
+        :param tablefmt: Tabulate format, defaults to "github"
+        :type tablefmt: str, optional
```
juliaputko, line 181: I dont think we need to add type in the docstring anymore? right?

juliaputko: Also tablefmt isnt typed as optional here. Do you want it to be optional?

smartsim/_core/_install/platform.py
```diff
@@ -0,0 +1,241 @@
...
+    :return: Enum associated with the device stack
+    """
+
+    CPU = "cpu"
+    CUDA118 = "cuda-11.8"
+    CUDA121 = "cuda-12.1"
+    ROCM57 = "rocm-5.7"
+
+    @classmethod
+    def from_str(cls, str_: str) -> "Device":
```
juliaputko, line 92: just wanna make sure you want this return type not to be optional, since it is optional elsewhere.

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,260 @@
...
+        self.package_path.mkdir(parents=True)
+
+        retrieve(self.source, self.src_path, depth=1, branch=self.version)
+
+        self._prepare_packages()
+
+        for package in self.mlpackages.values():
+            self._patch_source_files(package.rai_patches)
+        cmake_command = self._rai_cmake_cmd()
+        build_command = self._rai_build_cmd
```
juliaputko, line 145: just makin sure you didnt mean to call this as self._rai_build_cmd(). disregard if I am wrong

tests/install/test_redisai_builder.py
```diff
@@ -0,0 +1,50 @@
...
+pytestmark = pytest.mark.group_a
+
+
+def test_redisai_builder(monkeypatch, test_dir):
+    platform = Platform.from_str("linux", "x86_64", "cpu")
+    mlpackages = DEFAULT_MLPACKAGES[platform]
+    monkeypatch.setattr(PackageRetriever, "retrieve", lambda *args, **kwargs: None)
+    builder = RedisAIBuilder(platform, mlpackages, BuildEnv())
+
+    # builder.run_command = lambda *args, **kwargs: print(args)
```
juliaputko, line 45: comment?

smartsim/_core/_install/mlpackages.py
ashao: t.Optional would be appropriate if it could be None. In this case, it should be string

smartsim/_core/_install/platform.py
ashao: Yes, this should always return something valid (or throw an Exception)

smartsim/_core/_install/redisaiBuilder.py
ashao: Yes, `rai_build_cmd` is a property and not a method on this class. Might need to change this in the future to be more generic, but for now this seems to satisfy all the cases

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,283 @@
...
+
+        logger.info("Configuring CMake Build")
+        if self.verbose:
+            print(" ".join(cmake_command))
+        self.run_command(cmake_command, self.build_path)
+
+        logger.info("Building RedisAI")
+        if self.verbose:
+            print(" ".join(cmake_command))
+        self.run_command(build_command, self.build_path)
```
MattToast, line 165:
```suggestion
        logger.info("Building RedisAI")
        if self.verbose:
            print(" ".join(build_command))
        self.run_command(build_command, self.build_path)
```

tests/test_cli.py
```diff
@@ -439,14 +439,6 @@ def mock_execute(ns: argparse.Namespace, _unparsed: t.Optional[t.List[str]] = No
     "command,mock_location,exp_output,optional_arg,exp_valid,exp_err_msg,check_prop,exp_prop_val",
     [
         pytest.param("build", "build_execute", "verbose mocked-build", "-v", True, "", "v", True, id="verbose 'on'"),
-        pytest.param("build", "build_execute", "cpu mocked-build", "--device=cpu", True, "", "device", "cpu", id="device 'cpu'"),
-        pytest.param("build", "build_execute", "gpu mocked-build", "--device=gpu", True, "", "device", "gpu", id="device 'gpu'"),
-        pytest.param("build", "build_execute", "gpuX mocked-build", "--device=gpux", False, "invalid choice: 'gpux'", "", "", id="set bad device 'gpuX'"),
```
MattToast, line 444: Seems like these are still valid tests. Any reason they were removed?

tests/test_cli.py
```diff
@@ -439,14 +439,6 @@ def mock_execute(ns: argparse.Namespace, _unparsed: t.Optional[t.List[str]] = No
...
-        pytest.param("build", "build_execute", "no tensorflow mocked-build", "--no_tf", True, "", "no_tf", True, id="set no TF"),
-        pytest.param("build", "build_execute", "no torch mocked-build", "--no_pt", True, "", "no_pt", True, id="set no torch"),
-        pytest.param("build", "build_execute", "onnx mocked-build", "--onnx", True, "", "onnx", True, id="set w/onnx"),
```
MattToast, line 447: Same here, can we just change these to test `--skip-{tensorflow,torch,onnx}`

tests/test_cli.py
```diff
@@ -439,14 +439,6 @@ def mock_execute(ns: argparse.Namespace, _unparsed: t.Optional[t.List[str]] = No
...
-        pytest.param("build", "build_execute", "torch-dir mocked-build", "--torch_dir /foo/bar", True, "", "torch_dir", "/foo/bar", id="set torch dir"),
-        pytest.param("build", "build_execute", "bad-torch-dir mocked-build", "--torch_dir", False, "error: argument --torch_dir", "", "", id="set torch dir, no path"),
```
MattToast, line 449: And then maybe add a few here for `--install-python-packages` and/or `--alternate-config-dir`?

MattToast (changes requested):
Looks great! Just a couple of last minute nits and sanity checks, but otherwise looks about ready to go on my end!

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,287 @@
...
+            logger.info(f"Retrieving package: {package.name} {package.version}")
+            target_dir = self.package_path / package.name
+            package.retrieve(target_dir)
+            # Move actual contents to root of the expected location
+            actual_root = self._find_closest_object(target_dir, "include")
+            if actual_root and actual_root != target_dir:
+                logger.debug(
+                    (
+                        "Non-standard location found: \n",
+                        "{str(actual_root)} -> {str(target_dir)}",
```
MattToast, line 214:
```suggestion
                        f"{actual_root} -> {target_dir}",
```

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,287 @@
...
+        DEP_PATH/example_backend. Due to difficulties in retrieval methods and
+        naming conventions from different sources, this cannot be standardized.
+        Instead we try to find the parent of the "include" directory and assume
+        this is the root.
+        """
+
+        for package in self.mlpackages.values():
+            logger.info(f"Retrieving package: {package.name} {package.version}")
+            target_dir = self.package_path / package.name
+            package.retrieve(target_dir)
```
MattToast, line 207: Incredibly minor performance thing if we care, but it looks like the iterations of this loop are totally independent. We could open up each of these io bound processes in a separate thread to make retrieving assets go a bit faster!

MattToast (commented):
Two other very small comments I found while playing around with this PR some more!!

smartsim/_core/_install/mlpackages.py
```diff
@@ -0,0 +1,189 @@
...
+            # Convert the dictionary representation to a RAIPatch
+            if "rai_patches" in ml_package:
+                patch_list = ml_package.pop("rai_patches")
+                ml_package["rai_patches"] = [RAIPatch(**patch) for patch in patch_list]
+
+        ml_packages = {
+            ml_package["name"]: MLPackage(**ml_package)
+            for ml_package in config_json["ml_packages"]
+        }
+        return cls(platform, ml_packages)
```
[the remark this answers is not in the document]
ashao: Holdover from before making it a mutable mapping

tests/test_cli.py
ashao: Hah, I think they were failing so I just deleted them. Can't fail if they don't exist right?

I'll go ahead and fix them up

smartsim/_core/_install/redisaiBuilder.py
```diff
@@ -0,0 +1,283 @@
...
+        """Shell command to build RedisAI and modules
+
+        With the CMake based install, very little needs to be done here.
+        "make install" is used to ensure that all resulting RedisAI backends
+        and their dependencies end up in the same location with the correct
+        RPATH if applicable.
+
+        :return: Command used to compile RedisAI and backends
+        """
+        return "make install -j VERBOSE=1".split(" ")
```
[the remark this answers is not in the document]
ashao: Yeah, otherwise I think it's hard to determine whether we want to be verbose in not capturing output or make the builds noisy. Most of the time when we turn verbose on, we're troubleshooting so this force of `VERBOSE=1` effectively gives us the most output possible

smartsim/_core/config/config.py
```diff
@@ -94,13 +94,26 @@ class Config:
     def __init__(self) -> None:
         # SmartSim/smartsim/_core
         self.core_path = Path(os.path.abspath(__file__)).parent.parent
+        # TODO: Turn this into a property. Need to modify the configuration
+        # of KeyDB vs Redis at build time
+        self.conf_path = Path(self.dependency_path / "config" / "redis.conf")
+        self.conf_dir = Path(self.core_path, "config")
```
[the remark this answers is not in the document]
ashao: I think I changed these to a property and there was something that mypy started yelling about. I believe we kind of monkeypatch this somewhere and I couldn't be bothered to figure out the "right" way to do it. I think we could think about redoing the database builder at a later date and throw this TODO in there.

tests/test_dragon_launcher.py
```diff
@@ -593,11 +593,14 @@ def test_run_step_fail(test_dir: str) -> None:
...
 
-    mock_connector = MagicMock()  # DragonConnector()
+    mock_connector = MagicMock(spec=DragonConnector)
     mock_connector.is_connected = True
     mock_connector.send_request = MagicMock(
         return_value=DragonRunResponse(step_id=step0.name, error_message="mock fail!")
     )
+    mock_connector.merge_persisted_env = MagicMock(
+        return_value={"FOO": "bar", "BAZ": None}
+    )
```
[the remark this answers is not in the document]
ashao: These broke because of the need to update pydantic to fix the mypy error.

smartsim/_core/_cli/validate.py
```diff
@@ -283,10 +267,12 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
     net.eval()
 
     forward_input = torch.rand(1, 1, 3, 3).to(device_)
-    traced = torch.jit.trace(net, forward_input)  # type: ignore[no-untyped-call]
+    traced = torch.jit.trace(  # type: ignore[no-untyped-call, unused-ignore]
+        net, forward_input
+    )
```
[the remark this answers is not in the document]
ashao: Apparently in different versions of Python for the same version of PyTorch this doesn't throw a `no-untyped-call`. The `unused-ignore` allows this to pass

smartsim/_core/_cli/build.py
```diff
@@ -412,50 +282,30 @@ def execute(
...
-            check_py_onnx_version(versions)
-    except (SetupError, BuildError) as e:
-        logger.error(str(e))
-        return os.EX_SOFTWARE
+    logger.info(f"{backends_str} backend(s) available")
+
+    if args.install_python_packages:
+        for package in mlpackages.values():
+            logger.info(f"Installing python packages for {package.name}")
+            package.pip_install(quiet=not verbose)
```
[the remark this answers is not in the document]
ashao: Not entirely sure what you're suggesting here. In the example you gave, TF won't show up in `mlpackages` regardless of whether the user chose `--skip-tensorflow` or not. Are you saying that we should throw a warning since the user might expect that Tensorflow doesn't exist? If so, I suppose we could try to find the union of all ML packages across all the platforms in the config dir. Otherwise, I think we resigning ourselves to keeping a master list of supported packages which I'd prefer to avoid.

smartsim/_core/_install/redisaiBuilder.py
ashao: So I went down too deep a rabbit hole for this, but it looks like we basically saturate the device's bandwidth for each package since the provider's bandwidth (whether GCP, Meta, Github, whatever) is crazy high. Given that, I think we keep this for simplicity of code over performance

smartsim/_core/_install/platform.py
```diff
@@ -0,0 +1,241 @@
...
+        """Return enum associated with the OS
+
+        :param string: String representing the OS
+        :raises UnsupportedError: If not enumerated here, unsupported
+        :return: Enum for a specific OS
+        """
+        string = string.lower()
+        for type_ in cls:
+            if string in type_.value:
+                return type_
```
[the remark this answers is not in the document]
ashao: I think we can just do a direct string check since they're all single valued Enums

smartsim/_core/_cli/build.py
```diff
@@ -368,14 +213,36 @@ def _configure_keydb_build(versions: Versioner) -> None:
...
+    # The user should never have to specify the OS and Architecture
+    current_platform = Platform(
+        OperatingSystem.autodetect(), Architecture.autodetect(), device
+    )
+
+    # Configure the ML Packages
+    if args.alternate_config_dir:
+        configs = load_platform_configs(Path(args.alternate_config_dir))
+    else:
+        configs = DEFAULT_MLPACKAGES
```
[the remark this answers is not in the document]
ashao: Intention was to prevent the packages from being reloaded. Fixed the logic

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA118.json
```diff
@@ -0,0 +1,56 @@
...
+        {
+            "name": "libtorch",
+            "version": "2.4.0",
+            "pip_index": "https://download.pytorch.org/whl/cu118",
+            "python_packages": [
+                "torch==2.4.0+cu118",
+                "torchvision==0.19.0+cu118",
+                "torchaudio==2.4.0+cu118"
+            ],
+            "lib_source": "/home/users/shao/dev/ml_lib_builder/install/libtorch-2.4.0-linux-x64-cuda-11.8.0.tgz",
```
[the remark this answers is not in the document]
ashao: That will be the last commit before merging so I can continue to test

smartsim/_core/_cli/build.py
```diff
@@ -125,212 +111,71 @@ def build_database(
...
-    # TODO: Do not like how the default validation function will always look for
-    #       a `==` pinning. Maybe turn `BuildEnv.check_installed` into a factory
-    #       that takes a pinning and returns an appropriate validation fn?
-    validate_installed_version = validate_installed_version or BuildEnv.check_installed
-    missing, conflicts = _assess_python_env(
-        packages,
-        package_pinning,
-        validate_installed_version,
-    )
+    pattern = r"^([a-zA-Z0-9_\-]+)([<>=!~]+)?([\d\.\*]+)?$"
```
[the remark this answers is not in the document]
ashao: Actually also discovered a need to potentially parse the variant. For example, installing pytorch from the CUDA 11.8 pip index yields variant 2.4.0+cu118

smartsim/_core/_install/redisaiBuilder.py
MattToast: Fair enough! I'm not going to argue with simplicity!!

tests/test_cli.py
MattToast: LMAO, that is getting dangerously to a Toast™©® brand solution!

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA118.json
```diff
@@ -0,0 +1,56 @@
...
+            ]
+        },
+        {
+            "name": "libtensorflow",
+            "version": "2.14.1",
+            "pip_index": "",
+            "python_packages": [
+                "tensorflow==2.14.1"
+            ],
+            "lib_source": "/home/users/shao/dev/ml_lib_builder/install/libtensorflow-2.14.1-linux-x64-cuda-11.8.0.tgz"
```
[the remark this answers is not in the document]
MattToast: resolving per [previous discussion](https://github.com/CrayLabs/SmartSim/pull/669#discussion_r1750085482)

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA118.json
```diff
@@ -0,0 +1,56 @@
...
+            "name": "onnxruntime",
+            "version": "1.17.3",
+            "pip_index": "",
+            "python_packages": [
+                "onnx==1.15",
+                "skl2onnx",
+                "scikit-learn",
+                "onnxmltools"
+            ],
+            "lib_source": "/home/users/shao/dev/ml_lib_builder/install/onnxruntime-1.17.3-linux-x64-cuda-11.8.0.tgz"
```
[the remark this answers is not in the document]
MattToast: resolving per [previous discussion](https://github.com/CrayLabs/SmartSim/pull/669#discussion_r1750085482)

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
MattToast: > Are you saying that we should throw a warning since the user might expect that Tensorflow doesn't exist?

Exactly! My fear is that a user, on a linux machine, would do something to the nature of
```sh
$ smart build --device cpu
```
and they would then install and start using TF (backend and/or wheel).

They then, for whatever reason attempt to install SmartSim locally on their mac laptop using the same command, but then get confused when TF in silently not installed.

Its a very weird corner case, and I could be very easily talked into calling this "user error", especially if we have a compatibility matrix in the docs, but it seems like the kind of thing that _could_ happen and I wasn't sure how much we want to try and save the user from themselves at runtime.

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
MattToast: Updated the logic a bit more to remove `DEFAULT_MLPACKAGES` constant entirely. Now packages in the specified directory are loaded.

smartsim/_core/config/config.py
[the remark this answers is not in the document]
MattToast: ^^ That makes sense, probably best to not try and shove to much into this PR.

smartsim/_core/_install/platform.py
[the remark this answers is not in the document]
MattToast: That's a good point! After you mentioned it, I bothered to do the first bit of research into how python handles linux, and it doesn't look like python has appended the kernel version for linux since like version 3.3, hahaha!

Direct equality checks should be plenty sufficient!

tests/test_cli.py
MattToast: Tests re-added and passing!

tests/test_cli.py
MattToast: Skip tests added and passing!

tests/test_cli.py
MattToast: Test added and passing

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
ashao: I think let's not try to catch it; it's a solvable problem (we could just compile tensorflow and provide it). Like I suggested above, I might be ok with computing (based on the JSON files in the config files) what all possible packages might be, but I think it's also going to get noisy.

So long story short, I think maybe we chalk it up to user error and point to the docs.

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA125.json
```diff
@@ -0,0 +1,64 @@
...
+            "lib_source": "https://github.com/RedisAI/dlpack.git"
+        },
+        {
+            "name": "libtorch",
+            "version": "2.4.0",
+            "pip_index": "https://download.pytorch.org/whl/cu124",
+            "python_packages": [
+                "torch<=2.4.0",
+                "torchvision<=0.19.0",
+                "torchaudio<=2.4.0"
```
MattToast, line 22: At time of writing, from a clean build, these version numbers will currently throw a warning message during the install process despite the fact that the expected version of the package was installed. This is because the [check to make sure the appropriate python packages is installed looks for a strict match of any version suffix](https://github.com/CrayLabs/SmartSim/pull/669/commits/9cefd75c77fe417d0e3d8b957916642873fee805#diff-5b4ed2e1cb9524b32297844cf6e58d61fac9293474767fcb1c701e66e574e53fR162-R167) between the requested and found versions.

We should be able to fix this by appending `+cuda` to the version numbers here, but I wasn't sure if that this is the intended behavior. I just wanted to go ahead and call out this issue while I'm here and we can discuss how we want to resolve the issue.

A similar problem exists for the other gpu and linux+cpu builds of `torch`.

smartsim/_core/_cli/build.py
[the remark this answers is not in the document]
MattToast: I updated how the version number of python packages is checked, but [there is one outstanding issue](https://github.com/CrayLabs/SmartSim/pull/669#discussion_r1755702507) that I wanted to discuss with you before I feel comfortable marking this as 'resolved'.

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA125.json
ashao: Yes, I think I did this for the 11.8. We should just update the JSON files to be clear about the variant.

smartsim/_core/_install/configs/mlpackages/LinuxX64CPU.json
```diff
@@ -0,0 +1,56 @@
...
+        },
+        {
+            "name": "libtorch",
+            "version": "2.4.0",
+            "pip_index": "https://download.pytorch.org/whl/cpu",
+            "python_packages": [
+                "torch<=2.4.0",
+                "torchvision<=0.19.0",
+                "torchaudio<=2.4.0"
+            ],
```
MattToast, line 23: If we are going to get these packages from a the extra index, can we suffix them with `+cpu` just to avoid this warning in the build process?

image

MattToast (approved):
One last minor nit, but nothing worth holding approval up over! Should just need to host those custom backend builds and we should be ready to go!!

smartsim/_core/_cli/build.py
```diff
@@ -321,6 +324,18 @@ def execute(
...
 def configure_parser(parser: argparse.ArgumentParser) -> None:
     """Builds the parser for the command"""
+
+    available_devices = []
+    for platform in DEFAULT_MLPACKAGES:
+        if (
+            # (platform.operating_system == OperatingSystem.autodetect()) and
+            # (platform.architecture == Architecture.autodetect())
+            (platform.operating_system == OperatingSystem.LINUX) and
+            (platform.architecture == Architecture.X64)
```
MattToast, line 334: Sanity checking here: should this be reverted? I'm guessing this should be comparing to the actual hardware

smartsim/_core/_cli/build.py
```diff
@@ -331,7 +346,7 @@ def configure_parser(parser: argparse.ArgumentParser) -> None:
         "--device",
         type=str.lower,
         default=Device.CPU.value,
-        choices=[*(device.value for device in Device), "gpu"],
+        choices=available_devices,
```
MattToast, line 349: Another sanity check: do we want to explicitly disallow `--device=gpu` as an option?

MattToast (approved):
Even more last minute nits

smartsim/_core/_install/configs/mlpackages/LinuxX64CUDA125.json
```diff
@@ -0,0 +1,64 @@
...
+            "lib_source": "https://github.com/RedisAI/dlpack.git"
+        },
+        {
+            "name": "libtorch",
+            "version": "2.4.0",
+            "pip_index": "https://download.pytorch.org/whl/cu124",
+            "python_packages": [
+                "torch==2.4.0+cu125",
+                "torchvision==0.19.0+cu125",
+                "torchaudio==2.4.0+cu125"
```
MattToast, line 22: Sanity check: should these be
```suggestion
                "torch==2.4.0+cu124",
                "torchvision==0.19.0+cu124",
                "torchaudio==2.4.0+cu124"
```
? I don't think a `+cu125` version of these packages exist.

PR #672, opened 2024-08-21
Add event broadcasting capability

ankona:
Event broadcasting will enable the system to notify other MLI resources of changes. This PR contains the base capabilities required for publishing & consuming channel messages as events.


smartsim/_core/mli/comm/channel/channel.py
```diff
@@ -46,9 +46,10 @@ def send(self, value: bytes) -> None:
         :param value: The value to send"""
 
     @abstractmethod
-    def recv(self) -> t.List[bytes]:
+    def recv(self, timeout: int = 0) -> t.List[bytes]:
```
AlyssaCote, line 49: Not this line, but can you do me a favor and change `Receieve` to `Receive` please :)

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
+        """Initialize the DragonCommChannel instance
+
+        :param channel: a channel to use for communications
+        :param recv_timeout: a default timeout to apply to receive calls"""
+        serialized_ch = channel.serialize()
+        safe_descriptor = base64.b64encode(serialized_ch).decode("utf-8")
+        super().__init__(safe_descriptor)
+        self._channel = channel
+        self._recv_timeout = recv_timeout
+
```
AlyssaCote, line 56: Really just curious why we're moving away form attaching with a key? Although maybe I just haven't gotten to that part of the code yet!

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
         :returns: the received message"""
-        with self._channel.recvh(timeout=None) as recvh:
-            message_bytes: bytes = recvh.recv_bytes(timeout=None)
-            return [message_bytes]
+        timeout = timeout or self._recv_timeout
+        with self._channel.recvh(timeout=timeout) as recvh:
+            messages: t.List[bytes] = []
+
+            # todo: consider that this could (under load) never exit. do we need
+            # to configure a maximum number to pull at once?
```
AlyssaCote, line 78: Hmmmm didn't think about this. We'd want all of the messages within the channel though, right?

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -69,10 +104,16 @@ def from_descriptor(
...
-        :param descriptor: The descriptor that uniquely identifies the resource
+        :param descriptor: The descriptor that uniquely identifies the resource. Output
+        from `descriptor_string` is correctly encoded.
         :returns: An attached DragonCommChannel"""
         try:
-            return DragonCommChannel(base64.b64decode(descriptor))
+            utf8_descriptor = descriptor.encode("utf-8")
+            actual_descriptor = base64.b64decode(utf8_descriptor)
+            channel = dch.Channel.attach(actual_descriptor)
+            return DragonCommChannel(channel)
```
AlyssaCote, line 114: I get it now!

smartsim/_core/mli/comm/channel/dragonchannel.py
AlyssaCote: Disregard! I see that that logic was moved to the `from_descriptor` method.

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -24,15 +24,30 @@
...
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
 
+class ReservedKeys(str, enum.Enum):
+    MLI_NOTIFY_CONSUMERS = "_SMARTSIM_MLI_NOTIFY_CONSUMERS"
+
+    @classmethod
+    def from_string(cls, value: str) -> t.Optional["ReservedKeys"]:
```
AlyssaCote, line 42: Can we put some doc strings in here

smartsim/_core/mli/comm/channel/dragonchannel.py
ankona: The factory method `from_descriptor` is for re-creating a `CommChannel` from an existing descriptor. Having this constructor lets us do the initial construction of the object. 

From another perspective, we can just see that the way it was repeated the same thing for the constructor and `from_descriptor`. This keeps it DRY

smartsim/_core/mli/comm/channel/dragonchannel.py
ankona: I started at the bottom and wondered "what does she get!?" - now i get what you get...

smartsim/_core/mli/infrastructure/storage/featurestore.py
ankona: Yes! Done

smartsim/_core/mli/comm/channel/channel.py
ankona: Fixed misspelled word 5x

smartsim/_core/mli/comm/channel/dragonchannel.py
ankona: i think it's possible with a while to never run out of new messages to receive. it doesn't event need to be a ton of buffered messages, just constant. 

i think a `for i in range(max_message_retrievals)` may ensure that it doesn't get stuck and any loop triggering retrieval would end up getting called again, anyway...

tests/test_featurestore.py
```diff
@@ -0,0 +1,795 @@
...
+from smartsim.error import SmartSimError
+from tests.mli.channel import FileSystemCommChannel
+from tests.mli.featurestore import MemoryFeatureStore
+
+if t.TYPE_CHECKING:
+    import conftest
+
+
+# The tests in this file belong to the group_a group
+pytestmark = pytest.mark.group_a
```
AlyssaCote, line 54: I think these tests are getting skipped in group a

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -24,15 +24,39 @@
...
+    MLI_NOTIFY_CONSUMERS = "_SMARTSIM_MLI_NOTIFY_CONSUMERS"
+    """Storage location for the list of registered consumers that will receive
+    events from an EventBroadcaster"""
+
+    @classmethod
+    def from_string(cls, value: str) -> t.Optional["ReservedKeys"]:
+        """Convert a string representation into an enumeration member
+
+        :param value: the string to convert
+        :returns: the enumeration member if the conversion succeeded, otherwise None"""
```
AlyssaCote, line 51: When would this fail?

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -24,15 +24,39 @@
...
 
+from smartsim.error.errors import SmartSimError
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
 
+class ReservedKeys(str, enum.Enum):
+    """Contains constants used to identify all featurestore keys that
+    may not be to used by users. Avoids overwriting system data"""
```
AlyssaCote, line 40: "may not be used". Thank you for this doc string though! It helped me a lot!

tests/dragon/utils/channel.py
```diff
@@ -91,4 +111,5 @@ def from_descriptor(
                 path = pathlib.Path(descriptor.decode("utf-8"))
             return FileSystemCommChannel(path)
         except:
-            print("failed to create FS comm channel: {descriptor}")
+            print(f"failed to create fs comm channel: {descriptor}")
```
ankona, line 114: replace print w/logger.warning

ankona: done!

tests/mli/channel.py
```diff
@@ -62,29 +65,50 @@ def send(self, value: bytes) -> None:
...
         :returns: An attached FileSystemCommChannel"""
         try:
-            path = pathlib.Path(descriptor)
+            if isinstance(descriptor, str):
+                path = pathlib.Path(descriptor)
+            else:
+                path = pathlib.Path(descriptor.decode("utf-8"))
             return FileSystemCommChannel(path)
         except:
             print(f"failed to create fs comm channel: {descriptor}")
```
ankona, line 114: replace print w/logger.warning

ankona: Done!

tests/dragon/test_featurestore.py
```diff
@@ -0,0 +1,153 @@
...
+    directories for individual test outputs"""
+
+    mock_storage = storage_for_dragon_fs
+    backbone = BackboneFeatureStore(mock_storage)
+    mock_fs_descriptor = backbone.descriptor
+
+    # verify ability to write and read from ddict
+    backbone["test_dir"] = test_dir
+
+    print(f"foo value is {backbone['test_dir']}")
```
ankona, line 82: replace print w/assertion

ankona: Done!

smartsim/_core/mli/infrastructure/storage/featurestore.py
ankona: any string not in the enumeration. see `test_mli_reserved_keys_conversion`

tests/test_featurestore.py
ankona: it's ok for them to get skipped if dragon isn't installed). that will be done on purpose with `dragon = pytest.importorskip("dragon")`

However, if the dragon library is there, we can run all of these tests without needing a "real dragon" environment (only need the ability to import the dragon components/modules successfully).  For example, these will run on hotlum w/a normal pytest like `pytest ./tests/test_featurestore.py`

tests/test_featurestore.py
```diff
@@ -0,0 +1,795 @@
...
+        backbone.notification_channels = (target_descriptor,)
+
+        # send a message into the channel
+        with pytest.raises(SmartSimError) as ex:
+            publisher.send(event)
+
+        assert "construct" in ex.value.args[0]
+
+
+def test_eventpublisher_failure(test_dir: str, monkeypatch: pytest.MonkeyPatch) -> None:
```
ankona, line 593: how is this unique from `test_eventpublisher_serialize_failure` above?

tests/test_featurestore.py
```diff
@@ -0,0 +1,795 @@
...
+    storage_path.mkdir(parents=True, exist_ok=True)
+
+    mock_storage = {}
+
+    # note: file-system descriptors are just paths
+    target_descriptor = str(storage_path / "test-consumer")
+
+    backbone = BackboneFeatureStore(mock_storage)
+    comm_channel = FileSystemCommChannel.from_descriptor(target_descriptor)
+    # event = OnCreateConsumer(target_descriptor)
```
ankona, line 704: remove commented code

ankona: Done!

tests/test_featurestore.py
AlyssaCote: Okay so we're cool with these tests being skipped in github actions then. Just wanted to double check.

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,399 @@
...
+        self[ReservedKeys.MLI_NOTIFY_CONSUMERS] = ",".join(
+            [str(value) for value in values]
+        )
+
+
+class EventCategory(str, enum.Enum):
+    """Predefined event types raised by SmartSim backend"""
+
+    CONSUMER_CREATED: str = "consumer-created"
+    FEATURE_STORE_WRITTEN: str = "feature-store-written"
```
AlyssaCote, line 89: This event is when there's a change within a feature store, correct? Like if a model within a feature store changes?

tests/dragon/test_featurestore.py
```diff
@@ -0,0 +1,152 @@
...
+        capp_channel,
+        backbone,
+    )
+    back_consumer = EventConsumer(
+        back_channel,
+        backbone,
+        filters=[EventCategory.CONSUMER_CREATED],
+    )
+
+    # # create some broadcasters to publish messages
```
ankona, line 111: rm double comment

ankona: done!

tests/dragon/test_featurestore.py
```diff
@@ -0,0 +1,152 @@
...
+    # simulate the app updating a model a few times
+    event_2 = OnWriteFeatureStore(backbone.descriptor, "key-1")
+    event_3 = OnWriteFeatureStore(backbone.descriptor, "key-2")
+    event_4 = OnWriteFeatureStore(backbone.descriptor, "key-1")
+
+    mock_client_app.send(event_2)
+    mock_client_app.send(event_3)
+    mock_client_app.send(event_4)
+
+    # # worker manager should only get updates about feature update
```
ankona, line 142: rm double comment

ankona: done!

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: yes. we should create a naming convention per our prior convo but considering leaving as-is and tweaking in integration PR

AlyssaCote: I was thinking make `UPDATED` instead of `WRITTEN`, and `OnUpdateFeatureStore` instead of `OnWriteFeatureStore`. But we can talk more about it in the integration pr.

tests/test_featurestore.py
ankona: I think you've still got a valid point... I didn't achieve what I thought I did, after reviewing the gh action. Investigating...

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,399 @@
...
+        return self._num_discards
+
+    def _save_to_buffer(self, event: EventBase) -> None:
+        """Places a serialized event in the buffer until full. Continues adding events
+        to the buffer using a first-in, first-discarded strategy.
+
+        :param event: The event to serialize and buffer"""
+        if len(self._event_buffer) == self._buffer_size:
+            self._event_buffer.popleft()
+            self._num_discards += 1
```
AlyssaCote, line 226: Definitely worried about throwing away important messages. Do you think that a `buffer_size` of 100 is for sure large enough or do we need to experiment with that number a bit?

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,399 @@
...
+
+        try:
+            event_bytes = bytes(event)
+            self._event_buffer.append(event_bytes)
```
AlyssaCote, line 230: I don't think this is preventable, just want to draw some attention that there will be copies here and I should performance check it in the future!

tests/test_featurestore.py
ankona: Completed test implementation per original intent (failed to modify test body after copy/pasting & updating intent in docstring)

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,399 @@
...
+        self._descriptors = set(x for x in self._backbone.notification_channels if x)
+        if not self._descriptors:
+            logger.warning("No event consumers are registered")
+            return 0
+
+        self._prune_unused_consumers()
+        self._log_broadcast_start()
+
+        num_sent: int = 0
+        next_event: t.Optional[bytes] = self._event_buffer.popleft()
```
AlyssaCote, line 298: It looks like we're doing `popleft` for our next event, but also using `popleft` to discard events when the buffer is full. Is that what we want?

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: I profiled up to a buffer size of 1 million events. That would far exceed my buffering expectations for a single node (by 2 orders of magnitude), but still only results in ~200MB of allocations. 

I think this should alleviate your concern, here?

```py
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   814     99.6 MiB     99.6 MiB           1   @pytest.mark.skipif(DISABLE_PROFILING != 0, reason=profile_reason)
   815                                         @memory_profiler.profile
   816                                         def test_profile_event_buffer_growth(
   817                                             test_dir: str, monkeypatch: pytest.MonkeyPatch
   818                                         ) -> None:
   819                                             """
   820                                             Execute a test to memory profile when a massive event buffer is used.
   821                                         
   822                                             :param test_dir: pytest fixture automatically generating unique working
   823                                             directories for individual test outputs"""
   824                                             # pytest.mark.skipif(DISABLE_PROFILING, "do not profile during test runs")
   825                                         
   826     99.6 MiB      0.0 MiB           1       storage_path = pathlib.Path(test_dir) / "features"
   827     99.6 MiB      0.0 MiB           1       storage_path.mkdir(parents=True, exist_ok=True)
   828                                         
   829     99.6 MiB      0.0 MiB           1       mock_storage = {}
   830                                         
   831                                             # note: file-system descriptors are just paths
   832     99.6 MiB      0.0 MiB           1       target_descriptor = str(storage_path / "test-consumer")
   833     99.6 MiB      0.0 MiB           1       buffer_size = 1000000
   834                                         
   835     99.6 MiB      0.0 MiB           1       backbone = BackboneFeatureStore(mock_storage)
   836     99.6 MiB      0.0 MiB           2       publisher = EventBroadcaster(
   837     99.6 MiB      0.0 MiB           1           backbone,
   838     99.6 MiB      0.0 MiB           1           channel_factory=FileSystemCommChannel.from_descriptor,
   839     99.6 MiB      0.0 MiB           1           buffer_size=buffer_size,
   840                                             )
   841                                         
   842    309.1 MiB      0.0 MiB           2       with monkeypatch.context() as patch:
   843                                         
   844                                                 # cause buffer to grow to massive size by blocking delivery
   845                                                 # patch.setattr(publisher, "_broadcast", lambda: 0)
   846                                         
   847                                                 # backbone.notification_channels = (target_descriptor,)
   848                                         
   849     99.6 MiB      0.0 MiB           1           events = []
   850                                                 # hit max buffer size
   851    309.1 MiB      0.0 MiB     1000001           for i in range(buffer_size):
   852    309.1 MiB    199.0 MiB     1000000               event = OnCreateConsumer(target_descriptor)
   853    309.1 MiB     10.5 MiB     1000000               events.append(event)
   854                                                     # num_sent = publisher.send(event)
   855                                         
   856                                                     # assert num_sent == 0
   857                                                     # assert len(publisher._event_buffer) == i + 1
```

AlyssaCote: This was more of a comment on the bytes conversion in `_save_to_buffer`. It stuck out to me because of my `tobytes()` trauma lol

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: The `deque` is a FIFO queue w/the left being the oldest item.  I believe we're good using this in both cases for the following reasons:

- we want to publish in order, `popleft` is good
- we want to discard the oldest item, `popleft` is good

AlyssaCote: I see now, thanks for the clarification! I think there's something still nagging at me because the next event is either about to be broadcasted or discarded, which feels a bit weird.

AlyssaCote: This might just be a me issue.

AlyssaCote: Hang on I think I'm backwards on something.

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: I used `_WRITTEN` specifically in place of `_UPDATED` so the same event could be used in both cases and avoid registering a handler more than once. It also avoids the client having to track all keys written, or checking the ddict to decide if the key already exists. Let's chat about this in mli sync

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+        the supplied event to all registered broadcast consumers
+
+        :param event: an event to publish
+        :returns: the number of events successfully published
+        :raises ValueError: if event serialization fails
+        :raises KeyError: if channel fails to attach using registered descriptors
+        :raises SmartSimError: if any unexpected error occurs during send"""
+        try:
+            self._save_to_buffer(event)
+
```
mellis13, line 315: Extra space?

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+        # send each event to every consumer
+        while next_event is not None:
+            for descriptor in map(str, self._descriptors):
+                comm_channel = self._get_comm_channel(descriptor)
+
+                try:
+                    comm_channel.send(next_event)
+                    num_sent += 1
+                except Exception as ex:
```
mellis13, line 291: If one channel broadcast fails, then an error is raised.  This means that other channels that may be OK will never get the message because it was already popped and not placed back on the deque.  I wonder if we should stored failed sends and still go through other channels?  We would still need to think about if the failed broadcasts should be retired or just thrown away...

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
mellis13, line 315: extra space?

mellis13 (commented):
Just some comments before seeing the integration additions.

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
 
-    def __init__(self, key: bytes) -> None:
-        """Initialize the DragonCommChannel instance"""
-        super().__init__(key)
-        self._channel: dch.Channel = dch.Channel.attach(key)
+    def __init__(self, channel: "dch.Channel", recv_timeout: int = 0) -> None:
+        """Initialize the DragonCommChannel instance
+
+        :param channel: a channel to use for communications
+        :param recv_timeout: a default timeout to apply to receive calls"""
```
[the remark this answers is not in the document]
ankona: Done!

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
         :param value: The value to send"""
         with self._channel.sendh(timeout=None) as sendh:
             sendh.send_bytes(value)
 
-    def recv(self) -> t.List[bytes]:
-        """Receieve a message through the underlying communication channel
+    def recv(self, timeout: int = 0) -> t.List[bytes]:
+        """Receives message(s) through the underlying communication channel
 
+        :param timeout: maximum time to wait for messages to arrive
```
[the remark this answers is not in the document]
ankona: Done!

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
+            except dch.ChannelEmpty:
+                ...  # emptied the queue, swallow this ex
+
+            return messages
+
+    @property
+    def descriptor_string(self) -> str:
+        """Return the channel descriptor for the underlying dragon channel
+        as a string. Automatically performs base64 encoding to ensure the
+        string can be used in a call to `from_descriptor`"""
```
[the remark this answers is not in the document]
ankona: Killed off this property now that i've normalized the descriptor as a string across all of these entities.

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: removed

ankona: removed

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+            [str(value) for value in values]
+        )
+
+
+class EventCategory(str, enum.Enum):
+    """Predefined event types raised by SmartSim backend"""
+
+    CONSUMER_CREATED: str = "consumer-created"
+    FEATURE_STORE_WRITTEN: str = "feature-store-written"
+    UNKNOWN: str = "unknown"
```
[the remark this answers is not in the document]
ankona: unknown was unused. removed!

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+        return pickle.dumps(self)
+
+    def __str__(self) -> str:
+        """Convert the event to a string
+
+        :returns: a string representation of this instance"""
+        return f"{self.uid}|{self.category}"
+
+
+class OnCreateConsumer(EventBase):
```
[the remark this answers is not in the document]
ankona: per offline convo, switching to `Consumer` and `Producer`

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+    def notification_channels(self, values: t.Sequence[str]) -> None:
+        """Set the notification channels to be sent events
+
+        :param values: the list of channel descriptors to save"""
+        self[ReservedKeys.MLI_NOTIFY_CONSUMERS] = ",".join(
+            [str(value) for value in values]
+        )
+
+
+class EventCategory(str, enum.Enum):
```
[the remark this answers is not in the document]
ankona: this is planned but needs to happen in the follow up integration PR to avoid losing changes. added todo that will be cleaned up immediately in follow up branch that already exists

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -24,15 +24,39 @@
...
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 
+from smartsim.error.errors import SmartSimError
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
 
+class ReservedKeys(str, enum.Enum):
```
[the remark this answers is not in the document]
ankona: As you noted, if we put something in a backbone feature store, then instantiate a `DragonFeatureStore` with the same descriptor it will incorrectly allow a write to the reserved location. This is the entire reason for the existence of the `ReservedKeys`. 

- I believe I misused ReservedKeys in `BackboneFeatureStore` when getting property values. I now avoid using the reserved keys for anything but reserved key write checks
-  I did a quick refactor to avoid needing to use it in child classes at all

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -57,6 +81,22 @@ class FeatureStore(ABC):
...
+        self._reserved_write_enabled = False
+
+    def _check_reserved(self, key: str) -> None:
+        """A utility method used to verify access to write to a reserved key
+        in the FeatureStore. Used by subclasses in __setitem___ implementations
+
+        :param key: a key to compare to the reserved keys
+        :raises SmartSimError: if the key is reserved"""
+        reserved_key_match = ReservedKeys.from_string(key)
+        if reserved_key_match and not self._reserved_write_enabled:
```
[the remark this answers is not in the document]
ankona: Your sample combines into the same statement in reversed order, but that order does let me avoid the constructor exception, e.g. `return cls("not-in-there")` so I think it's probably the better route to avoid unnecessary exceptions being raised. Good call.

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -43,24 +43,59 @@
...
-        """Initialize the DragonCommChannel instance"""
-        super().__init__(key)
-        self._channel: dch.Channel = dch.Channel.attach(key)
+    def __init__(self, channel: "dch.Channel", recv_timeout: int = 0) -> None:
+        """Initialize the DragonCommChannel instance
+
+        :param channel: a channel to use for communications
+        :param recv_timeout: a default timeout to apply to receive calls"""
+        serialized_ch = channel.serialize()
+        safe_descriptor = base64.b64encode(serialized_ch).decode("utf-8")
```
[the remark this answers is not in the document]
ankona: renamed to avoid unhelpful name

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -69,10 +104,16 @@ def from_descriptor(
...
+            utf8_descriptor = descriptor.encode("utf-8")
+            actual_descriptor = base64.b64decode(utf8_descriptor)
+            channel = dch.Channel.attach(actual_descriptor)
+            return DragonCommChannel(channel)
         except:
-            logger.error(f"Failed to create dragon comm channel: {descriptor}")
+            logger.error(
+                f"Failed to create dragon comm channel: {descriptor}", exc_info=True
+            )
             raise
```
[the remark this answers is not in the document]
ankona: absolutely. don't want to leak the dragon errors. done!

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
```diff
@@ -0,0 +1,381 @@
...
+    """A DragonFeatureStore wrapper with utility methods for accessing shared
+    information stored in the MLI backbone feature store"""
+
+    def __init__(self, storage: "dragon_ddict.DDict") -> None:
+        """Initialize the DragonFeatureStore instance
+
+        :param storage: A distributed dictionary to be used as the underlying
+        storage mechanism of the feature store"""
+        super().__init__(storage)
+        self._reserved_write_enabled = True
```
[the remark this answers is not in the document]
ankona: Great call for additional documentation. It's not clear yet what I intended.

- added better doc string
- added properties for modifying value from subclasses
- added to `FeatureStore` base class init params

tests/test_featurestore.py
ankona: I'm modifying the action to always install the dragon package if the platform supports it. This will let us run tests that use modules that import dragon but also don't require a running dragon environment.

al-rigazzi (changes requested):
Just one thing that I'd like to be addressed, namely the different timeouts. The rest looks very promising!

tests/test_featurestore.py
AlyssaCote: I don't see those changes yet but I'm cool with that direction!

AlyssaCote (commented):
A few more tiny comments :)

smartsim/_core/mli/comm/channel/channel.py
```diff
@@ -46,14 +47,16 @@ def send(self, value: bytes) -> None:
...
 
+        :param timeout: maximum time to wait for messages to arrive
         :returns: the received message"""
 
     @property
     def descriptor(self) -> bytes:
         """Return the channel descriptor for the underlying dragon channel"""
         if isinstance(self._descriptor, str):
-            return self._descriptor.encode("utf-8")
+            # return self._descriptor.encode("utf-8")
```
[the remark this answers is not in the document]
ankona: Done!

smartsim/_core/mli/infrastructure/storage/featurestore.py
```diff
@@ -57,29 +81,105 @@ class FeatureStore(ABC):
...
+            )
+
     def __getitem__(self, key: str) -> t.Union[str, bytes]:
         """Retrieve an item using key
 
         :param key: Unique key of an item to retrieve from the feature store"""
+        try:
+            return self._get(key)
+        except KeyError as ex:
+            # logger.warning(f"An unknown key was requested: {key}")
```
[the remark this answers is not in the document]
ankona: gone!

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -39,24 +39,59 @@
...
 
     def send(self, value: bytes) -> None:
         """Send a message throuh the underlying communication channel
         :param value: The value to send"""
         with self._channel.sendh(timeout=None) as sendh:
             sendh.send_bytes(value)
 
-    def recv(self) -> t.List[bytes]:
-        """Receieve a message through the underlying communication channel
+    def recv(self, timeout: int = 0) -> t.List[bytes]:
```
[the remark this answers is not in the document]
ankona: Fixed inconsistent base class type hint, fixed docstrings where necessary

smartsim/_core/mli/comm/channel/dragonchannel.py
```diff
@@ -39,24 +39,59 @@
...
-            message_bytes: bytes = recvh.recv_bytes(timeout=None)
-            return [message_bytes]
+        with self._channel.recvh(timeout=timeout) as recvh:
+            messages: t.List[bytes] = []
+
+            # todo: consider that this could (under load) never exit. do we need
+            # to configure a maximum number to pull at once?
+            try:
+                timeout = timeout or self._recv_timeout
+                while message_bytes := recvh.recv_bytes(timeout=timeout):
```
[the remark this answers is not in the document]
ankona: per offline convo, removed the instance variables and relied on per-call parameters

smartsim/_core/mli/comm/channel/dragonchannel.py
ankona: - Completed extraction of the looping within a channel. 
- Updated tests as necessary.
- Implemented a batch timeout loop in the `EventConsumer` and added tests.

smartsim/_core/mli/infrastructure/storage/backbone_feature_store.py
```diff
@@ -0,0 +1,405 @@
...
+            # remove any empty messages that will fail to decode
+            msg_bytes_list = [msg for msg in msg_bytes_list if msg]
+
+            msg: t.Optional[EventBase] = None
+            if msg_bytes_list:
+                for message in msg_bytes_list:
+                    msg = pickle.loads(message)
+
+                    if not msg:
+                        print("unable to unpickle message")
```
AlyssaCote, line 390: `logger.debug` instead of `print`?

smartsim/_core/mli/infrastructure/storage/dragon_feature_store.py
```diff
@@ -47,48 +47,38 @@ def __init__(self, storage: "dragon_ddict.DDict") -> None:
...
-        except KeyError:
-            logger.warning(f"An unknown key was requested: {key}")
-            raise
-        except Exception as ex:
-            # note: explicitly avoid round-trip to check for key existence
-            raise SmartSimError(
-                f"Could not get value for existing key {key}, error:\n{ex}"
-            ) from ex
+    def _get(self, key: str) -> t.Union[str, bytes]:
+        """Retrieve a value from the underlying stroage mechanism
```
AlyssaCote, line 59: stroage -> storage

AlyssaCote (commented):
Just a couple more comments :)

smartsim/_core/mli/infrastructure/storage/dragon_feature_store.py
ankona: typo retyped

smartsim/_core/mli/infrastructure/storage/backbonefeaturestore.py
ankona: per offline discussion, handling in follow-up PR

smartsim/_core/mli/comm/channel/dragon_channel.py
```diff
@@ -28,47 +28,136 @@
...
-            return [message_bytes]
+        with self._channel.recvh(timeout=timeout) as recvh:
+            messages: t.List[bytes] = []
+
+            try:
+                message_bytes = recvh.recv_bytes(timeout=timeout)
+                messages.append(message_bytes)
+                logger.debug(f"DragonCommChannel {self.descriptor!r} received message")
+            except dch.ChannelEmpty:
+                ...  # emptied the queue, swallow this ex
```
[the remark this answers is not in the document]
ankona: nice catch. done

smartsim/_core/mli/comm/channel/dragon_channel.py
```diff
@@ -28,47 +28,136 @@
...
-        with self._channel.recvh(timeout=None) as recvh:
-            message_bytes: bytes = recvh.recv_bytes(timeout=None)
-            return [message_bytes]
+        with self._channel.recvh(timeout=timeout) as recvh:
+            messages: t.List[bytes] = []
+
+            try:
+                message_bytes = recvh.recv_bytes(timeout=timeout)
+                messages.append(message_bytes)
+                logger.debug(f"DragonCommChannel {self.descriptor!r} received message")
```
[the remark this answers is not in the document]
ankona: w00t.

al-rigazzi (approved):
LGTM!

smartsim/_core/mli/comm/channel/dragon_channel.py
```diff
@@ -28,47 +28,140 @@
...
+
 import smartsim._core.mli.comm.channel.channel as cch
+from smartsim.error.errors import SmartSimError
 from smartsim.log import get_logger
 
 logger = get_logger(__name__)
 
 import dragon.channels as dch
 
+DEFAULT_CHANNEL_BUFFER_SIZE = 500
```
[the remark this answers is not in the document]
ankona: Fantastic q. Amanda let me know that if I add comments, they will be included. Added a comment to ensure this constant is documented

PR #675, opened 2024-08-26
fix init reordering bug

ankona:
Fix 3 bugs:
1. reordering the init sequence in the dragon backend resulted in an un-set collection being used
2. fix tests that should have been updated to compare set contents instead of individual items
3. remove newly added validation on empty host lists that broke existing tests


al-rigazzi (commented):
LGTM, thanks for this fix!

al-rigazzi (approved):

PR #676, opened 2024-08-26
Added type checking to params on model

juliaputko:


smartsim/entity/model.py
```diff
@@ -46,6 +46,24 @@
...
+def _read_model_parameters(params_dict: t.Dict[str, t.Any]) -> t.Dict[str, str]:
+    """Convert the values in a params dict to strings
+    :raises TypeError: if params are of the wrong type
+    :return: param dictionary with values and keys cast as strings
+    """
+    param_names: t.List[str] = []
+    parameters: t.List[t.List[str]] = []
+    for name, val in params_dict.items():
+        param_names.append(name)
+        if isinstance(val, (int, str)):
```
ashao, line 58: All number types should be valid, so let's just abstract it to:
```
import numbers
...
if isinstance(val, (str, numbers.Number)):
```

Alternatively, if we wanted to be even more generic we could just check to see if the object can be converted to a string:

```
if hasattr(val, "__str__"):
...
else:
  raise TypeError(f"Parameter {name} cannot be converted to a string")
```

smartsim/entity/model.py
```diff
@@ -46,6 +46,24 @@
...
+            parameters.append(str(val))
+        else:
+            raise TypeError(
+                "Incorrect type for ensemble parameters\n" + "Must be int, or string."
```
ashao, line 62: `ensemble` -> `model`

`int, or string` to `numeric value or string`

smartsim/entity/model.py
```diff
@@ -46,6 +46,24 @@
 logger = get_logger(__name__)
 
 
+def _read_model_parameters(params_dict: t.Dict[str, t.Any]) -> t.Dict[str, str]:
```
ashao, line 49: Call this `_parse_model_parameters` or `_set_model_parameters`?

tests/test_model.py
```diff
@@ -176,3 +177,17 @@ def test_models_batch_settings_are_ignored_in_ensemble(
     step_cmd = step.step_cmds[0]
     assert any("srun" in tok for tok in step_cmd)  # call the model using run settings
     assert not any("sbatch" in tok for tok in step_cmd)  # no sbatch in sbatch
+
```
ashao, line 180: Split this test up (right now you're testing both the happy and bad routes) and parametrize the bad routes.

```
import numpy as np

@pytest.mark.parametrize("dtype", [int, np.float32, str]):
def test_good_params(dtype):
  params = {"foo": dtype(0.6)}
  assert all(isinstance(val, str) for val in _read_model_parameters(params).values())
  
@pytest.mark.parametrize("bad_val", [["eggs"], {"n": 5}, object)
def test_bad_model_params(bad_val):
  with pytest.raises(TypeError):
    _read_model_parameters({"foo": bad_val})
```

ashao (changes requested):
Thanks for taking care of this. Some suggestions to generalize the check and to split up the tests.

smartsim/entity/model.py
juliaputko: changed

smartsim/entity/model.py
juliaputko: changed to parse

tests/test_model.py
juliaputko: changed

smartsim/entity/model.py
juliaputko: changed to numbers.number

doc/changelog.md
```diff
@@ -15,6 +15,7 @@ To be released at some future point in time
 
 Description
 
+- Add type check to model params
```
ashao, line 18: Make these more descriptive (you're doing more than just type checking):

```suggestion
Allow specifying Model and Ensemble parameters with number-like types (e.g. numpy types)
```

doc/changelog.md
```diff
@@ -31,6 +32,8 @@ Description
 
 Detailed Notes
 
+- Added type checking to params on model
```
ashao, line 35:
```suggestion
The serializer would fail if a parameter for a Model or Ensemble was specified as a numpy dtype. The constructors for these methods now validate that the input is number-like and convert them to strings
```

ashao (approved):
Thanks for covering this! Minor changes to make the changelog more descriptive, otherwise it's good to go

doc/changelog.md
juliaputko: changed

doc/changelog.md
juliaputko: changed

PR #687, opened 2024-08-29
Refactor `exception_handler` to avoid unnecessary building and serialization of failure responses.

AlyssaCote:
In this PR I fix the `exception_handler` so that it only builds and serializes a failure response if a reply channel is not None. I also needed to tweak the tests a bit by mocking out the reply channels.

### DIFF
--- a/smartsim/_core/mli/infrastructure/control/error_handling.py
+++ b/smartsim/_core/mli/infrastructure/control/error_handling.py
@@ -61,10 +61,10 @@ def exception_handler(
         f"Exception type: {type(exc).__name__}\n"
         f"Exception message: {str(exc)}"
     )
-    serialized_resp = MessageHandler.serialize_response(
-        build_failure_reply("fail", failure_message)
-    )
     if reply_channel:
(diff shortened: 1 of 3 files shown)


mellis13 (approved):
Thanks!

smartsim/_core/mli/infrastructure/control/error_handling.py
```diff
@@ -61,10 +61,10 @@ def exception_handler(
...
-        build_failure_reply("fail", failure_message)
-    )
     if reply_channel:
+        serialized_resp = MessageHandler.serialize_response(
+            build_failure_reply("fail", failure_message)
+        )
         reply_channel.send(serialized_resp)
     else:
-        logger.warning("Unable to notify client of error without reply_channel")
+        logger.warning("Unable to notify client of error without a reply channel")
```
al-rigazzi, line 70: Yep, you're right!

al-rigazzi (approved):
LGTM, another step towards super-short response times

PR #690, opened 2024-08-30
Pin watchdog version to prevent mypy errors

ashao:
The release of watchdog v5 introduced new types which caused further errors with mypy. To mitigate these errors for now, we pin the watchdog version to 4.x and will resolve these errors in the future.


al-rigazzi (approved):
LGTM, thanks for this important fix!

PR #698, opened 2024-09-09
Clean up error handling in MLI

AlyssaCote:


smartsim/_core/mli/comm/channel/dragon_fli.py
```diff
@@ -93,6 +98,8 @@ def recv(self, timeout: float = 0.001) -> t.List[bytes]:
                     )
                 except fli.FLIEOT:
                     eot = True
+                except Exception:
+                    break
```
al-rigazzi, line 102: Should we re-raise?

smartsim/_core/mli/infrastructure/control/device_manager.py
```diff
@@ -66,15 +66,23 @@ def remove_model(self, key: str) -> None:
...
         """Get the model corresponding to a given key.
 
         :param key: The model key
         :returns: The model for the given key
         """
-        return self._models[key]
+        try:
+            return self._models[key]
+        except KeyError:
+            logger.warning(f"An unknown key was requested: {key}")
```
al-rigazzi, line 84: Shouldn't we raise this?

al-rigazzi: (or let the function throw an Exception)

smartsim/_core/mli/infrastructure/worker/torch_worker.py
```diff
@@ -142,6 +170,17 @@ def execute(
...
+
+        :param batch: The batch of requests that triggered the pipeline
+        :param load_result: The result of loading the model onto device memory
+        :param transform_result: The result of transforming inputs for model consumption
+        :param device: The device on which the model will be executed
+        :returns: The result of inference wrapped in an ExecuteResult
+        :raises SmartSimError: If model is not loaded
+        :raises IndexError: If memory slicing is out of range
+        :raises ValueError: If tensor creation fails or is unable to evaluate the model
+        """
```
al-rigazzi, line 183: Thanks for all the new docstrings, they look terrific!

al-rigazzi (changes requested):
A couple of questions about the expected behavior when encountering some exceptions, otherwise looks really great!

smartsim/_core/mli/comm/channel/dragon_fli.py
AlyssaCote: I was going back and forth on this, but I think you're right. I was thinking if there was an error that occurred after we got some messages, then we should just send the messages that we have. But really we should be informing the user that there was an error because the data will be incomplete and possibly break down the road. Fixing!

smartsim/_core/mli/infrastructure/control/device_manager.py
AlyssaCote: I have a `raise` under the `logger.warning`. Do you mean like raise a `KeyError` with a different message or something?

smartsim/_core/mli/infrastructure/control/error_handling.py
```diff
@@ -63,10 +66,13 @@ def exception_handler(
     :param reply_channel: The channel used to send replies
     :param failure_message: Failure message to log and send back
     """
+    if failure_message is None:
+        failure_message = str(exc)
     logger.exception(
```
ankona, line 71: What is missing from a standard `logger.exception` output that causes us to have this custom format string? doesn't this result in duplication in the final message? Here's a quick example showing the traceback logged twice, the exception type included twice, etc:

```
>>> try:
...  raise_ex()
... except Exception as ex:
...  logger.exception(f"Exception type: {type(ex).__name__}\nTraceback: {traceback.format_exc()}")
... 
ERROR:root:Exception type: Exception
Traceback: Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in raise_ex
Exception
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in raise_ex
Exception
```

AlyssaCote: No you're totally right. This started out as me wanted to ensure that our failure message gets logged with the exception but now it's just kind of crazy. I think I just need to simply do 
```
logger.exception(exc)
```

but still set the `failure_message` to `str(exc)` if there wasn't one passed in so we can send it back in the response.

smartsim/_core/mli/infrastructure/worker/torch_worker.py
```diff
@@ -123,12 +145,18 @@ def transform_input(
...
+            try:
+                mem_view[:alloc_size] = b"".join(
+                    [
+                        fetch_result.inputs[result_tensor_idx]
+                        for fetch_result in fetch_results
+                    ]
+                )
+            except IndexError as e:
+                raise IndexError(
+                    f"Error accessing elements in fetch_result.inputs "
```
ankona, line 157: unnecessary f""

smartsim/_core/mli/message_handler.py
```diff
@@ -401,8 +401,18 @@ def serialize_request(request: request_capnp.RequestBuilder) -> bytes:
...
         :param request: Request to be serialized
         :returns: Serialized request bytes
+        :raises ValueError: If serialization fails
         """
-        return request.to_bytes()
+        display_name = request.schema.node.displayName  # type: ignore
+        class_name = display_name.split(":")[-1]
+        if class_name != "Request":
+            raise ValueError(
+                "Error serializing the request. Value passed in is not a request"
```
ankona, line 410: we should log identifying info, where possible, to help fix the defect. we might want the `class_name` value here so we know where the mistake came from.

ditto on others, like .... in load_model, `model_id.key`, `device`, etc

AlyssaCote: To be super clear, you mean like adding like `f"Error serializing the request. Value passed in is not a request: {request}"` in this example, and not using the actual logger to log more stuff?

smartsim/_core/mli/infrastructure/control/device_manager.py
al-rigazzi: Oh, I think I missed the `raise` line! Should we add `:raises KeyError:` to the docstring?

al-rigazzi (changes requested):
Just a last nit-pick and then it looks good to go!

smartsim/_core/mli/infrastructure/control/device_manager.py
al-rigazzi: (same comment applies for `remove_model`)

AlyssaCote: fixed x2!

al-rigazzi (approved):
Thanks for addressing that last detail! LGTM!

PR #699, opened 2024-09-10
Update MLI docstrings part 2

AlyssaCote:
Part 2 of docstring updates! Keeping the description as a refresher.

I added and updated the MLI docstrings so that everything renders correctly and they are more consistent. This means I needed to make some decisions on what "consistent" means, because there were many different combinations going on. Feel free to push back on these! I just went with what looked like the best/what we were doing the most.

- The docstring description uses punctuation to end the description. 
- The `:param:`, `:returns:`, etc do not use punctuation to end their description line UNLESS the line needs multiple sentences, in which case, use punctuation. (I think this happens once or twice).
- All description lines are capitalized. 

example:
```py
def serialize_response(response: response_capnp.ResponseBuilder) -> bytes:
   """
   Serializes a built response message.

   :param response: Response to be serialized
   :returns: Serialized response bytes
   """
   return response.to_bytes()
```


smartsim/_core/mli/comm/channel/dragon_channel.py
```diff
@@ -83,37 +84,42 @@ def create_local(capacity: int = 0) -> dch.Channel:
...
             logger.debug(f"DragonCommChannel {self.descriptor!r} sent message")
 
     def recv(self, timeout: float = 0.001) -> t.List[bytes]:
-        """Receives message(s) through the underlying communication channel
+        """Receives message(s) through the underlying communication channel.
 
-        :param timeout: maximum time to wait (in seconds) for messages to arrive
-        :returns: the received message"""
+        :param timeout: Maximum time to wait (in seconds) for messages to arrive
+        :returns: The received message
```
al-rigazzi, line 121: For consistency with the docstring, should we say `The received message(s)`?

smartsim/_core/mli/comm/channel/dragon_fli.py
```diff
@@ -66,19 +66,21 @@ def __init__(
...
             logger.debug(f"DragonFLIChannel {self.descriptor!r} sent message")
 
     def recv(self, timeout: float = 0.001) -> t.List[bytes]:
-        """Receives message(s) through the underlying communication channel
+        """Receives message(s) through the underlying communication channel.
 
-        :param timeout: maximum time to wait (in seconds) for messages to arrive
-        :returns: the received message"""
+        :param timeout: Maximum time to wait (in seconds) for messages to arrive
+        :returns: The received message
```
al-rigazzi, line 82: Same `(s)` question

smartsim/_core/mli/comm/channel/dragon_fli.py
```diff
@@ -98,10 +100,12 @@ def from_descriptor(
         cls,
         descriptor: str,
     ) -> "DragonFLIChannel":
-        """A factory method that creates an instance from a descriptor string
+        """A factory method that creates an instance from a descriptor string.
 
         :param descriptor: The descriptor that uniquely identifies the resource
-        :returns: An attached DragonFLIChannel"""
+        :returns: An attached DragonFLIChannel
+        :raises: If creation of DragonFLIChanenel fails
```
al-rigazzi, line 107: What type of error does it raise?

al-rigazzi: (if we raise whatever is raised by the underlying call, do we need to document it? Asking because I really don't know what we should do. Should we re-cast the Exception to something we own?)

smartsim/_core/mli/infrastructure/environment_loader.py
```diff
@@ -72,7 +73,8 @@ def get_backbone(self) -> t.Optional[FeatureStore]:
         an environment variable. The backbone is a standalone, system-created
```
al-rigazzi, line 73: I know you did not write this, but since we know (and mention in `:returns:`) that the env variable is `_SMARTSIM_INFRA_BACKBONE`, should we write it here?

smartsim/_core/mli/infrastructure/environment_loader.py
```diff
@@ -90,7 +92,8 @@ def get_queue(self) -> t.Optional[CommChannelBase]:
         """Attach to a queue-like communication channel using the descriptor
```
al-rigazzi, line 92: Same question about explicitly calling out the env variable

smartsim/_core/mli/infrastructure/storage/backbone_feature_store.py
```diff
@@ -316,14 +326,15 @@ def _broadcast(self, timeout: float = 0.001) -> int:
...
-        :param timeout: maximum time to wait (in seconds) for messages to send
-        :returns: the number of events successfully published
-        :raises ValueError: if event serialization fails
-        :raises KeyError: if channel fails to attach using registered descriptors
-        :raises SmartSimError: if any unexpected error occurs during send"""
+        the supplied event to all registered broadcast consumers.
+
+        :param event: An event to publish
+        :param timeout: Maximum time to wait (in seconds) for messages to send
+        :returns: The number of events successfully published
```
al-rigazzi, line 333: Just wanted to note that we are inconsistent with the usage of `:returns:` vs `:return:`. Both are fine for Sphinx, anyhow.

al-rigazzi (commented):
Looks great! I have a few comments that I'd like to be addressed and then I think it's good to merge

smartsim/_core/mli/comm/channel/dragon_fli.py
AlyssaCote: Oops I forgot to come back to this one! I actually don't know what error this raises, but I could raise a `SmartSimError`. As far as what to do about documenting exceptions and where, that's something I'm trying to work through in my error handling ticket as well. If we explicitly raise something here then I think we document it.

smartsim/_core/mli/infrastructure/storage/backbone_feature_store.py
AlyssaCote: I changed them all to be `:returns:` because that seems to be used more!

al-rigazzi (approved):
LGTM, thanks for taking care of those small details I pointed out!

PR #709, opened 2024-09-17
MLI helper methods

AlyssaCote:
In this PR I added helper methods in the form of `@property`s to classes like `InferenceRequest` and `InferenceReply`. I had also written helper methods for `RequestBatch`, but mypy would complain about not knowing the type for the attributes when trying to use them after using my helper methods. I do think I may have gone overboard, so if anyone has any push back on these please let me know. Also if there are helper methods anyone remembers that they wanted and I didn't include them, please let me know.


al-rigazzi (approved):
Looks great! Thanks for adding some level of simplicity to future coding!

al-rigazzi:
We're going to rewrite part of `RequestBatch` anyhow, we'll add the helpers then.

PR #712, opened 2024-09-18
Allow for sequences into exp.start(), and unpack iterables

juliaputko:


smartsim/experiment.py
```diff
@@ -151,14 +162,20 @@ def __init__(self, name: str, exp_path: str | None = None):
         experiment
         """
 
-    def start(self, *jobs: Job) -> tuple[LaunchedJobID, ...]:
+    #
```
juliaputko, line 165: I need some guidance with types

tests/test_experiment.py
```diff
@@ -51,6 +53,17 @@
 
 pytestmark = pytest.mark.group_a
 
+_ID_GENERATOR = (str(i) for i in itertools.count())
+
+
+@pytest.fixture
+def get_gen_symlink_dir(fileutils):
+    yield fileutils.get_test_conf_path(osp.join("generator_files", "to_symlink_dir"))
```
MattToast, line 61: Fixture looks to be unused, is it needed?

MattToast (changes requested):
Just some minor typing suggestions, but otherwise looks about ready to go on my end!!

smartsim/experiment.py
```diff
@@ -55,6 +55,17 @@
...
 
+def _unpack(jobs: t.Sequence[t.Tuple[Job]] | t.Sequence[str]) -> t.Iterator[str]:
+    """Unpack any iterable input into exp.start in order to obtain a
+    single sequence of jobs that can be launched"""
+    for item in jobs:
+
+        if isinstance(item, t.Iterable):
+            yield from _unpack(item)
+        else:
+            yield item
```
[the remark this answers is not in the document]
juliaputko: moved it

smartsim/experiment.py
```diff
@@ -151,14 +162,19 @@ def __init__(self, name: str, exp_path: str | None = None):
...
         """Execute a collection of `Job` instances.
 
         :param jobs: A collection of other job instances to start
         :returns: A sequence of ids with order corresponding to the sequence of
             jobs that can be used to query or alter the status of that
             particular execution of the job.
         """
+        # If item is instance iterable then unpack and extend the list
+        if isinstance(jobs, t.Iterable):
+            jobs = list(tuple(_unpack(jobs)))
```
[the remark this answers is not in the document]
juliaputko: removed

smartsim/_core/utils/helpers.py
```diff
@@ -43,18 +43,35 @@
...
 
 _T = t.TypeVar("_T")
 _HashableT = t.TypeVar("_HashableT", bound=t.Hashable)
 _TSignalHandlerFn = t.Callable[[int, t.Optional["FrameType"]], object]
 
+_nested_seq: TypeAlias = "t.Sequence[Job | _nested_seq]"
+
+
+def unpack(value: _nested_seq) -> t.Generator[Job, None, None]:
+    """Unpack any iterable input in order to obtain a
```
amandarichardsonn, line 66: would you add :param: and :return: here?

:param: Sequence containing elements of type Job or other sequences that are also of type _nested_seq
:return: flattened list of Jobs

smartsim/_core/utils/helpers.py
```diff
@@ -43,18 +43,35 @@
...
+    from smartsim.launchable.job import Job
+
     _Ts = TypeVarTuple("_Ts")
 
 
 _T = t.TypeVar("_T")
 _HashableT = t.TypeVar("_HashableT", bound=t.Hashable)
 _TSignalHandlerFn = t.Callable[[int, t.Optional["FrameType"]], object]
 
+_nested_seq: TypeAlias = "t.Sequence[Job | _nested_seq]"
```
amandarichardsonn, line 62: do you mind explaining this type? Idk why my brain is farting but its a sequence of Jobs or input that matches the _nested_seq type?

tests/test_experiment.py
```diff
@@ -611,3 +620,128 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
     assert stat == InvalidJobStatus.NEVER_STARTED
     after_cancel = exp.get_status(*all_known_ids)
     assert before_cancel == after_cancel
+
+
+def test_start_sequences_of_jobs_good(test_dir, wlmutils, monkeypatch, job_list):
```
amandarichardsonn, line 625: Add type hints to test signature

tests/test_experiment.py
```diff
@@ -611,3 +620,128 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+    """Test the unpacking of a tuple of tuples
+    exp.start(job1, (job2, job3))
```
amandarichardsonn, line 627: This test might need to be changed, looks like rn you start a sequence of Jobs but you might need to add some logic to directly test the work you added! This could be removing Ensemble and App inits as well as Experiment init and writing a unit test instead of integration, this would allow you to test the nitty gritty.  Instead you could use unittest.mock.patch to assert that _helpers was called during start by calling assert_called_once() on the mock object, I dont think you need to check the return here. Checking the return might fit better in a test_helper directory and write a test for the _helper where you pass in MockJobs and assert the return is flattened! Your test below where you change the sequence pattern would work really well here in unit test form.

tests/test_experiment.py
```diff
@@ -611,3 +620,128 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+                            launch_settings.LaunchSettings("local"),
+                        ),
+                    ),
+                ]
+            ],
+            id="[job_1, (job_2, job_3)]",
+        ),
+    ),
+)
+def test_start_unpack(test_dir, wlmutils, monkeypatch, job_list):
```
amandarichardsonn, line 738: this test might work better in the helpers test directory where you can actually assert that the sequence was flattened!

amandarichardsonn (changes requested):
Nice work! Got some suggestions for writing test lets go!

smartsim/experiment.py
```diff
@@ -151,19 +151,23 @@ def __init__(self, name: str, exp_path: str | None = None):
...
         """Execute a collection of `Job` instances.
 
         :param jobs: A collection of other job instances to start
         :returns: A sequence of ids with order corresponding to the sequence of
             jobs that can be used to query or alter the status of that
             particular execution of the job.
         """
+        # If item is instance iterable then unpack and extend the list
+
+        jobs_ = list(tuple(_helpers.unpack(jobs)))
```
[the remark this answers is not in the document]
MattToast: This is a common pattern to get around a rule I put in place in mypy that dissallows you changing the type of a variable once it has been assigned.

For e.g.
```py
# example.py
def concat_ints(x: int, y: int) -> int:
    x = str(x)
    y = str(y)
    return int(x + y)
```
when run against mypy with the `--strict` flag (turns on all warnings, basically the equivilient of `-Wall -Werror` for C/C++) will return the following errors:
```
$ mypy --strict example.py
example.py:3: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
example.py:4: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
Found 2 errors in 1 file (checked 1 source file)
```

A common way to get around this problem is to simply append the name of the variable with an underscore to assign to a new variable instead
```py
# example.py
def concat_ints(x: int, y: int) -> int:
    x_ = str(x)
    y_ = str(y)
    return int(x_ + y_)
```

TL;DR: It's coming up with new variable names is hard, lol

smartsim/_core/utils/helpers.py
MattToast: Naming Nit: Traditionally type alias follow the same naming rules as classes (e.g. PascalCase) and should end with `Type`. I think `pylint` will complain about this name if/when we turn that back on. Can we change the name to:
```suggestion
_NestedJobSequenceType: TypeAlias = "t.Sequence[Job | _NestedJobSequenceType]"
```
?

smartsim/experiment.py
MattToast: No need to cast an iterable to a `tuple` and then to a `list`. Consider just casting to a `list` directly!
```suggestion
        jobs_ = list(_helpers.unpack(jobs))
```

smartsim/experiment.py
```diff
@@ -151,19 +151,23 @@ def __init__(self, name: str, exp_path: str | None = None):
...
-    def start(self, *jobs: Job) -> tuple[LaunchedJobID, ...]:
+    def start(self, *jobs: Job | t.Sequence[Job]) -> tuple[LaunchedJobID, ...]:
         """Execute a collection of `Job` instances.
 
         :param jobs: A collection of other job instances to start
         :returns: A sequence of ids with order corresponding to the sequence of
             jobs that can be used to query or alter the status of that
             particular execution of the job.
         """
+        # If item is instance iterable then unpack and extend the list
```
MattToast: Minor comment nit: `jobs` will always be an iterable because it is a varaible length `tuple[Job | Sequence[Job]]` and `tuple`s are always iterable.

tests/test_experiment.py
```diff
@@ -611,3 +620,128 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+    )
+
+    ensemble = Ensemble("ensemble-name", "echo", replicas=2)
+
+    launch_settings = launch_settings.LaunchSettings(wlmutils.get_test_launcher())
+    job_list = ensemble.build_jobs(launch_settings)
+
+    exp = Experiment(name="exp_name", exp_path=test_dir)
+
+    exp.start(job_list)
```
MattToast: Is this test supposed to be parameterized? It looks like it's requesting a `job_list` fixture, but is unable to find one.

If not, feel free to remove `job_list` from the fn signature.


If so, just a heads up, line 637 is overwriting the value before it is used. You may want to drop that line!

MattToast (changes requested):
A couple more small nits in addition to Amanda's comments. Overall, looking really good!

tests/test_experiment.py
```diff
@@ -611,3 +620,111 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+                            exe_args=["spam", "eggs"],
+                        ),
+                        launch_settings.LaunchSettings("local"),
+                    ),
+                    Ensemble("ensemble-name", "echo", replicas=2).build_jobs(
+                        launch_settings.LaunchSettings("local")
+                    ),
+                )
+            ],
+            id="(job1, job2)",
```
MattToast: Nit: `Ensemble.build_jobs` actually returns a tuple of jobs. This ID might be more accurate if changed to
```suggestion
                (
                    job.Job(
                        Application(
                            "test_name",
                            exe="echo",
                            exe_args=["spam", "eggs"],
                        ),
                        launch_settings.LaunchSettings("local"),
                    ),
                    Ensemble("ensemble-name", "echo", replicas=2).build_jobs(
                        launch_settings.LaunchSettings("local")
                    ),
                )
            ],
            id="(job1, (job2, job_3))",
```

tests/test_experiment.py
```diff
@@ -611,3 +620,111 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+                                "test_name_2",
+                                exe="echo",
+                                exe_args=["spam", "eggs"],
+                            ),
+                            launch_settings.LaunchSettings("local"),
+                        ),
+                    ),
+                )
+            ],
+            id="(job1, (job2, job3))",
```
MattToast: Ditto here
```suggestion
                    Ensemble("ensemble-name", "echo", replicas=2).build_jobs(
                        launch_settings.LaunchSettings("local")
                    ),
                    (
                        job.Job(
                            Application(
                                "test_name",
                                exe="echo",
                                exe_args=["spam", "eggs"],
                            ),
                            launch_settings.LaunchSettings("local"),
                        ),
                        job.Job(
                            Application(
                                "test_name_2",
                                exe="echo",
                                exe_args=["spam", "eggs"],
                            ),
                            launch_settings.LaunchSettings("local"),
                        ),
                    ),
                )
            ],
            id="((job1, job2), (job3, job4))",
```

tests/test_experiment.py
```diff
@@ -611,3 +620,111 @@ def test_experiment_stop_does_not_raise_on_unknown_job_id(
...
+                                "test_name_2",
+                                exe="echo",
+                                exe_args=["spam", "eggs"],
+                            ),
+                            launch_settings.LaunchSettings("local"),
+                        ),
+                    ),
+                ]
+            ],
+            id="[job_1, (job_2, job_3)]",
```
MattToast: And then one last one here:
```suggestion
                [
                    job.Job(
                        Application(
                            "test_name",
                            exe="echo",
                            exe_args=["spam", "eggs"],
                        ),
                        launch_settings.LaunchSettings("local"),
                    ),
                    (
                        Ensemble("ensemble-name", "echo", replicas=2).build_jobs(
                            launch_settings.LaunchSettings("local")
                        ),
                        job.Job(
                            Application(
                                "test_name_2",
                                exe="echo",
                                exe_args=["spam", "eggs"],
                            ),
                            launch_settings.LaunchSettings("local"),
                        ),
                    ),
                ]
            ],
            id="[job_1, ((job_2, job_3), job_4)]",
```

MattToast (approved):
A couple of last minute nits about some test ids, I agree with the need for [an additional test to test `unpack` directly](https://github.com/CrayLabs/SmartSim/pull/712#discussion_r1778996594), but nothing there is worth holding approval up over!

LGTM!! Thanks for all the hard work on this one!!

smartsim/_core/utils/helpers.py
```diff
@@ -43,18 +43,39 @@
...
 _TSignalHandlerFn = t.Callable[[int, t.Optional["FrameType"]], object]
 
+_NestedJobSequenceType: TypeAlias = "t.Sequence[Job | _NestedJobSequenceType]"
+
+
+def unpack(value: _NestedJobSequenceType) -> t.Generator[Job, None, None]:
+    """Unpack any iterable input in order to obtain a
+    single sequence of values
+
+    :param: Sequence containing elements of type Job or other
```
amandarichardsonn: This :param: should be :param value:

amandarichardsonn (changes requested):

amandarichardsonn (approved):
LGTM!

PR #714, opened 2024-09-19
Ensemble Move short-task

amandarichardsonn:
The Ensemble class is moved to core/builder/ in this PR -> as_jobs is changed to build_jobs -> and documentation is updated.


smartsim/entity/__init__.py
```diff
@@ -27,7 +27,7 @@
 from .application import Application
 from .dbnode import FSNode
 from .dbobject import *
-from .ensemble import Ensemble
+from .._core.builders.ensemble import Ensemble
```
mellis13, line 30: I think ``_core`` is mostly reserved for non-user infrastructure.  I think maybe ``builders`` should be in top-level SmartSim.  Thoughts?

mellis13 (changes requested):

smartsim/entity/__init__.py
amandarichardsonn: I believe that strategies.py might also need to be moved, am I correct on this?

amandarichardsonn: strategies moved

mellis13 (approved):
LGTM

PR #719, opened 2024-09-25
Change 'conda activate' to 'source activate' for Frontier

ashao:
On Frontier, the recommended way to activate conda environments is
to go through source activate. This also means that ``conda init``
is not needed. The instructions for Frontier have been updated to 
reflect this.


MattToast (approved):
LGTM!!

PR #722, opened 2024-09-26
TensorFlow and ONNX Workers

al-rigazzi:
This PR adds TensorFlow and ONNX workers to the MLI. It also slightly modifies the throughput example mock applications.


smartsim/_core/mli/comm/channel/dragon_fli.py
```diff
@@ -94,7 +94,7 @@ def recv(self, timeout: float = 0.001) -> t.List[bytes]:
         with self._fli.recvh(timeout=timeout) as recvh:
             while not eot:
                 try:
-                    message, _ = recvh.recv_bytes(timeout=timeout)
+                    message, _ = recvh.recv_bytes(timeout=None)
```
mellis13: I think in Chris' broadcast PR this timeout was left as ``timeout=timeout`` -- what ist the reason for the change?

smartsim/_core/mli/comm/channel/dragon_channel.py
```diff
@@ -131,7 +131,7 @@ def recv(self, timeout: float = 0.001) -> t.List[bytes]:
             messages: t.List[bytes] = []
 
             try:
-                message_bytes = recvh.recv_bytes(timeout=timeout)
+                message_bytes = recvh.recv_bytes(timeout=None)
```
mellis13: Same question as dragon_fli.py -- just wondering why the change to the ``timeout`` behavior

smartsim/_core/mli/infrastructure/control/worker_manager.py
```diff
@@ -265,6 +273,7 @@ def _on_iteration(self) -> None:
                 return
 
             for request, transformed_output in zip(batch.requests, transformed_outputs):
+                self._sent_responses += 1
```
mellis13, line 276: should ``_set_responses`` be incremented after the responses are sent in case errors are raised?  (simliar question for the integer tracking batch processing)

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -195,12 +195,22 @@ def has_output_keys(self) -> bool:
 class LoadModelResult:
     """A wrapper around a loaded model."""
 
-    def __init__(self, model: t.Any) -> None:
+    def __init__(
+        self,
+        model: t.Any,
+        inputs: t.Optional[t.List[str]] = None,
```
mellis13, line 200: Just pointing this out, although I think maybe we don't bother to address in this PR, but I think ``LoadModelResult`` should probably not contain all of the customizations for all of the ML backends.  Maybe we address this as part of torch geometric support?  If we post an issue regarding TorchGeometric, maybe we add this as an acceptance criteria?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,316 @@
...
+    """A worker that executes a TensorFlow model."""
+
+    @staticmethod
+    def load_model(
+        batch: RequestBatch, fetch_result: FetchModelResult, device: str
+    ) -> LoadModelResult:
+        """Given a loaded MachineLearningModel, ensure it is loaded into
+        device memory.
+
+        :param request: The request that triggered the pipeline
```
mellis13, line 76: fetch_result is missing from param doc strings (same is missing from torch worker if you have time to update)

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,316 @@
...
+    @staticmethod
+    def transform_input(
+        batch: RequestBatch,
+        fetch_results: list[FetchInputResult],
+        mem_pool: MemoryPool,
+    ) -> TransformInputResult:
+        """Given a collection of data, perform a transformation on the data and put
+        the raw tensor data on a MemoryPool allocation.
+
+        :param request: The request that triggered the pipeline
```
mellis13: Is this doc string out of date?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,316 @@
...
+                item_meta is None for item_meta in fetch_result.meta
+            ):
+                raise ValueError("Cannot reconstruct tensor without meta information")
+            first_tensor_desc: tensor_capnp.TensorDescriptor = fetch_result.meta[0]
+            num_samples = first_tensor_desc.dimensions[0]
+            slices.append(slice(total_samples, total_samples + num_samples))
+            total_samples = total_samples + num_samples
+
+            if res_idx == len(fetch_results) - 1:
+                # For each tensor in the last input, get remaining dimensions
```
mellis13, line 187: Any thoughts on may where we could enforce this requirement?  I could see this being ignored later by a testing case / user.  Doesn't necessarily need to be now, we could post a ticket.

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,316 @@
...
+        transformed_list: list[TransformOutputResult] = []
+        cpu_predictions = execute_result.predictions
+
+        for result_slice in execute_result.slices:
+            transformed = []
+            for cpu_item in cpu_predictions:
+                try:
+                    transformed.append(cpu_item[result_slice].tobytes())
+
+                    # todo: need the shape from latest schemas added here.
```
mellis13, line 309: Is this todo still applicable

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,316 @@
...
+        with tf.device(device):
+            result = LoadModelResult(
+                tf.compat.v1.Session(graph=graph),
+                list(input_layers),
+                list(output_tensors),
+            )
+        return result
+
+    @staticmethod
+    def transform_input(
```
mellis13, line 149: In general, this method seems super dense and difficult to parse.  Although, I think the logic is correct is there any clear way to add some secondar structs or objects to make this easier to test and grok?

mellis13 (changes requested):
Just a couple of comments/questions

smartsim/_core/mli/comm/channel/dragon_fli.py
al-rigazzi: Yeah, this is something that we need to address. The two timeouts have very different semantics (from what I understood), the `recvh` one determines how long we wait for an incoming message, the `recv_bytes` one determines how long we wait for a message to be received. There is no value which is good for both: setting them both to a small value means that we cannot transfer large messages, setting them both to a large value means that we wait for too long (and in loops this penalizes us).

smartsim/_core/mli/comm/channel/dragon_channel.py
al-rigazzi: Answered on `dragon_fli.py`.

smartsim/_core/mli/infrastructure/worker/worker.py
al-rigazzi: We may add some catch-all field, which varies in structure depending on the worker/toolkit?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: Indeed.

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: Yeah, you're right, and it also shares a lot of code with the Torch version -- although I don't know how much will be shared by other workers. We may create some helper methods - but also possibly introduce a class between `TensorFlowWorker` and `MachineLearningWorkerBase`, for those workers which can concatenate tensors this way? Thoughts?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: Yeah, you're right, and it also shares a lot of code with the Torch version -- although I don't know how much will be shared by other workers. We may create some helper methods - but also possibly introduce a class between TensorFlowWorker and MachineLearningWorkerBase, for those workers which can concatenate tensors this way? Thoughts?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: We may do an explicit check at the beginning of this function, and this is a requirement which holds for all classes for which request tensors can be concatenated.

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: Thanks

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: Yes, I believe we have a ticket for it, otherwise we need to create it.

al-rigazzi: The interplay with the `Client` will have to be factored in.

smartsim/_core/mli/infrastructure/control/worker_manager.py
al-rigazzi: I'm a bit on the fence about this one. Maybe I could change the variable name to `_processed_requests`? Or maybe I _add_ that and do as you suggested? I believe both statistics are of interest (I did this with some future telemetry in mind).

ex/high_throughput_inference/mock_app.py
```diff
@@ -101,43 +108,49 @@ def run_model(self, model: bytes | str, batch: torch.Tensor):
...
+        self.perf_timer.measure_time("receive_response")
+        response = MessageHandler.deserialize_response(resp[0])
+        self.perf_timer.measure_time("deserialize_response")
+        # list of data blobs? recv depending on the len(response.result.descriptors)?
+        if len(resp) > 1:
+            data_blob = resp[1]
+        else:
+            data_blob: bytes = self._from_worker_ch.recv(timeout=None)[0]
+        self.perf_timer.measure_time("receive_tensor")
+        result = torch.from_numpy(
```
AlyssaCote: I think a change I had made a while ago got reversed in one of the merges! I had changed it so that `result` is a `numpy array` instead of a `torch tensor`. Should we go back to that or is a torch tensor what we need to return now?

al-rigazzi: Totally right, I must have missed that! Let me revert it to numpy array!

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
 import torch
 
 from smartsim.log import get_logger
+from smartsim._core.mli.client.protoclient import ProtoClient
 
 torch.set_num_interop_threads(16)
 torch.set_num_threads(1)
 
 logger = get_logger("App")
 logger.info("Started app")
```
ankona, line 40: do you want to move this until the app is actually started (e.g. in `if __name__ == "__main__"`)? This would log on import and feels misleading (has the app actually started?)

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
 
-from collections import OrderedDict
-
-from smartsim.log import get_logger, log_to_file
-from smartsim._core.mli.client.protoclient import ProtoClient
 
 logger = get_logger("App")
```
ankona, line 43: repeated call to `get_logger`

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
         :param name: The name to use for the model
         :param model: The path to the pre-trained PyTorch model"""
-        self._model = torch.jit.load(model)
+        self._model = None  # torch.jit.load(model)
         self._name = name
-        buffer = io.BytesIO()
-        scripted = torch.jit.trace(self._model, self.get_batch())
-        torch.jit.save(scripted, buffer)
+
+        with open(model, "rb") as model_file:
```
ankona, line 57: it's a minor annoyance to ensure that a local copy of the model is in a certain place with a specific name. would it be a lot of work to build/load the model on demand and avoid the path parameter?

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
 logger = get_logger("App")
 
 
-CHECK_RESULTS_AND_MAKE_ALL_SLOWER = False
-
-
 class ResNetWrapper:
     """Wrapper around a pre-rained ResNet model."""
+
     def __init__(self, name: str, model: str):
```
ankona, line 49: Could we use a `pathlib.Path` instead of string (or PathLike) to let the type hints give another hint on what the "model" parameter value should be

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
         self._name = name
-        buffer = io.BytesIO()
-        scripted = torch.jit.trace(self._model, self.get_batch())
-        torch.jit.save(scripted, buffer)
+
+        with open(model, "rb") as model_file:
+            buffer = io.BytesIO(model_file.read())
         self._serialized_model = buffer.getvalue()
 
+    # pylint: disable-next=no-self-use
```
ankona, line 61: Any reason we don't just make the method static and avoid disabling the linter?

ex/high_throughput_inference/mock_app.py
```diff
@@ -96,6 +82,11 @@ def name(self) -> str:
         return self._name
 
 
+def log(msg: str, rank_: int) -> None:
```
ankona, line 85: This method probably needs a docstring. It's not obvious from usage that we'll only see logs from rank 1.

ex/high_throughput_inference/mock_app.py
```diff
@@ -24,55 +24,41 @@
...
-        torch.jit.save(scripted, buffer)
+
+        with open(model, "rb") as model_file:
+            buffer = io.BytesIO(model_file.read())
         self._serialized_model = buffer.getvalue()
 
+    # pylint: disable-next=no-self-use
     def get_batch(self, batch_size: int = 32):
         """Create a random batch of data with the correct dimensions to
         invoke a ResNet model.
```
ankona, line 64:
```suggestion
        invoke a ResNet model.
        :param batch_size: Number of samples to be generated
        :returns: A tensor with dimensions (`batch_size`, 3, 224, 224)
```

ex/high_throughput_inference/mock_app.py
```diff
@@ -96,6 +82,11 @@ def name(self) -> str:
         return self._name
 
 
+def log(msg: str, rank_: int) -> None:
+    if rank_ == 0:
+        logger.info(msg)
```
ankona, line 87: Consider using built-in functionality of the logging system. An example might be:

```py
import logging
import typing as t


class MpiRankLogFilter(logging.Filter):
    """A logging filter that ensures duplicate logs are not created
    for every MPI rank running an application."""

    def init(self, ranks_to_log: t.List[int]) -> None:
        """Initialize the log filter.
        
        :param ranks_to_log: MPI ranks that will produce logs
        """
        self._to_log = [rank for rank in ranks_to_log]
        """MPI ranks that will produce logs"""
        self._rank = -1
        """Rank for the host attempting to log"""

    def filter(self, record: logging.LogRecord) -> bool:
        """Allow logging only from specific MPI ranks

        :param record: the record to evaluate for filtering
        :returns: True if the MPI rank is the desired rank for logging, False otherwise
        """
        if self._rank == -1:
            comm_world = MPI.COMM_WORLD
            self._rank = comm_world.Get_rank()

        return self._rank in self._to_log


logger = logging.getLogger("foo")
logger.addFilter(MpiRankLogFilter([0])

```

This would avoid having extra log methods and increases our ability to filter logs within smartsim instead of just the driver. Potentially, the only benefit is using a standard interface (but even that might be enough).

ankona: - Does this point to us executing the wrong thing on all the ranks?
- If it executes on multiple ranks, why wouldn't you want the logs?
- Would a log format that includes rank help to identify the source?

ex/high_throughput_inference/mock_app.py
```diff
@@ -105,38 +96,24 @@ def name(self) -> str:
...
 
-    for log2_bsize in range(args.log_max_batchsize + 1):
+    for log2_bsize in range(args.log_max_batchsize, args.log_max_batchsize + 1):
         b_size: int = 2**log2_bsize
-        logger.info(f"Batch size: {b_size}")
-        for iteration_number in range(TOTAL_ITERATIONS + int(b_size == 1)):
-            logger.info(f"Iteration: {iteration_number}")
-            sample_batch = resnet.get_batch(b_size)
+        log(f"Batch size: {b_size}", rank)
+        for iteration_number in range(TOTAL_ITERATIONS):
```
ankona, line 113: this doesn't really let us parameterize `TOTAL_ITERATIONS` and having the constant buried here doesn't expose it, nicely. 

Would you like to add it as a parameter on `parser` with a default value of 100? That would allow you to vary the iterations without changing the code

ex/high_throughput_inference/mock_app.py
```diff
@@ -105,38 +96,24 @@ def name(self) -> str:
...
-                res_norm = torch.linalg.vector_norm(remote_result, ord=1).item()
-                local_res_norm = torch.linalg.vector_norm(local_res, ord=1).item()
-                logger.info(
-                    f"Avg norm of error {err_norm.item()/b_size} compared to result norm of {res_norm/b_size}:{local_res_norm/b_size}"
-                )
-                torch.cuda.synchronize()
-
-    client.perf_timer.print_timings(to_file=True)
+
+    client.perf_timer.print_timings(to_file=True, to_stdout=rank == 0)
```
ankona, line 119: will the proposed log filter help us avoid having the `to_stdout` parameter? if not, would a `LogHandler` work instead?

ex/high_throughput_inference/mock_app_onnx.py
```diff
@@ -0,0 +1,130 @@
...
+    ):
+        self._get_onnx_model(model)
+        self._name = name
+        self._poly = PolynomialFeatures
+
+    def _get_onnx_model(self, model: onnx.onnx_ml_pb2.ModelProto):
+        self._serialized_model = model.SerializeToString()
+
+    # pylint: disable-next=no-self-use
+    def get_batch(self, batch_size: int = 32):
```
ankona, line 60: Ditto here on using `@staticmethod` instead of disabling the linter

ex/high_throughput_inference/mock_app_onnx.py
```diff
@@ -0,0 +1,130 @@
...
+        """Create a random batch of data with the correct dimensions to
+        invoke a ResNet model.
+
+        :param batch_size: The desired number of samples to produce
+        :returns: A PyTorch tensor"""
+        x = numpy.random.randn(batch_size, 1).astype(numpy.float32)
+        return poly.fit_transform(x.reshape(-1, 1))
+
+    @property
+    def model(self):
```
ankona, line 70:
```suggestion
    def model(self) -> bytes:
```

ex/high_throughput_inference/mock_app_onnx.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @property
+    def name(self):
+        """The name applied to the model.
+
+        :returns: The name"""
+        return self._name
+
+
+def log(msg: str, rank_: int) -> None:
```
ankona, line 84: ditto on log filter

ex/high_throughput_inference/mock_app_onnx.py
```diff
@@ -0,0 +1,130 @@
...
+    comm_world = MPI.COMM_WORLD
+    rank = comm_world.Get_rank()
+    client = ProtoClient(timing_on=True, rank=rank)
+
+    if rank == 0:
+        client.set_model(linreg.name, linreg.model)
+
+    MPI.COMM_WORLD.Barrier()
+
+    TOTAL_ITERATIONS = 100
```
ankona, line 116: ditto on parameterizing total_iterations in the arg parser

ex/high_throughput_inference/mock_app_onnx.py
```diff
@@ -0,0 +1,130 @@
...
+    poly_reg_model = LinearRegression()
+    poly_reg_model.fit(poly_features, p(X))
+
+    onnx_model = to_onnx(poly_reg_model, poly_features, target_opset=13)
+
+    linreg = LinRegWrapper("LinReg", onnx_model)
+
+    comm_world = MPI.COMM_WORLD
+    rank = comm_world.Get_rank()
+    client = ProtoClient(timing_on=True, rank=rank)
```
ankona, line 109: If we're passing rank into the client, we should be able to avoid having conditionals in the app.

e.g. no need for `if rank == 0` if the client can hide already doing that? 

I'm not a big fan of having the client have MPI-specific parameters, but this is simple and works. Minor point in the MVP client.

ex/high_throughput_inference/mock_app_redis.py
```diff
@@ -56,35 +58,61 @@ def model(self):
     def name(self):
         return self._name
 
+
+def log(msg: str, rank: int) -> None:
```
ankona, line 62: ditto, log filter

ex/high_throughput_inference/mock_app_redis.py
```diff
@@ -26,26 +26,28 @@
...
         self._name = name
-        buffer = io.BytesIO()
-        scripted = torch.jit.trace(self._model, self.get_batch())
-        torch.jit.save(scripted, buffer)
+        with open(model, "rb") as model_file:
+            buffer = io.BytesIO(model_file.read())
         self._serialized_model = buffer.getvalue()
 
-    def get_batch(self, batch_size: int=32):
+    def get_batch(self, batch_size: int = 32):
```
ankona, line 50: 1. missing a docstring
2. missing typehint for retval

ex/high_throughput_inference/mock_app_redis.py
```diff
@@ -26,26 +26,28 @@
...
+
+class ResNetWrapper:
     def __init__(self, name: str, model: str):
-        self._model = torch.jit.load(model)
+        self._model = None
         self._name = name
-        buffer = io.BytesIO()
-        scripted = torch.jit.trace(self._model, self.get_batch())
-        torch.jit.save(scripted, buffer)
+        with open(model, "rb") as model_file:
```
ankona, line 46: Can we avoid duplicating the ResNetWrapper in every app?

ex/high_throughput_inference/mock_app_redis.py
```diff
@@ -56,35 +58,61 @@ def model(self):
...
     args = parser.parse_args()
 
-    resnet = ResNetWrapper("resnet50", f"resnet50.{args.device.upper()}.pt")
+    resnet = ResNetWrapper("resnet50", f"resnet50.{args.device}.pt")
 
     client = Client(cluster=False, address=None)
-    client.set_model(resnet.name, resnet.model, backend='TORCH', device=args.device.upper())
 
-    perf_timer: PerfTimer = PerfTimer(debug=False, timing_on=timing_on, prefix=f"redis{rank}_")
+    if rank == 0:
```
ankona, line 81: if we don't like using rank in the client, maybe something like passing a conditional ...

```py
def set_model(self, ..., filter_fn: t.Callable[[], bool]) -> None:
  if not filter_fn():
    return
  ...
```

ex/high_throughput_inference/mock_app_redis.py
```diff
@@ -56,35 +58,61 @@ def model(self):
...
     total_iterations = 100
-    timings=[]
-    for batch_size in [1, 2, 4, 8, 16, 32, 64, 128]:
-        logger.info(f"Batch size: {batch_size}")
-        for iteration_number in range(total_iterations + int(batch_size==1)):
+    timings = []
+    for log2_bsize in range(args.log_max_batchsize, args.log_max_batchsize + 1):
+        batch_size: int = 2**log2_bsize
+        log(f"Batch size: {batch_size}", rank)
+        for iteration_number in range(total_iterations):
```
ankona, line 97: would definitely love to avoid all the iteration code duplicated in these apps, but understand they're only samples... nothing to see here.

ex/high_throughput_inference/mock_app_tensorflow.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if rank == 0:
+        client.set_model(resnet.name, resnet.model)
+
+    comm_world.Barrier()
+
+    TOTAL_ITERATIONS = 100
+
+    for log2_bsize in range(args.log_max_batchsize, args.log_max_batchsize + 1):
+        b_size: int = 2**log2_bsize
```
ankona, line 109: no additional calls to `comm_world.Barrier()` in the loop. Is this what you intended?

ex/high_throughput_inference/redis_driver.py
```diff
@@ -63,4 +91,4 @@
         break
     time.sleep(5)
 
-print("Exiting.")
\ No newline at end of file
+print("Exiting.")
```
ankona, line 94: should we make use of a `logger` instance instead of `print`?

ex/high_throughput_inference/redis_driver.py
```diff
@@ -24,29 +24,57 @@
...
+
+NUM_NODES = args.num_nodes_app
+
+exp_path = os.path.join(
+    filedir,
+    "benchmark",
+    f"redis_ai_multi_n{NUM_NODES}_rpn{NUM_TASKS_PER_NODE}",
+    f"samples{2**args.log_max_batchsize}",
+)
+try:
```
ankona, line 55: this is another good snippet for a prototype utilities module (`def cleanup_exp_dir(path) -> None`) since we repeat the cleanup 3x

ex/high_throughput_inference/redis_driver.py
```diff
@@ -24,29 +24,57 @@
...
+    time.sleep(2)
+except:
+    pass
 
-exp_path = os.path.join(filedir, "redis_ai_multi")
 os.makedirs(exp_path, exist_ok=True)
 exp = Experiment("redis_ai_multi", launcher="slurm", exp_path=exp_path)
 
-db = exp.create_database(interface="hsn0")
+db = exp.create_database(interface="hsn0", hosts=["pinoak0036"])
```
ankona, line 64: does this work for all allocations? is pinoak0036 the head node?

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -124,13 +119,15 @@ def service_as_dragon_proc(
         "--batch_size",
         type=int,
         default=1,
-        help="How many requests the workers will try to aggregate before processing them",
+        help="How many requests the workers will try "
```
ankona, line 122: nitpick... I'm not a fan of "how many requests..." for some reason. Consider

```suggestion
        help="Maximum number of samples to aggregat into a batch inference request"
```

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -124,13 +119,15 @@ def service_as_dragon_proc(
...
+        "to aggregate before processing them",
     )
     parser.add_argument(
         "--batch_timeout",
         type=float,
         default=0.001,
-        help="How much time (in seconds) should be waited before processing an incomplete aggregated request",
+        help="How much time (in seconds) should be waited "
```
ankona, line 129: Ditto on "how much time..." nitpick

```suggestion
        help="Maximum time allowed (in seconds) to aggregate samples "
             "before processing a partial batch",
```

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -143,7 +140,9 @@ def service_as_dragon_proc(
     to_worker_fli = fli.FLInterface(main_ch=to_worker_channel, manager_ch=None)
     to_worker_fli_comm_ch = DragonFLIChannel(to_worker_fli)
 
+    backbone._allow_reserved_writes = True
     backbone.worker_queue = to_worker_fli_comm_ch.descriptor
+    backbone._allow_reserved_writes = False
```
ankona, line 145: dupe sets of `backbone._allow_reserved_writes`. Consider if you need the backbone to allow this worker manager to write system config... you should be able to not set this

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -210,9 +210,11 @@ def service_as_dragon_proc(
...
     all_procs = [dispatcher_proc, *worker_manager_procs]
 
-    print(f"Dispatcher proc: {dispatcher_proc}")
     for proc in all_procs:
         proc.start()
 
     while all(proc.is_alive for proc in all_procs):
         time.sleep(1)
+
+    for proc in all_procs:
```
ankona, line 219: Consider adding a return code (if available). If processes are not alive, would that help guide debugging effort?

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -163,6 +162,7 @@ def service_as_dragon_proc(
         batch_size=args.batch_size,
         config_loader=config_loader,
         worker_type=arg_worker_type,
+        mem_pool_size=128 * 1024**3,
```
ankona, line 165: Could we pass `mem_pool_size` using the `config_loader`? That avoids an extra constructor parameter and keeps access to that config more silo'd. May not apply if you need specific values for specific dispatchers...

ex/high_throughput_inference/standalone_worker_manager.py
```diff
@@ -210,9 +210,11 @@ def service_as_dragon_proc(
     # TODO: use ProcessGroup and restart=True?
```
ankona, line 210: is todo still applicable? can it be removed or turned into a ticket instead?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -532,10 +532,10 @@ def _stop_steps(self) -> None:
                         and proc_group.status == DragonStatus.RUNNING
                     ):
                         try:
-                            proc_group.kill()
+                            proc_group.stop()
                         except dragon_process_group.DragonProcessGroupError:
                             try:
-                                proc_group.stop()
+                                proc_group.kill()
```
ankona, line 538: Does process group API help us avoid extraneous logging (e.g. having "process group already stopped") by adding a conditional?

```py
if proc_group.is_alive:
  proc_group.kill()
```

smartsim/_core/mli/client/protoclient.py
```diff
@@ -179,18 +168,16 @@ def __init__(
...
+        self._from_worker_ch = DragonCommChannel(
+            create_local(self._DEFAULT_WORK_QUEUE_SIZE)
+        )
         self._to_worker_ch = create_local(self._DEFAULT_WORK_QUEUE_SIZE)
 
         self._publisher = self._create_broadcaster()
 
         self.perf_timer: PerfTimer = PerfTimer(
-            debug=False, timing_on=timing_on, prefix=f"a{rank}_"
+            debug=False, timing_on=timing_on, prefix=f"a{self._rank}_"
```
ankona, line 179: should value of `debug` param be based on log level, env var, config_loader, or maybe a parameter to the client? Seems like we might have to track down `PerfTimer` instantiations if we want to change to True

smartsim/_core/mli/comm/channel/channel.py
```diff
@@ -52,19 +52,28 @@ def __init__(
         """A user-friendly identifier for channel-related logging"""
 
     @abstractmethod
-    def send(self, value: bytes, timeout: float = 0.001) -> None:
+    def send(
+        self,
+        value: bytes,
+        timeout: t.Optional[float] = 0.001,
```
ankona, line 58: Just want to confirm our expectation is that None is a valid value for timeout, but not for handle_timeout?

smartsim/_core/mli/infrastructure/control/worker_manager.py
```diff
@@ -171,9 +170,18 @@ def _on_iteration(self) -> None:
...
         try:
-            batch: RequestBatch = self._dispatcher_queue.get(timeout=0.0001)
+            batch: RequestBatch = self._dispatcher_queue.get(timeout=None)
         except Empty:
             return
+        except Exception as exc:
+            exception_handler(
+                exc,
+                None,
+                "Error receiving batch.",
```
ankona, line 180: is there any additional metadata we can add to log msg that could help investigate seeing this in logs?

smartsim/_core/mli/infrastructure/storage/dragon_util.py
```diff
@@ -84,18 +84,18 @@ def create_ddict(
...
             "A dragon dictionary requires at least "
-            f"{dragon_ddict.DDICT_MIN_SIZE / 1024} MB"
+            f"{dragon_ddict.DDICT_MIN_SIZE / (1024**2)} MB"
         )
 
     mem_total = num_nodes * mem_per_node
 
     logger.debug(
-        f"Creating dragon dictionary with {num_nodes} nodes, {mem_total} MB memory"
+        f"Creating dragon dictionary with {num_nodes} nodes, {mem_total} bytes memory"
```
ankona, line 93: tys. i missed these

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+        device memory.
+
+        :param request: The request that triggered the pipeline
+        :param fetch_result: Raw outputs from fetching model
+        :param device: The device on which the model must be placed
+        :returns: LoadModelResult wrapping the model loaded for the request
+        :raises ValueError: If model reference object is not found
+        :raises RuntimeError: If loading and evaluating the model failed
+        """
+        if fetch_result.model_bytes:
```
ankona, line 76: We should think about how we can avoid writing this "is it raw or is it in the featurestore" code repeatedly. 

I think my original intent was to make sure this method always got some bytes and it didn't matter to the worker where they came from (we can solve these 6 lines once in `MLWorkerCore`) but I probably saw how easy it was to use the request and forgot all my principles

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+            model_bytes = fetch_result.model_bytes
+        elif batch.raw_model and batch.raw_model.data:
+            model_bytes = batch.raw_model.data
+        else:
+            raise ValueError("Unable to load model without reference object")
+
+        providers = []
+        provider_options = []
+
+        if "gpu" in device.lower():
```
ankona, line 86: Should we pull this provider stuff into a little class `DeviceProvider` (with a better name) and maybe re-use the prototype in different workers so new workers can avoid this parsing. That also lets the worker code do something like `dp.allow_fallback = True` without worrying about the detail of the ordering in the list...

I see that this is ONNX-specific. It may not be necessary. I don't see a nice way right now to combine this list-based api w/the string based device parsing in tensorflow worker (though those two do look like they could pull device mapping into a new, shared location...

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+            onnx_deserialized = load_model_from_string(model_bytes)
+            output_tensors = [n.name for n in onnx_deserialized.graph.output]
+            input_layers = [n.name for n in onnx_deserialized.graph.input]
+            session = InferenceSession(
+                model_bytes, providers=providers, provider_options=provider_options
+            )
+        except Exception as e:
+            raise RuntimeError(
+                "Failed to load and evaluate the model: "
+                f"Model key {batch.model_id.key}, Device {device}"
```
ankona, line 111: Should we log anything passed to the `InferenceSession`?

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+                providers = ["CUDAExecutionProvider"]
+
+        # Fallback
+        providers.append("CPUExecutionProvider")
+        provider_options.append({})
+
+        try:
+            onnx_deserialized = load_model_from_string(model_bytes)
+            output_tensors = [n.name for n in onnx_deserialized.graph.output]
+            input_layers = [n.name for n in onnx_deserialized.graph.input]
```
ankona, line 104: If these two list comprehensions can't blow up, it might be nice to pull them out of this try/catch to make tracking down what's breaking easier.

If they can blow up, we might consider catching more specific errors so we can log info about input_layers/output_tensors or onnx_deserialized.graph that we wouldn't log for the InferenceSession

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+                for item_meta in fetch_result.meta:
+                    tensor_desc: tensor_capnp.TensorDescriptor = item_meta
+                    tensor_dims = list(tensor_desc.dimensions)
+                    all_dims.append([total_samples, *tensor_dims[1:]])
+                    all_dtypes.append(str(tensor_desc.dataType))
+
+        for result_tensor_idx, (dims, dtype) in enumerate(zip(all_dims, all_dtypes)):
+            itemsize = np.empty((1), dtype=dtype).itemsize
+            alloc_size = int(np.prod(dims) * itemsize)
+            mem_alloc = mem_pool.alloc(alloc_size)
```
ankona, line 172: It feels like we should move the memory pool allocation up into the superclass to avoid each worker needing to know about allocating dragon memory? Perhaps just letting them request a `memoryview` by size?

smartsim/_core/mli/infrastructure/worker/onnx_worker.py
```diff
@@ -0,0 +1,290 @@
...
+        except Exception as e:
+            raise ValueError(
+                f"Error while evaluating the model: Model {batch.model_id.key}"
+            ) from e
+
+        transform_result.transformed = []
+
+        execute_result = ExecuteResult(results, transform_result.slices)
+        for mem_alloc in mem_allocs:
+            mem_alloc.free()
```
ankona, line 251: Is there a way (or need) to defer this free to be outside the inference pipeline? I wonder if waiting for the free to complete in this serial manner hurts throughput...

smartsim/_core/mli/client/protoclient.py
```diff
@@ -200,83 +187,18 @@ def backbone_timeout(self) -> float:
...
-        value_array = numpy.transpose(value_array)
-        for i in range(value_array.shape[0]):
-            print(" ".join(self._format_number(value) for value in value_array[i]))
-        if to_file:
-            numpy.save("timings.npy", value_array)
-            numpy.savetxt("timings.txt", value_array)
-
-    def run_model(self, model: t.Union[bytes, str], batch: torch.Tensor) -> t.Any:
+    def run_model(
+        self, model: t.Union[bytes, str], batch: numpy.ndarray[t.Any, t.Any]
```
ankona, line 191: super stoked to see these logging / timing methods move out of here!

smartsim/_core/mli/infrastructure/control/worker_manager.py
```diff
@@ -108,6 +103,10 @@ def __init__(
         """Object responsible for model caching and device access"""
         self._perf_timer = PerfTimer(prefix="w_", debug=False, timing_on=True)
         """Performance timer"""
+        self._processed_batches: int = 0
```
ankona, line 106: Consider implementing a small tracking object to perform this feature instead of a couple instance vars. It should improve your ability to print those pertinent stats. Per your other comment - if you find yourself wanting more things recorded, it will keep the worker manager cleaner, too.

smartsim/_core/mli/infrastructure/control/worker_manager.py
```diff
@@ -303,18 +312,21 @@ def _on_iteration(self) -> None:
...
-                    if reply.has_outputs:
-                        # send tensor data after response
-                        for output in reply.outputs:
-                            request.callback.send(output)
+                    try:
+                        request.callback.send(serialized_resp)
+                        if reply.has_outputs:
+                            # send tensor data after response
+                            for output in reply.outputs:
+                                request.callback.send(output, timeout=None)
```
ankona, line 320: Should we create a new ticket to come back and try to avoid the serial nature of these calls? e.g. Should we be performing the callbacks in a separate thread? The `None` timeout feels like one channel could cause delays everywhere else, but even just thinking that a batch of 100 requires the 100th callback to wait for 99 to complete makes me itch.

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+                for item_meta in fetch_result.meta:
+                    tensor_desc: tensor_capnp.TensorDescriptor = item_meta
+                    tensor_dims = list(tensor_desc.dimensions)
+                    all_dims.append([total_samples, *tensor_dims[1:]])
+                    all_dtypes.append(str(tensor_desc.dataType))
+
+        for result_tensor_idx, (dims, dtype) in enumerate(zip(all_dims, all_dtypes)):
+            itemsize = np.empty((1), dtype=dtype).itemsize
+            alloc_size = int(np.prod(dims) * itemsize)
+            mem_alloc = mem_pool.alloc(alloc_size)
```
ankona, line 200: Could we sum all the dims from the loop and do a single memory allocation (i'm not sure what perf impact of those allocations is, maybe non-issue)? Do we want the workers to do this or the base class?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+            mem_view = mem_alloc.get_memview()
+            try:
+                joined = b"".join(
+                    [
+                        fetch_result.inputs[result_tensor_idx]
```
ankona, line 205: Does this list comprehension force us to load all the tensors into memory at once? would it be preferable to implement an object w/the `cloudpickle` serializer + `itertools.chain` so the bytes could stream one tensor at a time into the mem_view?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+        :param transform_result: The result of transforming inputs for model consumption
+        :param device: The device on which the model will be executed
+        :returns: The result of inference wrapped in an ExecuteResult
+        :raises SmartSimError: If model is not loaded
+        :raises IndexError: If memory slicing is out of range
+        :raises ValueError: If tensor creation fails or is unable to evaluate the model
+        """
+        if not load_result.model:
+            raise SmartSimError("Model must be loaded to execute")
+        device_to_tf = {"cpu": "/CPU", "gpu": "/GPU"}
```
ankona, line 241: This looks like the same map as above - should it be a `ClassVar` to keep it in one spot?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+        for old, new in device_to_tf.items():
+            device = device.replace(old, new)
+
+        tensors = []
+        mem_allocs = []
+        for transformed, dims, dtype in zip(
+            transform_result.transformed, transform_result.dims, transform_result.dtypes
+        ):
+            mem_alloc = MemoryAlloc.attach(transformed)
```
ankona, line 250: ditto question on if memory allocation should be done in base...

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+                        mem_alloc.get_memview()[0 : np.prod(dims) * itemsize],
+                        dtype=dtype,
+                    ).reshape(dims)
+                )
+            except IndexError as e:
+                raise IndexError("Error during memory slicing") from e
+            except Exception as e:
+                raise ValueError("Error during tensor creation") from e
+
+        sess = load_result.model
```
ankona, line 265: is `sess` short for `session`? Could we add those extra chars just to make name clear?

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
```diff
@@ -0,0 +1,322 @@
...
+                    feed_dict=dict(zip(load_result.inputs, tensors)),
+                )
+        except Exception as e:
+            raise ValueError(
+                f"Error while evaluating the model: Model {batch.model_id.key}"
+            ) from e
+
+        transform_result.transformed = []
+
+        execute_result = ExecuteResult(results, transform_result.slices)
```
ankona, line 281: bump. mem alloc

smartsim/_core/utils/timings.py
```diff
@@ -144,21 +144,21 @@ def max_length(self) -> int:
             return 0
         return max(len(value) for value in self._timings.values())
 
-    def print_timings(self, to_file: bool = False) -> None:
+    def print_timings(self, to_file: bool = False, to_stdout: bool = True) -> None:
```
ankona, line 147: nitpick alert! now that you've moved this method onto a timing object, it might be unnecessary to call this `print_timings` instead of just `print`

ankona (commented):
I've got some questions and maybe some items that become new tickets, but overall, looks great.

ex/high_throughput_inference/mock_app_redis.py
al-rigazzi: They need to store the model in different ways, but yeah, we can do that.

smartsim/_core/mli/comm/channel/channel.py
al-rigazzi: Fixed in new PR

al-rigazzi:
Closed to change base

smartsim/_core/utils/timings.py
al-rigazzi: Done

smartsim/_core/mli/infrastructure/worker/tensorflow_worker.py
al-rigazzi: To my defense, I'll say that it's a pretty common abbreviation, but I agree. Fixed.

ex/high_throughput_inference/mock_app.py
al-rigazzi: Added

ex/high_throughput_inference/mock_app.py
al-rigazzi: Done

ex/high_throughput_inference/mock_app.py
al-rigazzi: Done

ex/high_throughput_inference/mock_app.py
al-rigazzi: Good catch

ex/high_throughput_inference/mock_app.py
al-rigazzi: Good idea

ex/high_throughput_inference/mock_app.py
al-rigazzi: Yeah, the issue is that if you want to load and jit-trace a GPU model, you need a GPU, and that's something we do not want to enforce on the App.

ex/high_throughput_inference/mock_app.py
al-rigazzi: OK, exposed in `mli_driver` too, as that's what is usually run by users

ex/high_throughput_inference/mock_app_onnx.py
al-rigazzi: Ditto done

ex/high_throughput_inference/mock_app_onnx.py
al-rigazzi: Thanks

ex/high_throughput_inference/mock_app_onnx.py
al-rigazzi: Documented, will address the rest in ticket

ex/high_throughput_inference/mock_app_onnx.py
al-rigazzi: Ditto done

ex/high_throughput_inference/mock_app_onnx.py
al-rigazzi: Yeah, that's clearly not a production-ready choice. In general, the Client will not know anything about MPI, but this makes things easier.

ex/high_throughput_inference/mock_app_redis.py
al-rigazzi: Ditto documented, and will address

ex/high_throughput_inference/mock_app_redis.py
al-rigazzi: Thanks

ex/high_throughput_inference/mock_app_redis.py
al-rigazzi: Regarding this and other rightful improvements, I will open a ticket with the goal of understanding:
- whether we want to keep this App in the repo
- if so, how good we want it to be

ex/high_throughput_inference/redis_driver.py
al-rigazzi: Done

ex/high_throughput_inference/redis_driver.py
al-rigazzi: Yeah, adding this to the todo list

ex/high_throughput_inference/redis_driver.py
al-rigazzi: I now added it as a parameter, it is supposed to be the node for the workload manager (if it is the only one with a GPU)

ex/high_throughput_inference/standalone_worker_manager.py
al-rigazzi: Done

ex/high_throughput_inference/standalone_worker_manager.py
al-rigazzi: Thanks

ex/high_throughput_inference/standalone_worker_manager.py
al-rigazzi: Let's discuss this - it seems that creating it from a descriptor does not allow you to set `allow_reserver_writes`

ex/high_throughput_inference/standalone_worker_manager.py
al-rigazzi: This will be addressed as part of a future refactor

smartsim/_core/mli/client/protoclient.py
al-rigazzi: I believe it could use the log level, good idea.

ex/high_throughput_inference/mock_app_redis.py
al-rigazzi: Yeah, refer to comment about future ticket

PR #728, opened 2024-09-27
Update release action to remove CI Build Wheel

MattToast:
Removes the use of CI Build Wheel now that SmartSim is a pure python package.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,108 +32,61 @@ on:
   release:
     types: [published]
 
-
-env:
-  HOMEBREW_NO_ANALYTICS: "ON" # Make Homebrew installation a little quicker
-  HOMEBREW_NO_AUTO_UPDATE: "ON"
-  HOMEBREW_NO_BOTTLE_SOURCE_FALLBACK: "ON"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -32,108 +32,61 @@ on:
...
-        with:
-          path: ./wheelhouse/*.whl
-
-
-  build_sdist:
-    name: Build source distribution
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
+          sudo apt-get install -y build-essential wget
```
ashao, line 52: Do we actually even need this step?

ashao (approved):
Thanks for tracking this down. One suggestion to simplify even further, but we can always test this later

.github/workflows/release.yml
MattToast: Almost certainly not, but I was afraid to remove more than I had to. I'll throw up a ticket to address after release

MattToast: Ticket posted [here](https://github.com/CrayLabs/SmartSim/issues/729)

PR #732, opened 2024-09-30
Application.files refactor

amandarichardsonn:
This PR adds a operations.py file that holds that holds the functionality that allows user to add files to an Application.


smartsim/_core/commands/command.py
```diff
@@ -35,6 +35,10 @@ class Command(MutableSequence[str]):
     """Basic container for command information"""
 
     def __init__(self, command: t.List[str]) -> None:
+        if not command:
+            raise ValueError("Command list cannot be empty")
+        if not all(isinstance(item, str) for item in command):
+            raise ValueError("All items in the command list must be strings")
```
MattToast: Minor nit: can we change this to a `TypeError` since the values are of the incorrect type

smartsim/_core/generation/generator.py
```diff
@@ -240,119 +269,60 @@ def _execute_commands(cls, cmd_list: CommandList) -> None:
...
-                base_source_name = os.path.basename(src)
-                destination = os.path.join(dest, base_source_name)
-                cmd.append(str(destination))
-                cmd.append("--dirs_exist_ok")
-            else:
-                cmd.append(str(dest))
-            cmd_list.commands.append(cmd)
+        for file in files:
+            cmd_list.append(file.format(context))
         return cmd_list
```
MattToast: Style nit if you want to be a hair more declarative
```suggestion
        return CommandList([file.format(context) for file in files])
```

smartsim/_core/generation/generator.py
```diff
@@ -240,119 +269,60 @@ def _execute_commands(cls, cmd_list: CommandList) -> None:
...
-                    "smartsim._core.entrypoints.file_operations",
-                    "symlink",
-                    src,
-                    new_dest,
-                ]
-            )
-            cmd_list.append(cmd)
+        for file in files:
+            cmd_list.append(file.format(context))
         return cmd_list
```
MattToast: Ditto "style nit to be a bit more declarative" comment here
```suggestion
        return CommandList([file.format(context) for file in files])
```

smartsim/_core/generation/generator.py
```diff
@@ -240,119 +269,60 @@ def _execute_commands(cls, cmd_list: CommandList) -> None:
...
-                        path,
-                        str(dest),
-                        tag_delimiter,
-                        encoded_dict,
-                    ]
-                )
-                cmd_list.commands.append(cmd)
+        for file in files:
+            cmd_list.append(file.format(context))
         return cmd_list
```
MattToast: And here
```suggestion
        return CommandList([file.format(context) for file in files])
```

smartsim/_core/generation/generator.py
```diff
@@ -199,33 +206,55 @@ def _build_commands(
...
+        cls, cmd_list: CommandList, job_path: pathlib.Path, log_path: pathlib.Path
+    ) -> None:
+        """Append file operation Commands (mkdir) for a Job's run and log directory.
+
+        :param cmd_list: A CommandList object containing the commands to be executed
+        :param job_path: The file path for the Job run folder
+        :param log_path: The file path for the Job log folder
+        """
+        cmd_list.commands.append(cls._mkdir_file(job_path))
+        cmd_list.commands.append(cls._mkdir_file(log_path))
```
MattToast: Can we actually call the `append` method off of the `CommandList` here rather than off `commands` on the off chance that we decide to change the internal state of this class?
```suggestion
        cmd_list.append(cls._mkdir_file(job_path))
        cmd_list.append(cls._mkdir_file(log_path))
```

smartsim/_core/generation/generator.py
```diff
@@ -199,33 +207,55 @@ def _build_commands(
...
+        :param context: A GenerationContext object that holds the Job's run directory
+        """
+        copy_ret = cls._copy_files(entity.files.copy_operations, context)
+        cmd_list.commands.extend(copy_ret.commands)
+
+        symlink_ret = cls._symlink_files(entity.files.symlink_operations, context)
+        cmd_list.commands.extend(symlink_ret.commands)
+
+        configure_ret = cls._configure_files(entity.files.configure_operations, context)
+        cmd_list.commands.extend(configure_ret.commands)
```
MattToast: Similar "use command list directly on off chance we change the internal structure" question here:
```suggestion
        copy_ret = cls._copy_files(entity.files.copy_operations, context)
        cmd_list.extend(copy_ret)

        symlink_ret = cls._symlink_files(entity.files.symlink_operations, context)
        cmd_list.extend(symlink_ret)

        configure_ret = cls._configure_files(entity.files.configure_operations, context)
        cmd_list.extend(configure_ret)
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+    :return: Combined path
+    :raises ValueError: An error occurred during path combination
+    """
+    try:
+        combined_path = job_root_path
+        if dest:
+            combined_path = job_root_path / dest
+        return str(combined_path)
+    except Exception as e:
+        raise ValueError(f"Error combining paths: {e}")
```
MattToast: Just to make the stack trace a bit nicer
```suggestion
    except Exception as e:
        raise ValueError(f"Error combining paths: {e}") from e
```

MattToast: It doesn't look like any thing here can throw; is there a reason in its in a `try`/`except` block?

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+    if not isinstance(src, pathlib.Path):
+        raise TypeError(f"src must be of type pathlib.Path, not {type(src).__name__}")
+    if dest is not None and not isinstance(dest, pathlib.Path):
+        raise TypeError(
+            f"dest must be of type pathlib.Path or None, not {type(dest).__name__}"
+        )
+    if isinstance(dest, pathlib.Path) and not dest.is_absolute():
+        raise ValueError("Invalid destination path")
+    if isinstance(dest, pathlib.Path) and " " in str(dest):
+        raise ValueError("Path contains spaces, which are not allowed")
```
MattToast: Do we want to enforce this? This isn't strictly true for unix in general as I could
```sh
mkdir -p '/tmp/this is/a path/with spaces/in/it'
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # TODO I want to add the check below but I do not think this works for remote jobs
+    # full_path = path.abspath(file_path)
+    #     if path.isfile(full_path):
+    #         return full_path
+    #     if path.isdir(full_path):
+    #         return full_path
```
MattToast: Can confirm that this would not work a remote job :(

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+        self, src: pathlib.Path, dest: t.Optional[pathlib.Path] = None
+    ) -> None:
+        """Initialize a CopyOperation object
+
+        :param src: Path to source
+        :param dest: Path to destination
+        """
+        check_src_and_dest_path(src, dest)
+        self.src = src
+        self.dest = dest
```
MattToast: Can we resolve `dest` to some relative path right away here so that we don't have to deal with the nullish possiblitly? Like could we do something like
```suggestion
        self.src = src
        self.dest = dest or src.name
```
?

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+        self, src: pathlib.Path, dest: t.Optional[pathlib.Path] = None
+    ) -> None:
+        """Initialize a SymlinkOperation object
+
+        :param src: Path to source
+        :param dest: Path to destination
+        """
+        check_src_and_dest_path(src, dest)
+        self.src = src
+        self.dest = dest
```
MattToast: Similar "can we avoid the nullish possibility" question here:
```suggestion
        check_src_and_dest_path(src, dest)
        self.src = src
        self.dest = dest or src.name
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+        """Initialize a ConfigureOperation
+
+        :param src: Path to source
+        :param file_parameters: File parameters to find and replace
+        :param dest: Path to destination
+        :param tag: Tag to use for find and replacement
+        """
+        check_src_and_dest_path(src, dest)
+        self.src = src
+        self.dest = dest
```
MattToast: And one here!
```suggestion
        self.src = src
        self.dest = dest or src.name
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+        if dest:
+            combined_path = job_root_path / dest
+        return str(combined_path)
+    except Exception as e:
+        raise ValueError(f"Error combining paths: {e}")
+
+
+def check_src_and_dest_path(
+    src: pathlib.Path, dest: t.Union[pathlib.Path, None]
+) -> None:
```
MattToast: Style nit: If this function is never meant to be used outside of this module (from what I can tell, it is not), can we mark it as such?
```suggestion
def _check_src_and_dest_path(
    src: pathlib.Path, dest: t.Union[pathlib.Path, None]
) -> None:
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+"""Copy file operations command"""
+symlink_cmd = "symlink"
+"""Symlink file operations command"""
+configure_cmd = "configure"
+"""Configure file operations command"""
+
+
+def create_final_dest(
+    job_root_path: pathlib.Path, dest: t.Union[pathlib.Path, None]
+) -> str:
```
MattToast: Similar "can we mark this protected" question here
```suggestion
def _create_final_dest(
    job_root_path: pathlib.Path, dest: t.Union[pathlib.Path, None]
) -> str:
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # TODO I want to add the check below but I do not think this works for remote jobs
+    # full_path = path.abspath(file_path)
+    #     if path.isfile(full_path):
+    #         return full_path
+    #     if path.isdir(full_path):
+    #         return full_path
+
+
+def check_run_path(run_path: pathlib.Path) -> None:
```
MattToast: Similar "can we mark this protected" question here
```suggestion
def _check_run_path(run_path: pathlib.Path) -> None:
```

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+    :param run_path: The run path to be checked
+    :raises TypeError: If either run path is not an instance of pathlib.Path
+    :raises ValueError: If the run path is not a directory
+    """
+    if not isinstance(run_path, pathlib.Path):
+        raise TypeError(
+            f"run_path must be of type pathlib.Path, not {type(run_path).__name__}"
+        )
+    if run_path == pathlib.Path(""):
+        raise ValueError(f"Job root path '{run_path}' is not a directory.")
```
MattToast: Little confused by this check. The error message and the check seem to be slightly different. Should this read
 ```suggestion
    if run_path == pathlib.Path(""):
        raise ValueError(f"Job root path '{run_path}' cannot be the current working directory.")
```
?

Comment (MattToast) [reply smartsim/_core/generation/operations.py]: Note that `pathlib.Path("") == pathlib.Path(".")`

File: smartsim/_core/generation/operations.py [hunk 17]
```diff
@@ -0,0 +1,298 @@
...
+                configure_cmd,
+                str(self.src),
+                final_dest,
+                self.tag,
+                self.file_parameters,
+            ]
+        )
+
+
+T = t.TypeVar("T", bound=GenerationProtocol)
```
MattToast: Nit: Usually if I am making a type var, and I am binding it to something, I like to try and give the variable a slightly more descriptive name just so that it is clear what it is bound to. Thoughts on renaming to
```suggestion
GenerationProtocolT = t.TypeVar("GenerationProtocolT", bound=GenerationProtocol)

```
?

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+
+entry_point_path = "smartsim._core.entrypoints.file_operations"
+"""Path to file operations module"""
+
+copy_cmd = "copy"
+"""Copy file operations command"""
+symlink_cmd = "symlink"
+"""Symlink file operations command"""
+configure_cmd = "configure"
+"""Configure file operations command"""
```
MattToast: Consider making top level constants `SCREAMING_SNAKE_CASE` or opting out of the lint check for when we turn pylint back on with a `# pylint: disable=invalid-name`

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @property
+    def configure_operations(self) -> t.List[ConfigureOperation]:
+        """Property to get the list of configure files.
+
+        :return: List of ConfigureOperation objects
+        """
+        return self._filter(ConfigureOperation)
+
+    def _filter(self, type: t.Type[T]) -> t.List[T]:
```
MattToast: Nit that applies throughout this module. If you want, `t.List`, `t.Type`, and other builtins have been [deprecated in 3.9](https://docs.python.org/3/library/typing.html#aliases-to-built-in-types), feel free to just use `type`, `list`, etc. directly!

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+T = t.TypeVar("T", bound=GenerationProtocol)
+
+
+@dataclass
+class FileSysOperationSet:
+    """Dataclass to represent a set of file system operation objects"""
+
+    # TODO disallow modification - dunder function (post ticket to reevaluate API objects)
+    operations: t.List[GenerationProtocol] = field(default_factory=list)
+    """Set of file system objects that match the GenerationProtocol"""
```
MattToast: Is there a ticket for this? I don't recall seeing one in the backlog

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+        return self._filter(ConfigureOperation)
+
+    def _filter(self, type: t.Type[T]) -> t.List[T]:
+        """Filters the operations list to include only instances of the
+        specified type.
+
+        :param type: The type of operations to filter
+        :return: A list of operations that are instances of the specified type
+        """
+        return [x for x in self.operations if isinstance(x, type)]
```
MattToast: Just a heads up, pylint should complain about us "redefining builtin symbol `type`" if/when we turn that back on. Consider adding a trailing underscore or something to the symbol to avoid.
```suggestion
    def _filter(self, type_: t.Type[T]) -> t.List[T]:
        """Filters the operations list to include only instances of the
        specified type.

        :param type_: The type of operations to filter
        :return: A list of operations that are instances of the specified type
        """
        return [x for x in self.operations if isinstance(x, type_)]
```

smartsim/entity/application.py
```diff
@@ -58,9 +59,11 @@ def __init__(
         self,
         name: str,
         exe: str,
+        file_parameters: (
+            t.Mapping[str, str] | None
+        ) = None,  # TODO remove when Ensemble is addressed
+        files: t.Optional[EntityFiles] = None,  # TODO remove when Ensemble is addressed
         exe_args: t.Optional[t.Union[str, t.Sequence[str]]] = None,
-        files: t.Optional[EntityFiles] = None,
-        file_parameters: t.Mapping[str, str] | None = None,
```
MattToast: Small nit: I know these parameters are almost certainly going to go away in future, but can we keep the original order of the signature. I want to

1) keep `exe_args` immediately following `exe` and
1) this is purely selfish, but I don't want add this change as it will break some of the scripts that I have been running to test SmartSim by hand lol

smartsim/entity/application.py
```diff
@@ -58,9 +59,11 @@ def __init__(
         self,
         name: str,
         exe: str,
+        file_parameters: (
+            t.Mapping[str, str] | None
+        ) = None,  # TODO remove when Ensemble is addressed
+        files: t.Optional[EntityFiles] = None,  # TODO remove when Ensemble is addressed
```
MattToast: It doesn't look like this parameter is used. Feel free to remove now!

tests/test_generator.py
```diff
@@ -255,55 +243,110 @@ def test_mkdir_file(generator_instance: Generator, test_dir: str):
...
+        ),
+    ),
+)
+@pytest.mark.parametrize("source", ["files", "directory"], indirect=True)
+def test_symlink_files_valid_dest(
+    dest, source, generator_instance: Generator, test_dir: str
+):
+    to_copy = [CopyOperation(src=file, dest=dest) for file in source]
+    gen = GenerationContext(pathlib.Path(test_dir))
+    cmd_list = generator_instance._copy_files(files=to_copy, context=gen)
```
MattToast:
```suggestion
def test_symlink_files_valid_dest(
    dest, source, generator_instance: Generator, test_dir: str
):
    to_copy = [CopyOperation(src=file, dest=dest) for file in source]
    gen = GenerationContext(pathlib.Path(test_dir))
    cmd_list = generator_instance._symlink_files(files=to_copy, context=gen)
```
?

tests/test_generator.py
```diff
@@ -255,55 +243,110 @@ def test_mkdir_file(generator_instance: Generator, test_dir: str):
...
-    )
-    tagged_files = sorted(glob(conf_path + "/*"))
-    files = EntityFiles(tagged=tagged_files)
-    param_set = {
+    # Extract file paths from commands
+    cmd_src_paths = set()
+    for cmd in cmd_list.commands:
+        # Assert destination path exists in cmd
+        assert create_final_dest(pathlib.Path(test_dir), dest) in cmd.command
+        src_index = cmd.command.index("copy") + 1
```
MattToast:
```suggestion
        src_index = cmd.command.index("symlink") + 1
```
?

tests/test_generator.py
```diff
@@ -255,55 +243,110 @@ def test_mkdir_file(generator_instance: Generator, test_dir: str):
...
+        cmd_src_paths.add(cmd.command[src_index])
+    # Assert all file paths are in the command list
+    file_paths = {str(file) for file in source}
+    assert file_paths.issubset(
+        cmd_src_paths
+    ), "Not all file paths are in the command list"
```
MattToast: Should these sets be exactly the same?
```suggestion
    assert file_paths == cmd_src_paths, "Not all file paths are in the command list"
```

tests/test_generator.py
```diff
@@ -255,55 +243,110 @@ def test_mkdir_file(generator_instance: Generator, test_dir: str):
...
```
MattToast: Similar question here
```suggestion
    assert file_paths == cmd_src_paths, "Not all file paths are in the command list"
```

tests/test_generator.py
```diff
@@ -255,55 +243,110 @@ def test_mkdir_file(generator_instance: Generator, test_dir: str):
...
-    """Test Generator._symlink_files helper function with file list"""
-    symlink_files = sorted(glob(get_gen_symlink_dir + "/*"))
-    files = EntityFiles(symlink=symlink_files)
-    cmd_list = generator_instance._symlink_files(files, generator_instance.root)
+@pytest.fixture
+def source(request, files, directory):
+    if request.param == "files":
+        return files
+    elif request.param == "directory":
+        return directory
```
MattToast: rather than indirectly setting parameters for this fixture for every test, consider simply parametrizing the fixture itself
```suggestion
@pytest.fixture(params=["files", "directory"])
def source(request):
    yield request.getfixturevalue(request.param)
```

tests/test_generator.py
```diff
@@ -312,159 +355,160 @@ def test_write_tagged_file(fileutils, generator_instance: Generator):
...
+    for cmd in cmd_list.commands:
+        # Assert destination path exists in cmd
+        assert create_final_dest(pathlib.Path(test_dir), dest) in cmd.command
+        src_index = cmd.command.index("configure") + 1
+        cmd_src_paths.add(cmd.command[src_index])
+    # Assert all file paths are in the command list
+    file_paths = {str(file) for file in source}
+    assert file_paths.issubset(
+        cmd_src_paths
+    ), "Not all file paths are in the command list"
```
MattToast:
```suggestion
    assert file_paths == cmd_src_paths, "Not all file paths are in the command list"
```
?

tests/test_generator.py
```diff
@@ -312,159 +355,160 @@ def test_write_tagged_file(fileutils, generator_instance: Generator):
...
 
-def test_write_tagged_directory(fileutils, generator_instance: Generator):
-    """Test Generator._write_tagged_files helper function with directory path"""
-    config = get_gen_file(fileutils, "tag_dir_template")
-    files = EntityFiles(tagged=[config])
-    param_set = {"PARAM0": "param_value_1", "PARAM1": "param_value_2"}
-    cmd_list = generator_instance._write_tagged_files(
-        files=files, params=param_set, dest=generator_instance.root
-    )
+# TODO Add configure_file tests
```
MattToast: Looks like the previous test was a configure file test. Is this for making a test that actually configures and dumps a file to disk?

tests/test_generator.py
```diff
@@ -312,159 +355,160 @@ def test_write_tagged_file(fileutils, generator_instance: Generator):
...
-    _check_generated(0, 2, jobs_dir / "ensemble-name-0-0" / Generator.run_directory)
-    ids.clear()
+    """Test Generator._execute_commands calls with appropriate type and num times"""
+    with (
+        unittest.mock.patch(
+            "smartsim._core.generation.generator.subprocess.run"
+        ) as mock_run,
+    ):
+        generator_instance._execute_commands(formatted_command_list)
+        assert mock_run.call_count == len(operations_list)
```
MattToast: Should this be the length of the `formatted_command_list`?
```suggestion
        generator_instance._execute_commands(formatted_command_list)
        assert mock_run.call_count == len(formatted_command_list)
```

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    GenerationContext,
+    SymlinkOperation,
+    configure_cmd,
+    copy_cmd,
+    create_final_dest,
+    symlink_cmd,
+)
+
+# QUESTIONS
+# TODO test python protocol?
```
MattToast: I would argue that's unneeded. In theory protocols are only of type checking, it would be equivalent to trying to write a unit test for a java style interface.

smartsim/_core/generation/operations.py
```diff
@@ -0,0 +1,298 @@
...
+    :raises TypeError: If either src or dest is not an instance of pathlib.Path
+    """
+    if not isinstance(src, pathlib.Path):
+        raise TypeError(f"src must be of type pathlib.Path, not {type(src).__name__}")
+    if dest is not None and not isinstance(dest, pathlib.Path):
+        raise TypeError(
+            f"dest must be of type pathlib.Path or None, not {type(dest).__name__}"
+        )
+    if isinstance(dest, pathlib.Path) and not dest.is_absolute():
+        raise ValueError("Invalid destination path")
```
MattToast: Is this right? I thought the destination path was supposed to always be relative to the (in theory, at this point non-existent) run directory.

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+# TODO do I allow the paths to combine if src is empty?
```
MattToast: Can `src` ever be "empty"? The type hints suggest that it must always be a `pathlib.Path` (e.g. not `None`).

Do you mean if a `src` is pointing to an empty directory?

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+@pytest.fixture
+def copy_operation(mock_src: pathlib.Path, mock_dest: pathlib.Path):
+    """Fixture to create a CopyOperation object."""
+    return CopyOperation(src=mock_src, dest=mock_dest)
+
+
+@pytest.fixture
+def symlink_operation(mock_src: pathlib.Path, mock_dest: pathlib.Path):
+    """Fixture to create a SymlinkOperation object."""
+    return SymlinkOperation(src=mock_src, dest=mock_dest)
```
MattToast: These fixtures look very similar to the ones in `test_generator.py`. It might be worth re-organizing the files so that we can put these common fixtures into a shared `conftest.py`
```
<SmartSim root>
  ├── conftest.py
  └── tests
      └── generation
          ├── conftest.py  # <-- Shared fixtures go here
          ├── test_generator.py
          └── test_operations.py
```

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    (
+        pytest.param(None, pathlib.Path("valid/dest"), id="None as root path"),
+        pytest.param(1234, pathlib.Path("valid/dest"), id="Number as root path"),
+        pytest.param(pathlib.Path("valid/dest"), 1234, id="Number as dest"),
+    ),
+)
+def test_create_final_dest_invalid(job_root_path, dest):
+    """Test invalid path inputs for operations.create_final_dest"""
+    with pytest.raises(ValueError):
+        create_final_dest(job_root_path, dest)
```
MattToast: This feedback may be moot if you accept some of my previous comments, but if we do keep this test, could we actually get this to raise a `TypeError` since the problem here is due to trying to combine arguments of incompatible type, not value?

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def test_valid_init_generation_context(
+    test_dir: str, generation_context: GenerationContext
+):
+    """Validate GenerationContext init"""
+    assert isinstance(generation_context, GenerationContext)
+    assert generation_context.job_root_path == pathlib.Path(test_dir)
```
MattToast: If this is testing the initializer, can you actually initialize a new `GenerationContext` within this test, instead of using the pre-initialized one from a fixture? That way this test is still valid, even if we change the scope of the fixture

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+        GenerationContext("")
+
+
+def test_init_copy_operation(
+    copy_operation: CopyOperation, mock_src: pathlib.Path, mock_dest: pathlib.Path
+):
+    """Validate CopyOperation init"""
+    assert isinstance(copy_operation, CopyOperation)
+    assert copy_operation.src == mock_src
+    assert copy_operation.dest == mock_dest
```
MattToast: ditto here

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    assert create_final_dest(mock_src, mock_dest) in exec.command
+
+
+def test_init_symlink_operation(
+    symlink_operation: SymlinkOperation, mock_src: str, mock_dest: str
+):
+    """Validate SymlinkOperation init"""
+    assert isinstance(symlink_operation, SymlinkOperation)
+    assert symlink_operation.src == mock_src
+    assert symlink_operation.dest == mock_dest
```
MattToast: and here

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    configure_operation: ConfigureOperation, mock_src: str, mock_dest: str
+):
+    """Validate ConfigureOperation init"""
+    assert isinstance(configure_operation, ConfigureOperation)
+    assert configure_operation.src == mock_src
+    assert configure_operation.dest == mock_dest
+    assert configure_operation.tag == ";"
+    decoded_dict = base64.b64decode(configure_operation.file_parameters.encode("ascii"))
+    unpickled_dict = pickle.loads(decoded_dict)
+    assert unpickled_dict == {"FOO": "BAR"}
```
MattToast: and one last one here

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    generation_context: GenerationContext,
+):
+    """Validate SymlinkOperation.format"""
+    exec = symlink_operation.format(generation_context)
+    assert isinstance(exec, Command)
+    assert str(mock_src) in exec.command
+    assert symlink_cmd in exec.command
+
+    normalized_path = os.path.normpath(mock_src)
+    parent_dir = os.path.basename(normalized_path)
```
MattToast: Did you mean `dirname` here? `parent_dir` here will be the file name in the case where `mock_src` is a file
```py
>>> import os.path as osp
>>> osp.basename("a/b/c.txt")
'c.txt'
>>> osp.dirname("a/b/c.txt")
'a/b'
```

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    assert isinstance(exec, Command)
+    assert str(mock_src) in exec.command
+    assert configure_cmd in exec.command
+    assert create_final_dest(mock_src, mock_dest) in exec.command
+
+
+def test_init_file_sys_operation_set(file_system_operation_set: FileSysOperationSet):
+    """Test initialize FileSystemOperationSet"""
+    assert isinstance(file_system_operation_set.operations, list)
+    assert len(file_system_operation_set.operations) == 3
```
MattToast: Ditto "if you are testing the initialize, can you call it in the test?" request here!

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def test_add_copy_operation(
+    file_system_operation_set: FileSysOperationSet, copy_operation: CopyOperation
+):
+    """Test FileSystemOperationSet.add_copy"""
+    assert len(file_system_operation_set.copy_operations) == 1
+    file_system_operation_set.add_copy(src=pathlib.Path("src"))
```
MattToast: Did we want to allow for relative path source files? I thought source file paths had to be absolute.

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    assert len(file_system_operation_set.copy_operations) == 2
+
+
+def test_add_symlink_operation(
+    file_system_operation_set: FileSysOperationSet, symlink_operation: SymlinkOperation
+):
+    """Test FileSystemOperationSet.add_symlink"""
+    assert len(file_system_operation_set.symlink_operations) == 1
+    file_system_operation_set.add_symlink(src=pathlib.Path("src"))
+    assert len(file_system_operation_set.symlink_operations) == 2
```
MattToast: I'll leave to your own discretion, but it might be a good idea to nut check the number of operations at the begining in case someone updates the fixture down the line. That way this test will continue to pass
```suggestion
    """Test FileSystemOperationSet.add_symlink"""
   orig_num_ops =  len(file_system_operation_set.symlink_operations)
    file_system_operation_set.add_symlink(src=pathlib.Path("src"))
    assert len(file_system_operation_set.symlink_operations) == orig_num_ops + 1
```

MattToast: A similar comment applies for copy and configure.

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+    )
+    yield [pathlib.Path(directory)]
+
+
+@pytest.fixture
+def source(request, files, directory):
+    if request.param == "files":
+        return files
+    elif request.param == "directory":
+        return directory
```
MattToast: Ditto here, these fixtures look very familiar. Consider moving them to a shared namespace

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+        pytest.param(123, TypeError, id="dest as integer"),
+        pytest.param("", TypeError, id="dest as empty str"),
+        pytest.param("/absolute/path", TypeError, id="dest as absolute str"),
+        pytest.param(
+            pathlib.Path("relative/path"), ValueError, id="dest as relative Path"
+        ),
+        pytest.param(
+            pathlib.Path("/path with spaces"), ValueError, id="dest as Path with spaces"
+        ),
+        # TODO pytest.param(pathlib.Path("/path/with/special!@#"), id="dest as Path with special char"),
```
MattToast: Do we want this test? There is nothing stopping us unix-wise from making files with special characters
```sh
$ mkdir -p '/tmp/path/with/special!@#'
$ touch '/tmp/path/with/special!@#/!@#$%^&*()_+=.txt'
$ ls /tmp/path/with/special\!@\#
!@#$%^&*()_+=.txt
```

MattToast: Similar comment applies for symlink and configure

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+
+@pytest.mark.parametrize(
+    "dest,error",
+    (
+        pytest.param(123, TypeError, id="dest as integer"),
+        pytest.param("", TypeError, id="dest as empty str"),
+        pytest.param("/absolute/path", TypeError, id="dest as absolute str"),
+        pytest.param(
+            pathlib.Path("relative/path"), ValueError, id="dest as relative Path"
+        ),
```
MattToast: Wait, does this imply that a destination to copy to can neither be an absolute nor relative path?

tests/test_operations.py
```diff
@@ -0,0 +1,342 @@
...
+        pytest.param(
+            pathlib.Path("/path with spaces"), ValueError, id="dest as Path with spaces"
+        ),
```
MattToast: Similar "do we want to guard against this?" question here. There is nothing in *nix that is stopping us from doing this

MattToast (changes requested):
Looks really good! Just dropping some initial feedback so that we can start ironing out some of the finer points, but otherwise this looking really good!

smartsim/_core/commands/command.py
amandarichardsonn: So should the raise in setitem dunder also be TypeError? ValueError is when Im expecting 5 but get 1?

smartsim/_core/generation/generator.py
amandarichardsonn: heck yeah!

smartsim/_core/generation/generator.py
amandarichardsonn: oh yes might as well use the mutable sequence I guess MERP

smartsim/_core/generation/operations.py
amandarichardsonn: I thought line 36 could throw

smartsim/_core/generation/operations.py
amandarichardsonn: ah I did not know

smartsim/_core/generation/operations.py
amandarichardsonn: Ah yes that is super mega reasonable

smartsim/_core/generation/operations.py
amandarichardsonn: ah nice catch

smartsim/entity/application.py
amandarichardsonn: so valid I got u

tests/test_generator.py
amandarichardsonn: this never happened...

tests/test_generator.py
amandarichardsonn: gah this never happened!!

tests/test_generator.py
amandarichardsonn: yes but also should be same length tho

tests/test_operations.py
amandarichardsonn: not me spending time attempting, okay will move on

tests/test_operations.py
amandarichardsonn: ah I see

tests/test_operations.py
amandarichardsonn: Im not sure what I was doing here???

tests/test_generator.py
amandarichardsonn: yes, we don't manipulate the src paths, I am comparing that the list of src paths passed in the beginning match the source paths in the commandlist

tests/test_generator.py
amandarichardsonn: So I was thinking that since test_file_operations.py tests this, that I did not need to in this file but please let me know otherwise! I was thinking that testing _execute_commands calls the correct entry points was sufficient

tests/test_operations.py
amandarichardsonn: I was wondering if I need to check if src is None! I think people should be allowed to copy an empty dir right?

smartsim/_core/generation/operations.py
amandarichardsonn: ah hmm I did not realize `pathlib.Path("") == pathlib.Path(".")` -> I feel like I should allow someone to set to cwd if they would like, thoughts?

smartsim/_core/generation/operations.py
amandarichardsonn: noted to write

tests/test_operations.py
amandarichardsonn: thats a much better idea!

smartsim/_core/commands/command.py
MattToast: If you don't mind updating those as well, I would!

> ValueError is when Im expecting 5 but get 1?

It's all kinda up to personal interpenetration, but yeah, I like to raise a [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError) when I want a `str` but I get an `int`, where as I raise a [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError) when I want a positive `int`, but get `-1`

smartsim/_core/generation/operations.py
MattToast: Looks like it should be safe, if we assume that both `job_root_path` and `dest` are `pathlib.Path` objects (e.g. if we trust our typehints)!

[Here](https://github.com/python/cpython/blob/3.13/Lib/pathlib/_local.py#L148-L158) is the impl for `__truediv__` and `__rtruedive__` (which is how you override the `/` operator). And it doesn't look like [either can throw](https://github.com/python/cpython/blob/3.13/Lib/pathlib/_abc.py#L130-L135) for a `PosixPath` unless someone specifically overrides the method. And in that case, I think its probably a good idea to let the error raise!

tests/test_operations.py
MattToast: > I was wondering if I need to check if src is None!

Ah! Okay good! If you want to add a test to to test that an error is raised if a nullish path is passed in, I wouldn't stop you, but iirc the operations were supposed to be pretty removed from the public API, so if you'd rather have this enforced by our type checker, that is also fine with me!!

> I think people should be allowed to copy an empty dir right?

Yeah, I agree, I cannot think of a good reason to _not_ allow users to do it, and I could see it being super useful as a way to aggregate a bunch of output data amongst ensemble members. I was just confused on the meaning of the word "empty", lol!

tests/test_generator.py
MattToast: ^^ Totally makes sense to me! In that case can this TODO be removed entirely?

tests/test_generator.py
MattToast: Gotcha! I was just a little confused on if the `issubset` usage was an intentional choice and we expect potentially _more_ items to appear in `cmd_src_paths`. Looks like these tests have been updated to use `==` instead!

Marking as resolved!

smartsim/_core/generation/operations.py
MattToast: > feel like I should allow someone to set to cwd if they would like

Yeah, I can't think of a reason to _not_ let someone generate to the cwd. In that case we could change this check to
```suggestion
    if run_path.is_dir():
        raise ValueError(f"Job root path '{run_path}' is not a directory.")
```
but that could potentially run into problems for "remote jobs" (if/when they exist, lol).

The other last solution I can think of is to just drop this check altogether for now and revisit when we do start working on remote generation.

smartsim/_core/generation/operations.py
amandarichardsonn: Ok thank u!

tests/test_operations.py
amandarichardsonn: I will be more clear next time!

smartsim/entity/files.py
```diff
@@ -29,6 +29,7 @@
 from tabulate import tabulate
 
 
+# TODO remove when Ensemble is addressed
```
mellis13, line 32: Is this TODO still relevant?

smartsim/_core/generation/operations/operations.py
```diff
@@ -0,0 +1,276 @@
...
+import pickle
+import sys
+import typing as t
+from dataclasses import dataclass, field
+
+from ...commands import Command
+from .utils.helpers import check_src_and_dest_path
+
+# pylint: disable=invalid-name
+entry_point_path = "smartsim._core.entrypoints.file_operations"
```
MattToast: Minor nit: can we do this on a per-name basis instead of disabling the rule for the entire file?
```suggestion
# pylint: disable-next=invalid-name
entry_point_path = "smartsim._core.entrypoints.file_operations"
```

MattToast (approved):
One last small request, but otherwise LGTM!!

smartsim/_core/generation/operations/operations.py
amandarichardsonn: oop GAH yes!

PR #738, opened 2024-10-08
Add patch to Tensorflow backend for GCC 14 compatibility

ashao:
In libtensorflow, the `input` argument to `TF_SessionRun` seems to be mistyped to `TF_Output` instead of `TF_Input`. These two types differ only in name. GCC-14 catches this and throws an error, even though earlier versions allow this. To solve this problem, patches are applied to the Tensorflow backend in RedisAI. Future versions of Tensorflow may fix this problem, but for now this seems to be the best workaround.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -9,6 +9,24 @@ Jump to:
 
 ## SmartSim
 
+### develop
+
+To be released at some point in the future
+
+Description
+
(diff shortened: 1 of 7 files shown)


MattToast (approved):
Builds for me with GCC 14! LGTM!!

PR #741, opened 2024-10-10
Updated Slurm Launcher

MattToast:
Implement a Slurm Launcher for fine grain status tracking of Slurm jobs (so far just `srun`, but can be expanded to include `sbatch`) with the V1 API.


tests/test_slurm_launcher.py
```diff
@@ -0,0 +1,340 @@
...
+    launcher = SlurmLauncher()
+    srun = make_srun_command(
+        ["-N", "1", "-n", "1"], ["sleep", "120"], use_current_alloc=True
+    )
+    id_ = launcher.start(srun)
+    time.sleep(1)
+    assert launcher.get_status(id_)[id_] == JobStatus.RUNNING
+    launcher.stop_jobs(id_)
+    time.sleep(1)
+    assert launcher.get_status(id_)[id_] == JobStatus.CANCELLED
```
MattToast, line 342: Question for reviewers: Is this something we want to be able to handle? SmartSim V0.8 also fails this test. It's impl will set the status to `JobStatus.FAILED` and this impl currently sets the status to `JobStatus.UNKOWN`.

If so, should this be handled in this ticket or a follow on?

MattToast (commented):
One small question for reviewers in a departure from the original impl, and neither it nor this new one seem "correct" in my opinion.

pyproject.toml
```diff
@@ -154,7 +154,7 @@ module = [
   "smartsim._core.launcher.lsf.*",
   "smartsim._core.launcher.pbs.*",
   "smartsim._core.launcher.sge.*",
-  "smartsim._core.launcher.slurm.*",
+  "smartsim._core.launcher.slurm.slurmLauncher",
```
al-rigazzi: May I ask why, just out of my ignorance?

smartsim/_core/launcher_/slurm/slurm_parser.py
```diff
@@ -121,7 +121,10 @@ def parse_sstat_nodes(output: str, job_id: str) -> t.List[str]:
     return list(set(nodes))
 
 
-def parse_step_id_from_sacct(output: str, step_name: str) -> t.Optional[str]:
+StepID = t.NewType("StepID", str)
```
al-rigazzi, line 124: Are there plans on making this something more structured?

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+        :param interval: The amount of time in seconds to wait between `sacct`
+            calls to get the step id.
+        :returns: An id to reference the process for status tracking.
+        """
+        launchable.start()
+        trials = CONFIG.wlm_trials
+        step_id = None
+        while step_id is None and trials > 0:
+            # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+            # TODO: Really don't like the implied sacct format and required
```
al-rigazzi, line 162: Do we have a ticket for this?

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+        """Given a collection of step ids, interogate slurm for the status of
+        the steps.
+
+        :param id_: The first step id to get the status for.
+        :param ids: Any additional step ids to get the status for.
+        :returns: A mapping of step ids to statuses
+        """
+        ids = (id_,) + ids
+        # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+        # TODO: Really don't like the implied sacct format and required
```
al-rigazzi, line 234: Do we have a ticket for this?

al-rigazzi (approved):
Just a few comments, otherwise looks great! LGTM

pyproject.toml
MattToast: There use to be some files in that `smartsim._core.launcher.slurm` module that I did want to add to the type check, but I ended up just moving them to `smartsim._core.launcher_.slurm`. This change can (and actually probably should) be reverted!

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: Not yet, but I will write one!

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: Adding to my list of tickets to write!

smartsim/_core/launcher_/slurm/slurm_parser.py
MattToast: I would _LOVE_ to, but seeing as we are kinda shifting focus away from using launchers directly and more toward dragon, I wasn't planning on tackling it immediately. The `NewType` here is just so that the type checker could help me keep my sanity, lol.

I could definitely wrap the two other tickets requested into a "revamp slurm utilities" style ticket for when we have more time to work on this!!

tests/test_slurm_launcher.py
ankona: Do we care about the distinction between unknown and cancelled? If not, ignore it.  

If yes, new ticket. I wouldn't add more to your existing ticket since it's technically a breaking change and non-critical.

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+
+if t.TYPE_CHECKING:
+    from smartsim.experiment import Experiment
+
+
+logger = get_logger(__name__)
+
+
+class SrunCommand:
+    """A type to inject job important information such as the job name and ID
```
ankona: Consider revising for verbosity and awkward wording...
`Produces a properly formatted srun command from raw job information.`

The docstring as-is appears to say that SrunCommand is all about tracking but that appears to be not totally true. Is there another SrunCommand that _doesn't_ allow tracking? What about:

`Produces a properly formatted srun command from raw job information. Ensures the job status can be tracked using the supplied name.`

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+            raise errors.LauncherJobNotFound(msg)
+        return info
+
+    def get_status(self, *ids: LaunchedJobID) -> t.Mapping[LaunchedJobID, JobStatus]:
+        """Take a collection of job ids and return the status of the
+        corresponding slrum processes started by the slurm launcher.
+
+        :param ids: A collection of ids of the launched jobs to get the
+            statuses of.
+        :returns: A mapping of ids for jobs to stop to their reported status.
```
ankona: I have no idea what "...to stop to their reported status" means.. copypasta error from a method that cancels jobs, maybe?

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+        # a '+' in the step id, so that the format becomes 12345+1.0.  If we
+        # find it it can mean two things: a MPMD srun command, or a
+        # heterogeneous job.  If it is a MPMD srun, then stop parent step
+        # because sub-steps cannot be stopped singularly.
+        is_sub_step = "+" in step_id
+        is_het_job = os.getenv("SLURM_HET_SIZE") is not None
+        # If it is a heterogeneous job, we can stop them like this. Slurm will
+        # throw an error, but will actually kill steps correctly.
+        if is_sub_step and not is_het_job:
+            step_id_, *_ = step_id.split("+", maxsplit=1)
```
ankona: looks like we're doing parsing outside the parser. consider that the split operation could also be handled in the parser to avoid having tightly coupled the parser to the stop_job implementation. ditto for line 298 above

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+        """Take a collection of job ids and kill the corresponding processes
+        started by the slurm launcher.
+
+        :param ids: The ids of the launched jobs to stop.
+        :returns: A mapping of ids to their reported status after attempting to
+            stop them.
+        """
+        slurm_infos = tuple(map(self._get_slurm_info_from_job_id, ids))
+        for info in slurm_infos:
+            self._stop_job(info)
```
ankona: Should we do these asynchronously to avoid waiting for each job to stop sequentially?

smartsim/_core/launcher_/slurm/slurm_launcher.py
```diff
@@ -0,0 +1,334 @@
...
+            if is_het_job:
+                msg = (
+                    "SmartSim received a non-zero exit code while canceling"
+                    f" a heterogeneous job step {job_info.name}!\n"
+                    "The following error might be internal to Slurm\n"
+                    "and the heterogeneous job step could have been correctly"
+                    " canceled.\n"
+                    "SmartSim will consider it canceled.\n"
+                )
+                job_info.status_override = JobStatus.CANCELLED
```
ankona, line 312: I wonder if this could be a recursive structure (like an exception)? Does this keep the API fixed (e.g. you always ask for `job_info.status` but have the root cause info if you really need it. At first glance, having status_override be public feels like it forces a client to look at both.

tests/test_slurm_launcher.py
MattToast: Sounds like a plan! I think we care about the difference (or at least we cared enough to put it in the original API), so I'll add a note to my "tickets to write for slurm launcher" list!

smartsim/settings/arguments/launch/slurm.py
```diff
@@ -44,27 +44,45 @@
...
-        "srun",
+) -> SrunCommand:
+    # TODO: Should probably use the `SmartSimEntity` name here, but right now
+    #       there is no way to access it. In the meantime, we will just use the
+    #       exe being called.
+    app, *_ = exe
+    try:
+        _, job_name = app.rsplit(os.path.sep, maxsplit=1)
+    except ValueError:
+        job_name = app
```
MattToast: `ValueError` seems broad, consider
```suggestion
    *_, job_name = app.rsplit(os.path.sep, maxsplit=1)
```

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: copypasta error 🤦. Replaced with wording matching the launcher base class

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: Love it!! Stealing that verbatim!

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: Moved logic to the `slurm_parser` module!

smartsim/_core/launcher_/slurm/slurm_launcher.py
MattToast: Makes sense to me! Now running each `scancel` in a new thread!

PR #742, opened 2024-10-11
Rewrite RequestBatch

AlyssaCote:
In this PR I rewrote RequestBatch so that it doesn't hold on to unnecessary extra data, particularly in the requests.


smartsim/_core/mli/infrastructure/control/request_dispatcher.py
```diff
@@ -493,10 +493,8 @@ def flush_requests(self) -> None:
                 if queue.ready:
                     self._perf_timer.measure_time("find_queue")
                     try:
-                        batch = RequestBatch(
-                            requests=queue.flush(),
-                            inputs=None,
-                            model_id=queue.model_id,
+                        batch = RequestBatch.from_requests(
```
ankona, line 498: Something feels off about passing the requests and the inputs in here. It might be the type of the input...

~~If this method accepts a `TransformInputRequest`, I expect that input to have gone through the workers transformation stage.... That would be something like this, but the worker obviously isn't available:~~

I'm wrong.. I'm thinking about `TransformInputResponse`....

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -191,6 +191,18 @@ def has_output_keys(self) -> bool:
...
 
 
+@dataclass
+class TensorMeta:
+    """Metadata about a tensor, built from TensorDescriptors."""
+
+    dimensions: t.List[int]
+    """Dimensions of the tensor"""
+    order: str
+    """Order of the tensor"""
```
ankona: Consider being more specific here:
`In-memory order of the tensor. E.g. supply C or R to indicate column or row orientation`

ankona: @al-rigazzi what is a good, backend-agnostic description?

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -191,6 +191,18 @@ def has_output_keys(self) -> bool:
...
+    datatype: str
+    """Datatype of the tensor"""
```
ankona: I'm potentially suggesting something bad that makes it harder to keep the comment up to date, but this could have an example : foo.datatype = str(np.int32) or str(int) or ... anything to know what this string should really be

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -316,20 +332,31 @@ def __init__(self, result: bytes) -> None:
 class RequestBatch:
     """A batch of aggregated inference requests."""
 
-    requests: list[InferenceRequest]
-    """List of InferenceRequests in the batch"""
+    raw_model: t.Optional[Model]
+    """Raw bytes of the model"""
+    callbacks: t.List[CommChannelBase]
+    """The channels used for notification of inference completion"""
```
ankona: I think we should switch these callbacks to descriptors (and create a ticket for doing the same in the InferenceReqeust). It's my fault, but I think this has caused some headaches with serialization... The worker should be constructing their own channels from the descriptor, instead.

ankona: feel free to turn this into a follow up ticket. we want to avoid sending the actual channel around and just send descriptors so the sender can create what they need...

smartsim/_core/mli/infrastructure/control/request_dispatcher.py
AlyssaCote: After our conversation, I've removed the `inputs` parameter from `RequestBatch.from_requests`!

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -316,20 +332,31 @@ def __init__(self, result: bytes) -> None:
...
+    """A dictionary mapping callbacks to output keys"""
     inputs: t.Optional[TransformInputResult]
     """Transformed batch of input tensors"""
     model_id: "ModelIdentifier"
     """Model (key, descriptor) tuple"""
 
     @property
-    def has_valid_requests(self) -> bool:
-        """Returns whether the batch contains at least one request.
+    def has_callbacks(self) -> bool:
```
ankona, line 355: I want to argue that saying we should use `has_callbacks` as a proxy for `has_request` is going to bite us later. Can we just rephrase this to avoid mentioning requests?

`Determines if the batch has at least one callback channel available for sending results.`

tests/dragon/utils/msg_pump.py
```diff
@@ -174,6 +174,8 @@ def _mock_messages(
     # keep the process alive for an extra 15 seconds to let the processor
     # have access to the channels before they're destroyed
     for _ in range(15):
+        # TODO: Chris says he wants to fix something here!
```
al-rigazzi: Can @ankona create a ticket to keep track of this?

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -339,37 +366,49 @@ def has_raw_model(self) -> bool:
...
-        for request in self.requests:
-            keys.extend(request.output_keys)
-
-        return keys
+    @classmethod
+    def from_requests(
+        cls,
+        requests: t.List[InferenceRequest],
+        inputs: t.Optional[TransformInputResult],
+        model_id: ModelIdentifier,
```
ankona, line 375: Do we need `model_id`? Is it reliable? It makes some sense that we're sending a whole batch for processing by the same model...

But - if I put a request on the wrong queue, the code as-is would probably blow up because it doesn't confirm all the requests in the batch have the same model id as what was passed... If we say we trust the dispatcher, why send the model ID instead of just grabbing from the first request in the batch?

ankona: Is this unnecessary?

```
    @classmethod
    def from_requests(
        cls,
        requests: t.List[InferenceRequest],
        model_id: ModelIdentifier,
    ) -> "RequestBatch":
        """Create a RequestBatch from a list of requests.
        ...
        """
        my_requests = [r for r in requests if r.model_key == model_id]
        not_my_requests = [r for r in requests if r.model_key != model_id]

        if not_my_requests:
            logger.warning("Attempted inference call with model ID mismatch")
```

If yes, why not do:

```
return cls(model_id=requests[0].model_key, ...
```

ankona: including @al-rigazzi in case he knows this off the top of his head

tests/dragon/utils/channel.py
```diff
@@ -54,6 +54,16 @@ def __init__(self, key: pathlib.Path) -> None:
 
         self._file_path.touch()
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
```
ankona: this is danger-town and should go away. This is caused by inclusion of the channel in the InferenceRequest.  Swapping those channels with descriptors will avoid this ticking time-bomb

tests/dragon/utils/msg_pump.py
ankona: Mr Memory over here doesn't remember either. Feel free to remove this comment and I'll create a ticket when i remember..

tests/mli/channel.py
```diff
@@ -54,6 +54,16 @@ def __init__(self, key: pathlib.Path) -> None:
 
         self._file_path.touch()
 
+    def __getstate__(self):
```
ankona: ditto on dangerousness of removing the lock. Let's fix the channels instead

ankona (commented):
Looks pretty good. 

You're inheriting a problem of my making (attempting to serialize channels and locks). Now that you're wiping out the locks in a dangerous fashion, you probably need to inherit and fix the underlying cause or create a follow-up ticket to come back to it.

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: Working on changing these to all be descriptors now!

tests/mli/channel.py
AlyssaCote: removed!

tests/dragon/utils/msg_pump.py
AlyssaCote: removed!

tests/dragon/utils/channel.py
AlyssaCote: the danger town is gone and can't hurt us anymore!

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: Finished this but will definitely need your opinions on it

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: added more description

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: Added more description! This datatype string is pulled from enums in `tensor_capnp` so I stated that and added examples.

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: I'll wait for Al to chime in, but I totally see what you're saying!

al-rigazzi: The whole batching is based on `model_id`s. A batch is always only created from requests which come from the same queue and the dispatcher puts requests in queues based on their `model_id`s. So, at this point, either all requests have the same `model_id` or something went really wrong.

I would be in favor of 
```python
return cls(model_id=requests[0].model_key, ...
```
as @ankona suggests, but we could also enforce a (very defensive imho) check that all `model_id`s are the same.

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -507,49 +544,44 @@ def fetch_inputs(
...
         :raises SmartSimError: If a tensor for a given key cannot be retrieved
         """
-        fetch_results = []
-        for request in batch.requests:
-            if request.raw_inputs:
-                fetch_results.append(
-                    FetchInputResult(request.raw_inputs, request.input_meta)
-                )
-                continue
+        if not batch.raw_inputs and not batch.input_keys:
```
al-rigazzi, line 547: There are a couple issues with this refactor.
- this condition and the one at line 550 will always evaluate to `True`, because `batch.raw_inputs` and `batch.input_keys` are lists (possibly containing `None` many times) and are never empty. This would mean that even if `raw_inputs==[None]`, we would assign that to `FetchInputResults` and return at line 551.
- there is no guarantee that a user is not mixing `raw_inputs` and `input_keys` in the same (batched) request. Thus we need to iterate over `zip(raw_inputs, input_keys)` and resolve each couple individually, storing the resulting `raw_inputs` and `input_meta` as lists that are assigned then passed to `FetchInputResult` once the iterations are over.

al-rigazzi (changes requested):
I believe that a couple of changes are needed, but it looks like a great start!

smartsim/_core/mli/infrastructure/worker/worker.py
```diff
@@ -316,20 +334,31 @@ def __init__(self, result: bytes) -> None:
...
-        """Returns whether the batch contains at least one request.
+    def has_callbacks(self) -> bool:
+        """Determines if the batch has at least one callback channel
+        available for sending results.
 
-        :returns: True if at least one request is available
+        :returns: True if at least one callback is present
         """
-        return len(self.requests) > 0
+        return len(self.callback_descriptors) > 0
```
al-rigazzi, line 361: I think the list will always be populated (maybe with `None`) so this test should always return True. Would it make sense to turn it into something like `any(desc is not None for desc in self.callback_descriptors)`?

AlyssaCote: The way I gather the callback_descriptors from the requests is
```py
callback_descriptors=[
      request.callback_desc for request in requests if request.callback_desc
],
```
so there won't be any `None` values and if there are no callbacks in any of the requests it'll be an empty list.

smartsim/_core/mli/infrastructure/worker/worker.py
ankona: I just realized that a direct-inference model wouldn't have the model ID. I think I was wrong.. Removing model_id makes it not work, doesn't it?

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: Here's how I was understanding it but I could be totally wrong. In `InferenceRequest`, passed in None values for `raw_inputs,` `input_keys`, `input_meta`, and `output_keys` have `self.raw_inputs = raw_inputs or []`, so if  a `None` value is passed in, it becomes an empty list. When we instantiate a `RequestBatch` using `from_requests`, I do it like this:
```python
raw_inputs=[
       request.raw_inputs for request in requests if request.raw_inputs
],
```
meaning that if all the requests have an empty list in place of `raw_requests`, for example, they will just get skipped over and those empty lists are not placed into the `batch.raw_requests` list, leaving `batch.raw_requests` an empty list.

I added this test just to be sure.
```python
def test_fetch_inputs_no_input_source():
    """Verify that the ML worker fails when there are no inputs provided"""
    worker = MachineLearningWorkerCore
    request1 = InferenceRequest()
    request2 = InferenceRequest()

    batch = RequestBatch.from_requests([request1, request2], ModelKey("test-model", "desc"))

    with pytest.raises(ValueError) as exc:
        fetch_result = worker.fetch_inputs(batch, {})
    assert str(exc.value) == "No input source"
```

AlyssaCote: But in regards to mixing `input_keys` and `raw_inputs`, I definitely didn't think about that! I will fix that asap

smartsim/_core/mli/infrastructure/control/worker_manager.py
```diff
@@ -226,18 +228,20 @@ def _on_iteration(self) -> None:
...
                 return
             self._perf_timer.measure_time("load_model")
 
             if not batch.inputs:
-                for request in batch.requests:
+                for callback_desc in batch.callback_descriptors:
                     exception_handler(
                         ValueError("Error batching inputs"),
-                        request.callback,
+                        self._callback_factory(callback_desc),
```
ankona, line 244: not to see-saw here, but it might be just a little more convenient if you pass your descriptor to `exception_handler` and let it create the channel instead of each call having a copy of that call to the callback factory

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: I think I'd prefer keeping it as a value passed in because `model_id=requests[0].model_key` could potentially be `None`? I'm not sure anymore lol

smartsim/_core/mli/infrastructure/worker/worker.py
al-rigazzi: I see your point, you're right, an empty list is possible and that would be a False-y value.

smartsim/_core/mli/infrastructure/control/worker_manager.py
ankona: per convo, ok w/leaving as-is to just avoid churn

smartsim/_core/mli/infrastructure/worker/worker.py
al-rigazzi: Right again

tests/dragon/utils/msg_pump.py
ankona: it's ok to remove my note, now

AlyssaCote: already gone!

smartsim/_core/mli/infrastructure/worker/worker.py
ankona: LGTM

smartsim/_core/mli/infrastructure/worker/worker.py
AlyssaCote: I believe I've resolved both of your points in the original comment!

al-rigazzi (approved):
LGTM! Thanks for this great refactor!

ankona (approved):
LGTM once the base conflicts are resolved.

AlyssaCote:
Closed because this PR was merging into the wrong branch.

PR #744, opened 2024-10-14
`Experiment.start` Returns Records

MattToast:
Basic implementation of a `Record` class to keep track of launched jobs. Convert `Experiment.start` return `Record`s and all other methods to accept records.

### DIFF
--- a/smartsim/experiment.py
+++ b/smartsim/experiment.py
@@ -161,11 +161,10 @@ class Experiment:
         """Execute a collection of `Job` instances.
 
         :param jobs: A collection of other job instances to start
-        :raises TypeError: If jobs provided are not the correct type
         :raises ValueError: No Jobs were provided.
-        :returns: A sequence of ids with order corresponding to the sequence of
-            jobs that can be used to query or alter the status of that
-            particular execution of the job.
+        :returns: A sequence of records with order corresponding to the
+            sequence of jobs that can be used to query or alter the status of
+            that particular execution of the job.
         """
 
         if not jobs:
@@ -232,12 +231,12 @@ class Experiment:
         )
 
     def get_status(self, *records: Record) -> tuple[JobStatus | InvalidJobStatus, ...]:
-        """Get the status of jobs launched through the `Experiment` from their
-        launched job id returned when calling `Experiment.start`.
+        """Get the status of jobs launched through the `Experiment` from the
+        record returned when calling `Experiment.start`.
 
-        The `Experiment` will map the launched ID back to the launcher that
-        started the job and request a status update. The order of the returned
-        statuses exactly matches the order of the launched job ids.
+        The `Experiment` will map the launched id of the record back to the
+        launcher that started the job and request a status update. The order of
+        the returned statuses exactly matches the order of the records.
 
         If the `Experiment` cannot find any launcher that started the job
         associated with the launched job id, then a
@@ -450,7 +449,7 @@ class Experiment:
         """
         if not records:
             raise ValueError("No records provided")
-        if not all(isinstance(reocrd, Record) for reocrd in records):
+        if not all(isinstance(record, Record) for record in records):
             raise TypeError("record argument was not of type Record")
         ids = tuple(record.launched_id for record in records)
         by_launcher = self._launch_history.group_by_launcher(set(ids), unknown_ok=True)
--- a/smartsim/experiment.py
+++ b/smartsim/experiment.py
@@ -42,7 +42,7 @@ from smartsim._core.control import preview_renderer
 from smartsim._core.control.launch_history import LaunchHistory as _LaunchHistory
 from smartsim._core.utils import helpers as _helpers
 from smartsim.error import errors
-from smartsim.launchable.job import Job
+from smartsim.launchable.job import Job, Record
 from smartsim.status import TERMINAL_STATUSES, InvalidJobStatus, JobStatus
 
 from ._core import Generator, Manifest
@@ -52,7 +52,6 @@ from .error import SmartSimError
 from .log import ctx_exp_path, get_logger, method_contextualizer
 
 if t.TYPE_CHECKING:
-    from smartsim.launchable.job import Job
     from smartsim.types import LaunchedJobID
 
 logger = get_logger(__name__)
@@ -158,14 +157,7 @@ class Experiment:
         experiment
         """
 
-    def _set_dragon_server_path(self) -> None:
-        """Set path for dragon server through environment varialbes"""
-        if not "SMARTSIM_DRAGON_SERVER_PATH" in environ:
-            environ["_SMARTSIM_DRAGON_SERVER_PATH_EXP"] = osp.join(
-                self.exp_path, CONFIG.dragon_default_subdir
-            )
-
-    def start(self, *jobs: Job | t.Sequence[Job]) -> tuple[LaunchedJobID, ...]:
+    def start(self, *jobs: Job | t.Sequence[Job]) -> tuple[Record, ...]:
         """Execute a collection of `Job` instances.
 
         :param jobs: A collection of other job instances to start
@@ -179,12 +171,11 @@ class Experiment:
         if not jobs:
             raise ValueError("No jobs provided to start")
 
-        # Create the run id
         jobs_ = list(_helpers.unpack(jobs))
-
         run_id = datetime.datetime.now().replace(microsecond=0).isoformat()
-        root = pathlib.Path(self.exp_path, run_id.replace(":", "."))
-        return self._dispatch(Generator(root), dispatch.DEFAULT_DISPATCHER, *jobs_)
+        root = pathlib.Path(self.exp_path, run_id)
+        ids = self._dispatch(Generator(root), dispatch.DEFAULT_DISPATCHER, *jobs_)
+        return tuple(Record(id_, job) for id_, job in zip(ids, jobs_))
 
     def _dispatch(
         self,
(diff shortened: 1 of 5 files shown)


smartsim/experiment.py
```diff
@@ -158,7 +157,7 @@ def __init__(self, name: str, exp_path: str | None = None):
         experiment
         """
 
-    def start(self, *jobs: Job | t.Sequence[Job]) -> tuple[LaunchedJobID, ...]:
+    def start(self, *jobs: Job | t.Sequence[Job]) -> tuple[Record, ...]:
```
amandarichardsonn, line 160: Do you need to update the :return: in the start docstring?

smartsim/experiment.py
```diff
@@ -158,7 +157,7 @@ def __init__(self, name: str, exp_path: str | None = None):
...
         """Execute a collection of `Job` instances.
```
amandarichardsonn, line 161: It looks like there is a :raises TypeError: but I do not see a TypeError raised in this start?

smartsim/experiment.py
```diff
@@ -233,9 +231,7 @@ def execute_dispatch(generator: Generator, job: Job, idx: int) -> LaunchedJobID:
             execute_dispatch(generator, job, idx) for idx, job in enumerate(jobs, 1)
         )
 
-    def get_status(
-        self, *ids: LaunchedJobID
-    ) -> tuple[JobStatus | InvalidJobStatus, ...]:
+    def get_status(self, *records: Record) -> tuple[JobStatus | InvalidJobStatus, ...]:
         """Get the status of jobs launched through the `Experiment` from their
         launched job id returned when calling `Experiment.start`.
```
amandarichardsonn: docstring nit: job id -> Record?

smartsim/launchable/job.py
```diff
@@ -158,3 +159,35 @@ def __str__(self) -> str:  # pragma: no cover
         string = f"SmartSim Entity: {self.entity}\n"
         string += f"Launch Settings: {self.launch_settings}"
         return string
+
+
+@t.final
+class Record:
+    """A record a job that was launched and launch ID assigned to the
```
amandarichardsonn: doc string nit `A record a job that was launched and launch ID assigned to the launching` -> `A Record object to track a launched job along with its assigned launch ID`?

smartsim/launchable/job.py
```diff
@@ -158,3 +159,35 @@ def __str__(self) -> str:  # pragma: no cover
...
+    launching.
+    """
+
+    def __init__(self, launch_id: LaunchedJobID, job: Job) -> None:
+        """Initialize a new record of a launched job
+
+        :param launch_id: A unique identifier for the launch of the job.
+        :param job: The job that was launched.
+        """
+        self._id = launch_id
```
amandarichardsonn, line 177: Why do you deep copying the Job but not the launch_id here?

smartsim/launchable/job.py
```diff
@@ -158,3 +159,35 @@ def __str__(self) -> str:  # pragma: no cover
...
+    def launched_id(self) -> LaunchedJobID:
+        """The unique identifier for the launched job.
+
+        :returns: A unique identifier for the launched job.
+        """
+        return self._id
+
+    @property
+    def job(self) -> Job:
+        """A deep copy of the job that was launched.
```
amandarichardsonn, line 190: would adding `def __str__(self)` to the Record class be beneficial? For when a user logs the records or prints

smartsim/experiment.py
```diff
@@ -445,20 +439,20 @@ def summary(self, style: str = "github") -> str:
...
             respective of the order of the calling arguments.
         """
-        if ids:
-            if not all(isinstance(id, str) for id in ids):
-                raise TypeError("ids argument was not of type LaunchedJobID")
-        else:
-            raise ValueError("No job ids provided")
+        if not records:
+            raise ValueError("No records provided")
+        if not all(isinstance(reocrd, Record) for reocrd in records):
```
amandarichardsonn: it looks like you have `if not all(isinstance(reocrd, Record)` where record is misspelled

amandarichardsonn (changes requested):
Some small suggestions!

smartsim/experiment.py
MattToast: d'oh! 🤦 I knew I let  a few of these slip through the cracks... yeah! that should be updated!!

smartsim/experiment.py
MattToast: Whoops! I think that check got moved into `unpack` and we forgot to update the docstrings. I can move it there rq!

smartsim/launchable/job.py
MattToast: Every time I read my own writing, I become more and more convinced that I'm not actually fluent in English, hahaha. Updated to make it coherent!

smartsim/launchable/job.py
MattToast: In theory, the `launch_id` (at least for now), is just a string that we ask the type checker to treat like a different type, and strings are immutable in python, so a "deepcopy of" and a "reference to" a string are functionally equivalent.

There is actually some special logic in [`copy.deepcopy`](https://github.com/python/cpython/blob/3.13/Lib/copy.py#L119-L137) so that, by default, [it will just return the original reference to a string](https://github.com/python/cpython/blob/3.13/Lib/copy.py#L173-L183) if its encounter without change.

TL;DR I was lazy and didn't want to spell out "deepcopy". More than willing to add it though if we think that `LaunchedJobID` might become a more substantive type in future!

smartsim/launchable/job.py
MattToast: Probably not a bad idea! I'll add one!

smartsim/experiment.py
MattToast: This is what I get for relying too much on intellisense lol

smartsim/launchable/job.py
MattToast: Added in [2e367a0](https://github.com/CrayLabs/SmartSim/pull/744/commits/2e367a0defd708d7a71e8c233901d7e9686d0fc4)

smartsim/experiment.py
```diff
@@ -233,15 +230,13 @@ def execute_dispatch(generator: Generator, job: Job, idx: int) -> LaunchedJobID:
...
-        """Get the status of jobs launched through the `Experiment` from their
-        launched job id returned when calling `Experiment.start`.
+    def get_status(self, *records: Record) -> tuple[JobStatus | InvalidJobStatus, ...]:
+        """Get the status of jobs launched through the `Experiment` from the
+        record returned when calling `Experiment.start`.
 
-        The `Experiment` will map the launched ID back to the launcher that
-        started the job and request a status update. The order of the returned
-        statuses exactly matches the order of the launched job ids.
+        The `Experiment` will map the launched id of the record back to the
```
ankona: the ID isn't launched.  could just say `will map the record id to the ...`

smartsim/experiment.py
```diff
@@ -233,15 +230,13 @@ def execute_dispatch(generator: Generator, job: Job, idx: int) -> LaunchedJobID:
...
+        launcher that started the job and request a status update. The order of
+        the returned statuses exactly matches the order of the records.
```
ankona: can omit exactly and just say `matches`

smartsim/experiment.py
```diff
@@ -445,20 +438,20 @@ def summary(self, style: str = "github") -> str:
...
         :returns: A tuple of job statuses upon cancellation with order
             respective of the order of the calling arguments.
         """
-        if ids:
-            if not all(isinstance(id, str) for id in ids):
-                raise TypeError("ids argument was not of type LaunchedJobID")
-        else:
-            raise ValueError("No job ids provided")
+        if not records:
+            raise ValueError("No records provided")
```
ankona: consider making the log more descriptive about what was attempted "No records provided. No jobs will be stopped."

smartsim/launchable/job.py
```diff
@@ -158,3 +160,44 @@ def __str__(self) -> str:  # pragma: no cover
         string = f"SmartSim Entity: {self.entity}\n"
         string += f"Launch Settings: {self.launch_settings}"
         return string
+
+
+@t.final
+class Record:
+    """A Record object to track a launched job along with its assigned
```
ankona: does it really "track" the job or would something like `A unique identifier for a launched job` express what it does?

smartsim/launchable/job.py
```diff
@@ -158,3 +160,44 @@ def __str__(self) -> str:  # pragma: no cover
...
+        :returns: A deep copy of the launched job.
+        """
+        return deepcopy(self._job)
+
+    def __str__(self) -> str:
+        return textwrap.dedent(f"""\
+            Launch Record:
+                Launched Job ID:
+                    {self.launched_id}
+                Laucnehd Job:
```
ankona: typo in laucnehd

smartsim/experiment.py
MattToast: Rephrased!

smartsim/experiment.py
MattToast: "exactly" dropped!

smartsim/experiment.py
MattToast: More detailed message issued!

smartsim/launchable/job.py
MattToast: Good point! I updated the docstring to be a bit more descriptive of the composition of the object!

smartsim/launchable/job.py
MattToast: 🤦

PR #753, opened 2024-10-17
Update SmartSim to use Dragon V0.10

MattToast:
Update SmartSim to use [Dragon V0.10](https://github.com/DragonHPC/dragon/releases/tag/v0.10-beta) by:
- [x] Removing the use of Dragon attributes that have been removed in Dragon V0.10.
- [x] Updating `smart build --dragon ...` to fetch Dragon V0.10 once all required assets are hosted
- [ ] Run through full on WLM test suite without errors

### DIFF
--- a/smartsim/_core/launcher/dragon/dragonBackend.py
+++ b/smartsim/_core/launcher/dragon/dragonBackend.py
@@ -30,7 +30,6 @@ import itertools
 import time
 import typing as t
 from dataclasses import dataclass, field
-from enum import Enum
 from threading import RLock
 
 from tabulate import tabulate
--- a/smartsim/_core/launcher/dragon/dragonBackend.py
+++ b/smartsim/_core/launcher/dragon/dragonBackend.py
@@ -23,6 +23,7 @@
 # CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 import collections
 import functools
 import itertools
@@ -38,7 +39,6 @@ from tabulate import tabulate
 # isort: off
 import dragon.infrastructure.connection as dragon_connection
 import dragon.infrastructure.policy as dragon_policy
-import dragon.native.group_state as dragon_group_state
 import dragon.native.process as dragon_process
 import dragon.native.process_group as dragon_process_group
 import dragon.native.machine as dragon_machine
@@ -67,33 +67,47 @@ from ....status import TERMINAL_STATUSES, SmartSimStatus
 logger = get_logger(__name__)
 
 
-class DragonStatus(str, Enum):
-    ERROR = str(dragon_group_state.Error())
-    RUNNING = str(dragon_group_state.Running())
-
-    def __str__(self) -> str:
-        return self.value
-
-
 @dataclass
 class ProcessGroupInfo:
     status: SmartSimStatus
     """Status of step"""
     process_group: t.Optional[dragon_process_group.ProcessGroup] = None
     """Internal Process Group object, None for finished or not started steps"""
-    puids: t.Optional[t.List[t.Optional[int]]] = None  # puids can be None
-    """List of Process UIDS belonging to the ProcessGroup"""
-    return_codes: t.Optional[t.List[int]] = None
-    """List of return codes of completed processes"""
     hosts: t.List[str] = field(default_factory=list)
     """List of hosts on which the Process Group """
     redir_workers: t.Optional[dragon_process_group.ProcessGroup] = None
     """Workers used to redirect stdout and stderr to file"""
 
     @property
-    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.Optional[t.List[int]]]:
+    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.List[int]]:
         """Information needed by SmartSim Launcher and Job Manager"""
         return (self.status, self.return_codes)
+
+    @property
+    def puids(self) -> t.List[int]:
+        """List of Process UIDS belonging to the ProcessGroup"""
+        return list(set(itertools.chain(self.active_puids, self.inactive_puids)))
+
+    @property
+    def active_puids(self) -> t.List[int]:
+        if self.process_group is None:
+            return []
+        return list(self.process_group.puids)
+
+    @property
+    def inactive_puids(self) -> t.List[int]:
+        if self.process_group is None:
+            return []
+        return [puid for puid, _ in self.process_group.inactive_puids]
+
+    @property
+    def return_codes(self) -> t.List[int]:
+        """List of return codes of completed processes"""
+        if self.process_group is None:
+            return [-1]
+        if self.status == SmartSimStatus.STATUS_CANCELLED:
+            return [-9]
+        return [ret for _, ret in self.process_group.inactive_puids]
 
     def __str__(self) -> str:
         if self.process_group is not None and self.redir_workers is not None:
@@ -105,7 +119,7 @@ class ProcessGroupInfo:
 
         if self.hosts is not None:
             msg.append(f"Hosts: {','.join(self.hosts)}")
-        if self.return_codes is not None:
+        if self.return_codes:
             msg.append(f"{self.return_codes}")
 
         return ", ".join(msg)
@@ -404,10 +418,10 @@ class DragonBackend:
                 else:
                     # Technically we could just terminate, but what if
                     # the application intercepts that and ignores it?
-                    proc_group = self._group_infos[step_id].process_group
+                    group_info = self._group_infos[step_id]
                     if (
-                        proc_group is not None
-                        and proc_group.status == DragonStatus.RUNNING
+                        group_info.active_puids
+                        and (proc_group := group_info.process_group) is not None
                     ):
                         try:
                             proc_group.kill()
@@ -416,7 +430,7 @@ class DragonBackend:
                                 proc_group.stop()
                             except dragon_process_group.DragonProcessGroupError:
                                 logger.error("Process group already stopped")
-                    redir_group = self._group_infos[step_id].redir_workers
+                    redir_group = group_info.redir_workers
                     if redir_group is not None:
                         try:
                             redir_group.join(0.1)
@@ -425,7 +439,6 @@ class DragonBackend:
                             logger.error(e)
 
                 self._group_infos[step_id].status = SmartSimStatus.STATUS_CANCELLED
-                self._group_infos[step_id].return_codes = [-9]
 
     @staticmethod
     def create_run_policy(
@@ -438,7 +451,6 @@ class DragonBackend:
         if isinstance(request, DragonRunRequest):
             run_request: DragonRunRequest = request
 
-            affinity = dragon_policy.Policy.Affinity.DEFAULT
             cpu_affinity: t.List[int] = []
             gpu_affinity: t.List[int] = []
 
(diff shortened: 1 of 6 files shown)


smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -620,7 +607,7 @@ def _refresh_statuses(self) -> None:
                         except KeyError:
                             logger.error(f"Tried to free a non-allocated host: {host}")
                         self._free_hosts.append(host)
-                    group_info.process_group = None
+                    # group_info.process_group = None
                     group_info.redir_workers = None
```
MattToast: Question: Is there a reason that we were previously nulling out this attr(s)? I needed to do this to make sure that return codes did not come back as `[-1]`, but if there is a reason for doing this I can makes something a bit more sophisticated for how I retrieve return codes (e.g. cache them and then null out the process group).

al-rigazzi: I believe I wanted to release process groups which finished, but if we see that things work fine, we can keep them.

al-rigazzi (approved):
LGTM!

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -620,7 +606,7 @@ def _refresh_statuses(self) -> None:
                         except KeyError:
                             logger.error(f"Tried to free a non-allocated host: {host}")
                         self._free_hosts.append(host)
-                    group_info.process_group = None
+                    # group_info.process_group = None
```
ankona: don't forget to remove your commented code

tests/test_dragon_installer.py
```diff
@@ -187,26 +190,39 @@ def test_retrieve_cached(
...
-        pytest.param("0.9", "py3.8", False, True, id="0.9,python 3.8,CrayEX"),
-        pytest.param("0.9", "py3.9", True, True, id="0.9,python 3.9,CrayEX"),
-        pytest.param("0.9", "py3.10", True, True, id="0.9,python 3.10,CrayEX"),
-        pytest.param("0.9", "py3.11", True, True, id="0.9,python 3.11,CrayEX"),
-        pytest.param("0.9", "py3.12", False, True, id="0.9,python 3.12,CrayEX"),
+        pytest.param("0.9", "py3.8", False, False, id="0.9,python 3.8,CrayEX"),
+        pytest.param("0.9", "py3.9", True, False, id="0.9,python 3.9,CrayEX"),
+        pytest.param("0.9", "py3.10", True, False, id="0.9,python 3.10,CrayEX"),
+        pytest.param("0.9", "py3.11", True, False, id="0.9,python 3.11,CrayEX"),
+        pytest.param("0.9", "py3.12", False, False, id="0.9,python 3.12,CrayEX"),
```
ankona: It appears that these tests no longer do what they were intended to. I think 210-214 should all have true as `is_crayex` since 205-209 tested the !HSN case. Otherwise, the second set of parameterizations is the same as the first half?

ankona: Consider updating the verbiage since you're already here. Swap `CrayEX` with `HSN` in test names and in the parameterization (is_crayex becomes is_hsn)

tests/test_dragon_installer.py
```diff
@@ -102,23 +103,25 @@ def test_assets(monkeypatch: pytest.MonkeyPatch) -> t.Dict[str, GitReleaseAsset]
...
-                monkeypatch.setattr(
-                    asset,
-                    "_browser_download_url",
-                    _git_attr(value=f"http://foo/{archive_name}"),
-                )
-                monkeypatch.setattr(asset, "_name", _git_attr(value=archive_name))
-                assets.append(asset)
+    for python_version, dragon_version, platform in itertools.chain(
+        itertools.product(
+            ["py3.9", "py3.10", "py3.11"], ["dragon-0.8", "dragon-0.9"], ["", "CRAYEX-"]
```
ankona: leaving crayex here is confusing. the emptystring in the test was meant to simulate diff between `mypackage-py3.9.whl` and `mypackage-py3.9.CRAYEX.whl`. Here, the result is the same but it is misleading...

It looks like you're actually already solving the problem with the `itertools.product` and the second `.product` is just repeating it. If you keep only the `itertools.product[... "HSN-"]`, i think you're fine?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -438,33 +450,27 @@ def create_run_policy(
...
             if run_request.policy is not None:
                 # Affinities are not mutually exclusive. If specified, both are used
                 if run_request.policy.cpu_affinity:
-                    affinity = dragon_policy.Policy.Affinity.SPECIFIC
                     cpu_affinity = run_request.policy.cpu_affinity
 
                 if run_request.policy.gpu_affinity:
-                    affinity = dragon_policy.Policy.Affinity.SPECIFIC
                     gpu_affinity = run_request.policy.gpu_affinity
             logger.debug(
```
ankona, line 502: i caught this removal on my latest PR, too. looking forward to our merge conflict!

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -67,34 +66,48 @@
...
     @property
-    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.Optional[t.List[int]]]:
+    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.List[int]]:
         """Information needed by SmartSim Launcher and Job Manager"""
         return (self.status, self.return_codes)
 
+    @property
+    def puids(self) -> t.List[int]:
+        """List of Process UIDS belonging to the ProcessGroup"""
+        return list(set(itertools.chain(self.active_puids, self.inactive_puids)))
```
ankona, line 98: love this avoidance of none

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -67,34 +66,48 @@
...
+    def inactive_puids(self) -> t.List[int]:
+        if self.process_group is None:
+            return []
+        return [puid for puid, _ in self.process_group.inactive_puids]
+
+    @property
+    def return_codes(self) -> t.List[int]:
+        """List of return codes of completed processes"""
+        if self.process_group is None:
+            return [-1]
```
ankona: should -1 and -9 be enshrined somewhere (enum, constant) so they have a descriptive name? Where do -1 and -9 come from?

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -67,34 +66,48 @@
...
     redir_workers: t.Optional[dragon_process_group.ProcessGroup] = None
     """Workers used to redirect stdout and stderr to file"""
 
     @property
-    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.Optional[t.List[int]]]:
+    def smartsim_info(self) -> t.Tuple[SmartSimStatus, t.List[int]]:
         """Information needed by SmartSim Launcher and Job Manager"""
         return (self.status, self.return_codes)
 
+    @property
```
ankona, line 92: Confirm that your docstrings match latest standard from @AlyssaCote. I think that entails an `:returns:` notation and a period terminating the method description.

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -67,34 +66,48 @@
...
+    def puids(self) -> t.List[int]:
+        """List of Process UIDS belonging to the ProcessGroup"""
+        return list(set(itertools.chain(self.active_puids, self.inactive_puids)))
+
+    @property
+    def active_puids(self) -> t.List[int]:
+        if self.process_group is None:
```
ankona, line 106: docstring

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -67,34 +66,48 @@
...
+            return []
+        return list(self.process_group.puids)
+
+    @property
+    def inactive_puids(self) -> t.List[int]:
+        if self.process_group is None:
```
ankona, line 116: docstring

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Uhh... I know that it works _behaviorally_ but I've not checked for actual resource utilization (e.g. does this need to be nulled out in order for dragon to garbage collect, etc.). I might need to take some time to take a look at that!

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: > should -1 and -9 be enshrined somewhere (enum, constant) so they have a descriptive name?

Probably, hahahaha. I can add one!

> Where do -1 and -9 come from?

Honestly not sure. I used them for backward compatibility reasons since that was what they were manually set to when a group info was created w/o a process group and when a process group was manually canceled previously. @al-rigazzi might have some better insight here?

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: LMAO that dragon API break and our multiple long lived branches is going to be the death of us! 😅

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: I think we can se the standard System Error Codes?

tests/test_dragon_installer.py
MattToast: Oh! Was the point of this fixture not to mimic the names of the assets from the [dragon releases](https://github.com/DragonHPC/dragon/releases)?

If so, I'm tempted to keep the chain as dragon v0.8 and v0.9 use "CRAYEX" in the asset name, while v0.10 does (and presumably v0.11 will) use "HSN".

If not, totally willing to collapse this down to just the single `product`

tests/test_dragon_installer.py
MattToast: Whoops! I flipped the `is_crayex` flag when I meant to flip the `is_found` flag... 🤦 Let me fix that rq

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Removed!

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Docstrs added!

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Docstrs added!

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Docstrs added!

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -105,11 +142,19 @@ def __str__(self) -> str:
...
         if self.hosts is not None:
             msg.append(f"Hosts: {','.join(self.hosts)}")
-        if self.return_codes is not None:
+        if self.return_codes:
             msg.append(f"{self.return_codes}")
 
         return ", ".join(msg)
 
+    def mark_complete(self) -> None:
+        """Cached the final return codes and release any underlying dragon
```
MattToast:
```suggestion
        """Cache the final return codes and release any underlying dragon
```

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Changed to cache return codes when marking process complete!

tests/test_dragon_installer.py
ankona: It is definitely meant to mimic those release assets. There are two packages per python version, for example:

- `dragon-0.10-py3.10.10-7edea3b6f.tar.gz`
- `dragon-0.10-py3.10.10-HSN-4e64b74bf.tar.gz`

In your sample test data, cray ex carries a sort of historical connotation since that was the name of the files before HSN replaced it. I think having CRAYEX does prove out that we'd chose the HSN version when we're supposed to, but that historical meaning makes it appear to be an important distinction, which it no longer is.

In the examples for py3.10 above, we just see that you either have a segment denoting HSN capability or you don't... so, i feel like we don't need to look for `FOO-\` or `CRAYEX-` as long as the hsn and non-hsn is covered.

Your first set of tests will all be negative for hsn.
- `["py3.9", "py3.10", "py3.11"], ["dragon-0.8", "dragon-0.9"], ["", "CRAYEX-"]`

The second set of tests exercise both hsn and non-hsn packages
- `["py3.9", "py3.10", "py3.11"], ["dragon-0.10", "dragon-0.11"], ["", "HSN-"]`

So, why have the first set of tests at all if they don't do anything the product with `["", "HSN-"] already accomplishes.

MattToast: Oh! I can buy that!! I kept them around originally bc _technically_ there are assets that _would_ be found during the actual build process by the name `CRAYEX-` at version `v0.9`, but I can 100% get behind the argument that they are not actually _testing_ anything and are just adding noise to the tests. Will remove!

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: We still need the `[-1]` in the case were no process group is actually launched when creating a `ProcessGroupInfo` ([see here](https://github.com/CrayLabs/SmartSim/blob/c8f6decf36f9c1a148c1ae8739e67abcc82e7d5c/smartsim/_core/launcher/dragon/dragonBackend.py#L707-L715)), but we can 100% use the status codes in the case of a process group being cancelled! Will remove!

al-rigazzi:
@MattToast there have been some updates to Dragon's `ProcessGroup` management which require some modification of this code. Do you want me to take care of that by pushing to your branch?

al-rigazzi (changes requested):
A couple minor changes need to happen before this can be merged. They include:
1. change pinned version
2. add arg to ProcessGroup to avoid failing on process exception
3. handle exceptions when stopping groups and redir workers

MattToast:
> @MattToast there have been some updates to Dragon's ProcessGroup management which require some modification of this code. Do you want me to take care of that by pushing to your branch?

@al-rigazzi By all means! Feel free to take over ownership of this branch and push whatever changes are necessary!!

PR #761, opened 2024-10-24
Fix logging bug in dragon_install

AlyssaCote:
I added extra logging depending on the type of mismatch between the asset and the platform. I also return `None` if the asset is a `CRAY EX` asset but the platform is not a `CRAY EX` platform. I broke it down into these potential situations:

| is cray ex platform  | "crayex" in asset name | Results (asset, log output) |
| ------------- | ------------- | ------------- |
| True  | False  | asset, Platform-specific package not found. Using {asset.name} |
| True | True  | asset, no log output |
| False | False | asset, no log output |
| False | True | None, Platform does not support Cray EX assets |

NOTE: The string `crayex` that we check is in the asset name will need to be changed to `hsn` as soon as Matt's work goes in because the asset names now have `hsn` instead of `crayex`.


smartsim/_core/_cli/scripts/dragon_install.py
```diff
@@ -251,8 +251,15 @@ def filter_assets(
         asset = next((asset for asset in assets if _platform_filter(asset.name)), None)
 
     if not asset:
+        # there is a mismatch between the platform and the asset
         asset = assets[0]
-        logger.warning(f"Platform-specific package not found. Using {asset.name}")
+        if "crayex" in asset.name.lower():
```
ankona: I'd prefer it if you made use of the existing `_platform_filter` function to keep the number of copies of what makes an asset HSN in one place. That will ensure that once your code is merged with Toast's fix, you don't have to change anything extra.

smartsim/_core/_cli/scripts/dragon_install.py
```diff
@@ -251,8 +251,15 @@ def filter_assets(
         asset = next((asset for asset in assets if _platform_filter(asset.name)), None)
 
     if not asset:
+        # there is a mismatch between the platform and the asset
```
ankona: I like the intent of this comment but it's not really clear what a mismatch means in this context. Could we be more explicit with something like `# the platform filter eliminated all available assets, try to use what is available`

smartsim/_core/_cli/scripts/dragon_install.py
```diff
@@ -250,7 +250,11 @@ def filter_assets(
     if len(assets) > 0:
         asset = next((asset for asset in assets if _platform_filter(asset.name)), None)
 
-    if not asset:
+    if not asset and not is_crayex_platform():
+        # non-Cray platform, Cray EX asset
+        logger.warning(f"Platform does not support Cray EX assets")
```
ankona: replace "Cray EX" with the more generic HSTA acronym used below and I think we're set.

ankona (changes requested):

smartsim/_core/_cli/scripts/dragon_install.py
AlyssaCote: Replaced `Cray EX` with `HSN`

ankona (approved):

PR #767, opened 2024-10-28
Rewrite RequestBatch

AlyssaCote:
In this PR I rewrote RequestBatch so that it doesn't hold on to unnecessary extra data, particularly in the requests.


al-rigazzi (approved):
Still LGTM! Thanks for this double effort!

PR #769, opened 2024-10-29
Symlink-fix

al-rigazzi:
This PR is an attempt to fix an unexpected behavior encountered when using `SymlinkOperation`, where files were symlinked to their own directory.

I might have mis-interpreted what was going on with the code, though.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -13,6 +13,7 @@ Jump to:
 
 Description
 
+- Fix symlink operation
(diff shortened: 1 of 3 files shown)


MattToast (approved):
LGTM!! Thanks for the catch!

amandarichardsonn (approved):
Thank you for catching and correcting my error 🤦‍♂️

PR #770, opened 2024-10-29
Fix set-hostlist regression from merge

ankona:
Fix a regression introduced during the merge of core refactor into v1.0.

- Ensure host list argument is passed to `DragonRunRequestView` constructor


al-rigazzi (approved):
LGTM, thanks for fixing this!

PR #776, opened 2024-10-31
Fix request dispatcher test

ankona:
This PR includes the following fixes for dragon tests:

1. Remove a duplicate of `test_dragon_comm_utils.py`
2. Refactor `test_request_dispatcher.py` to run without multiprocessing


tests/test_dragon_comm_utils.py
AlyssaCote, line 1: Good catch!

AlyssaCote (approved):
LGTM! Quick question that I doesn't hold up my approval. Should we create a ticket to solve the `test_worker_manager` import issues as well?

tests/test_dragon_comm_utils.py
ankona: Created [ticket 852](https://github.com/CrayLabs/Poseidon/issues/852) to follow-up on failing worker manager test

PR #781, opened 2024-12-18
Drop python 3.9

al-rigazzi:
This PR drops support for Python 3.9


ashao (approved):
LGTM and thanks for catching the vestigial Summit doc page

PR #785, opened 2025-06-05
Add support for Python 3.12

al-rigazzi:
Routine update for SmartSim
- Add support for Python 3.12
- Remove deprecated `distutils` module import
- Bump up TF, Pytorch, and ONNX installed versions as much as possible
- Update installation instruction for new CUDA and ROCm `--device` options
- Bump up `pylint`
- Remove circular import


smartsim/_core/config/config.py
```diff
@@ -29,11 +29,32 @@
...
+from shutil import which
 
 import psutil
 
 from ...error import SSConfigError
-from ..utils import expand_exe_path
+
+
+# Duplicating code to avoid circular import
+def expand_exe_path(exe: str) -> str:
```
MattToast: Could you list a test or a build command that was causing this circular import to occur? I naively tried just running the modules via `python -m smartsim._core. ...` and both imported fine, but that's definitely a non-standard usage. 

I gave a quick scan over where I would _think_ the circular import would occur and I couldn't find any thing obvious...but my mental interpreter has been known to miss these types of issues all the time hahaha.

My hope is that we should just need to be a bit more specific with the import path or delay the import or something to remove the duplicated code!

setup.cfg
```diff
@@ -44,14 +44,15 @@ keywords = scientific, ai, workflow, hpc, analysis
...
+    Programming Language :: Python :: 3.12
     License :: OSI Approved :: BSD License
     Intended Audience :: Science/Research
     Topic :: Scientific/Engineering
 
 [options]
 packages = find:
 include_package_data = True
-python_requires = >=3.10,<3.12
+python_requires = >=3.10,<3.13
```
MattToast, line 55: 🎉 🥳 🎉

docker/prod-cuda12/Dockerfile
```diff
@@ -42,23 +42,18 @@ RUN apt-get update \
...
 ENV CUDA_HOME="/usr/local/cuda/"
 ENV PATH="${PATH}:${CUDA_HOME}/bin"
 
-# Install machine-learning python packages consistent with RedisAI
-# Note: pytorch gets installed in the smart build step
-# This step will be deprecated in a future update
-RUN pip install tensorflow==2.15.0
-
 # Build ML Backends
 RUN smart build --device=cuda121
```
MattToast: Just checking, is this build command still accurate? From what I can tell, it should read
```
RUN smart build --device=cuda-12
```
(Admittedly, this is probably broken in the last release as well, and I'm not sure we want to squeeze this fix into a potentially unrelated PR, but that's neither here-nor-there lol)

.github/workflows/run_tests.yml
```diff
@@ -102,7 +103,7 @@ jobs:
       # on developments of the client are brought in.
       - name: Install SmartSim (with ML backends)
         run: |
-          python -m pip install git+https://github.com/CrayLabs/SmartRedis.git@develop#egg=smartredis
+          python -m pip install git+https://github.com/al-rigazzi/SmartRedis.git@python_312#egg=smartredis
```
MattToast: Just marking this as a reviewer comment to make sure this gets changed back when SR is merged!

.github/workflows/release.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Build Distributions
         run: python -m build .
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
MattToast, line 57: Good idea updating these, btw!!

smartsim/_core/generation/generator.py
```diff
@@ -28,7 +28,6 @@
 import shutil
 import typing as t
 from datetime import datetime
-from distutils import dir_util  # pylint: disable=deprecated-module
```
MattToast, line 31: IT'S FINALLY GONE!! THANK YOU!!!!!!!!!!

smartsim/ml/data.py
```diff
@@ -24,6 +24,8 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+# mypy: disable-error-code="redundant-cast"
+
```
MattToast, line 28: Optional: If you prefer, you should just need to remove the `t.cast(...)` calls in lines 442 and 454 as opposed to a module wide suppression by changing
```py
t.cast("SomeType", x)
```
to just
```py
x
```
as `t.cast` is just a no-op at runtime

smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
MattToast, line 1: I cannot fully explain why, but when I attempted to naively build on an `x86_64` MacOS I seem to be running into CMake configuration errors when attempting to build RedisAI. Build logs are below:

```bash
$ smart build -v --device cpu

[SmartSim] INFO Running SmartSim build process...
[SmartSim] INFO Checking requested versions...
[SmartSim] INFO Build Environment:
| Environment    | Values   |
|----------------|----------|
| CC             | gcc      |
| CXX            | g++      |
| CFLAGS         |          |
| CXXFLAGS       |          |
| MALLOC         | libc     |
| JOBS           | 1        |
| PYTHON_VERSION | 3.12.10  |
| PLATFORM       | darwin   |

[SmartSim] INFO Version Information:
| Packages   | Versions   |
|------------|------------|
| SMARTSIM   | 0.8.0      |
| REDIS      | 7.2.4      |
| REDISAI    | 1.2.7      |

[SmartSim] INFO ML Packages
| Package       | Version   |
|---------------|-----------|
| dlpack        | v0.5_RAI  |
| libtorch      | 2.2.2     |
| libtensorflow | 2.16.2    |
| onnxruntime   | 1.22.0    |
[SmartSim] INFO Building Redis version 7.2.4 from https://github.com/redis/redis.git
cd src && /usr/local/opt/make/libexec/gnubin/make all
make[1]: Entering directory '/private/tmp/SmartSim/smartsim/_core/.third-party/redis/src'

<build output omitted, from what I can tell redis builds fine!>

make[1]: Leaving directory '/private/tmp/SmartSim/smartsim/_core/.third-party/redis/src'
[SmartSim] INFO Redis build complete!
[SmartSim] INFO Building RedisAI and backends...
[SmartSim] INFO Retrieving package: dlpack v0.5_RAI
[SmartSim] INFO Retrieving package: libtorch 2.2.2
libtorch-macos-x86_64-2.2.2.zip: 100%|██████████████████████████| 172M/172M [00:04<00:00, 38.9MB/s]
[SmartSim] INFO Retrieving package: libtensorflow 2.16.2
libtensorflow-cpu-darwin-x86_64.tar.gz: 100%|███████████████████| 120M/120M [00:05<00:00, 23.8MB/s]
[SmartSim] INFO Retrieving package: onnxruntime 1.22.0
onnxruntime-osx-x86_64-1.22.0.tgz: 100%|██████████████████████| 26.6M/26.6M [00:01<00:00, 16.3MB/s]
[SmartSim] INFO Configuring CMake Build
cmake -DBUILD_TF=ON -DBUILD_ORT=ON -DBUILD_TORCH=ON -DBUILD_TFLITE=OFF -DDEPS_PATH=/private/tmp/SmartSim/smartsim/_core/build/RedisAI/package -DDEVICE=cpu -DINSTALL_PATH=/private/tmp/SmartSim/smartsim/_core/lib -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ /private/tmp/SmartSim/smartsim/_core/build/RedisAI/src
CMake Error at CMakeLists.txt:1 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
  File "/private/tmp/venv/bin/smart", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/private/tmp/SmartSim/smartsim/_core/_cli/__main__.py", line 43, in main
    return smart_cli.execute(sys.argv)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/SmartSim/smartsim/_core/_cli/cli.py", line 88, in execute
    return menu_item.handler(args, unparsed_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/SmartSim/smartsim/_core/_cli/build.py", line 314, in execute
    build_redis_ai(current_platform, mlpackages, build_env, verbose)
  File "/private/tmp/SmartSim/smartsim/_core/_cli/build.py", line 136, in build_redis_ai
    rai_builder.build()
  File "/private/tmp/SmartSim/smartsim/_core/_install/redisaiBuilder.py", line 175, in build
    self.run_command(cmake_command, self.build_path)
  File "/private/tmp/SmartSim/smartsim/_core/_install/redisaiBuilder.py", line 249, in run_command
    raise RedisAIBuildError(
smartsim._core._install.redisaiBuilder.RedisAIBuildError: RedisAI build failed during command: \
    cmake \
        -DBUILD_TF=ON \
        -DBUILD_ORT=ON \
        -DBUILD_TORCH=ON \
        -DBUILD_TFLITE=OFF \
        -DDEPS_PATH=/private/tmp/SmartSim/smartsim/_core/build/RedisAI/package \
        -DDEVICE=cpu -DINSTALL_PATH=/private/tmp/SmartSim/smartsim/_core/lib \
        -DCMAKE_C_COMPILER=gcc \
        -DCMAKE_CXX_COMPILER=g++ /private/tmp/SmartSim/smartsim/_core/build/RedisAI/src
```

In theory, my CMake should be at the right version??
```bash
$ cmake --version
cmake version 4.0.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

If needed, I'd be willing to spend more time looking into this since I'm not sure if the SmartSim maintenance team has an intel mac or equivalent action runner readily available! I also know that this platform is significantly less utilized than Linux or MacOS+ARM and it is only becoming less popular, so there may be an argument for just dropping the platform altogether.

smartsim/settings/base.py
```diff
@@ -451,7 +451,7 @@ def add_exe_args(self, args: t.Union[str, t.List[str]]) -> None:
         args = self._build_exe_args(args)
         self._exe_args.extend(args)
 
-    def set(
+    def set_option(
```
MattToast: Not wrong, just curious: any reason for the method rename? I'm little concerned about this just because it is a public facing method.

MattToast: Also, if we do stick with this name change, can you update the doc string here? It still refers to this method as `rs.set(...)`

doc/changelog.md
```diff
@@ -21,6 +22,9 @@ Description
 
 Detailed Notes
 
+- Python 3.12 is now supported. TensorFlow 2.16.2 and PyTorch 2.7.1 library files are installed as part of `smart build` process when available. On Mac, ONNX runtime 1.22.0 is now installed, together with
+ONNX 1.16.
```
MattToast: Incredibly pedantic style nit: your choice to line wrap at column 204 for exactly 2 words confuses and slightly frightens me hahahaha

.pylintrc
```diff
@@ -71,6 +71,7 @@ disable=logging-fstring-interpolation,
         fixme,
         broad-exception-raised,
         duplicate-code, # TODO: remove this after fixing duplicate code
+        too-many-positional-arguments,
```
MattToast, line 74: Optional: if you'd rather change the number of allowed positional arguments rather than silencing the warning altogether, you could add a `max-positional-arguments=x` setting under the `[DESIGN]` section.

That said, I think we silenced the "too many arguments" warning by just turning it off entirely as well, so this actually may be more consistent.

MattToast (changes requested):
Looks great! I have a couple of very minor notes, a couple of very minor optional change requests, and a couple of places asking for some more information before I approve, but otherwise looks about ready to go!

smartsim/_core/config/config.py
al-rigazzi: Yeah, this was super strange - on *some* GH workflows, our `slurm` module would cause the circular import.

```
************* Module smartsim.wlm.slurm
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim._core.utils.shell -> smartsim._core.utils.helpers) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.log) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.entity -> smartsim.entity.model -> smartsim.log) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.entity -> smartsim.entity.ensemble -> smartsim.entity.model -> smartsim.log) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.helpers) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.entity -> smartsim.entity.model -> smartsim.settings.base -> smartsim._core.utils.helpers) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.entity -> smartsim.entity.dbnode) (cyclic-import)
smartsim/wlm/slurm.py:1:0: R0401: Cyclic import (smartsim._core.config -> smartsim._core.config.config -> smartsim._core.utils -> smartsim._core.utils.redis -> smartsim.entity -> smartsim.entity.model -> smartsim.settings.base -> smartsim.settings.containers -> smartsim.log) (cyclic-import)
```

.github/workflows/release.yml
al-rigazzi: Good and bad, as it came with some more work...

smartsim/ml/data.py
al-rigazzi: I know, but if I do _that_, `pylint` complains with some versions of Python, while if I don't, it complains with others...

smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
al-rigazzi: Interesting, tht must be caused by a dependency that is not updated. Can you try to 
```export CMAKE_POLICY_VERSION_MINIMUM=3.5```
and then `smart build` again?

File: smartsim/settings/base.py (hunk 8 above)
Comment (al-rigazzi) [reply smartsim/settings/base.py]: Yeah, so, as `set` has a meaning in Python, the fact that we are calling this member `set(...)` *and* we are creating an empty set a few lines before leads `pylint` to believe we are using the function before defining it.

File: docker/prod-cuda12/Dockerfile (hunk 3 above)
Comment (al-rigazzi) [reply docker/prod-cuda12/Dockerfile]: I'll start with that, but I have little hope it is still a working recipe.

File: doc/changelog.md (hunk 9 above)
Comment (al-rigazzi) [reply doc/changelog.md]: LOL, must have been some series of reformattings...

File: .pylintrc (hunk 10 above)
Comment (al-rigazzi) [reply .pylintrc:74]: I believe that would require some extensive understanding of what we would like to see.

File: docker/prod-cuda12/Dockerfile (hunk 3 above)
Comment (al-rigazzi) [reply docker/prod-cuda12/Dockerfile]: It will be the subject of a pre-release activity.

File: smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
Comment (al-rigazzi) [reply smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json:1]: And, as you correctly predicted, there is no way for us to test MacOS x86_64 anymore, so, if you want to test this a bit further, it is highly appreciated.

Comment (al-rigazzi) [reply smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json:1]: OK, I take that back, we have `macos-13` which is still Intel.

Comment (al-rigazzi) [reply smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json:1]: Just one question, are you using SmartRedis from the branch you reviewed the PR for?

File: smartsim/settings/base.py (hunk 8 above)
Comment (al-rigazzi) [reply smartsim/settings/base.py]: I'm open to a different solution

File: smartsim/_core/config/config.py (hunk 1 above)
Comment (MattToast) [reply smartsim/_core/config/config.py]: Oh! Is `pylint` complaining about it...? I'm less familiar with that rule since it was added in v3, but iirc a delayed import would still get flagged :(

Comment (MattToast) [reply smartsim/_core/config/config.py]: I'll poke around with this a bit today and hopefully find a solution that doesn't require creation/merging of namespaces, but if I cannot find a good solution by EOD I'll say this is "good enough"

File: docker/prod-cuda12/Dockerfile (hunk 3 above)
Comment (MattToast) [reply docker/prod-cuda12/Dockerfile]: Sounds good! In that case I'll mark this as resolved for this PR! If I'm really good I'll even remember to write up an issue for it hahahaha

File: smartsim/ml/data.py (hunk 7 above)
Comment (MattToast) [reply smartsim/ml/data.py:28]: Well.... that's just delightful lol.... I'll add this to my list of things to kick today, but again, if I cannot find a good solution by EOD, I'll call this "good enough"

File: smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
Comment (MattToast) [reply smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json:1]: ^^ Yeah I was using the `python_312` branch off of your fork. I'm going to try exporting that env var now!!

File: smartsim/settings/base.py (hunk 8 above)
Comment (MattToast) [reply smartsim/settings/base.py]: Let me see if I can get the error to reproduce! I know pylint would flag
```py
set = 123  # pylint should raise: redefined-builtin
```
but I'm a little surprised that it would flag a method name on a class, just because that is in its own namespace

.pylintrc
MattToast: Yeah, on second thought, I would rather we solve both the "too many args" and "too many positional args" issues at once, an I could see that getting hairy fast. Probably makes sense to suppress them in the same way for the time being!

smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
MattToast: Exporting `CMAKE_POLICY_VERSION_MINIMUM` seems to have done the trick!

MattToast: Thoughts on adding a `"CMAKE_POLICY_VERSION_MINIMUM": "3.5",` entry to [this dict](
https://github.com/al-rigazzi/SmartSim/blob/6945f92278142fe4b27c7188e80a45c2f1f1e3b0/smartsim/_core/_install/redisaiBuilder.py#L255-L274) in the `RedisAIBuilder` class so that users do not need to manually export this themselves? By adding one there manually I was able to unset the env var and everything built as expected!

There is also an argument that this could be added to the `BuildEnv` class and/or there should be some way to allow users to overwrite it, but I'll leave that to your discretion

smartsim/settings/base.py
MattToast: OHHHHH!! I see the issue at line 159 now!

Okay, so using this as an example:
```py
class Problematic:
    reserved_run_args = set()  # type: set[str]
    def set(
        self, arg: str, value: str | None = None, condition: bool = True
    ) -> None: print(self, arg, value, condition)
```

I can think of three ways to get this past the linter:

<details>

<summary>Option 1</summary>

We can explicitly call `set` off of the `builtins` module like so
```py
import builtins

class Problematic:
    reserved_run_args = builtins.set()  # type: set[str]
    def set(
        self, arg: str, value: str | None = None, condition: bool = True
    ) -> None: ...
```

Pros: Minimum viable change to make it past the lint check, not an API break

Cons: Doesn't properly convey the intent of `reserved_run_args`

</details>

<details>

<summary>Option 2</summary>

Iirc we never actually want the user to modify the `reserved_run_args`, so it may actually make sense for us to change this to a `frozenset` instead. We could also add some additional typing information here as well

```py
class Problematic:
    reserved_run_args: t.ClassVar[frozenset[str]] = frozenset()
    def set(
        self, arg: str, value: str | None = None, condition: bool = True
    ) -> None: ...
```

Pros: Adds more static type information to be "more correct" during static analysis, better expresses the intent behind `reserved_run_args`

Cons: Because we allowed for mutation of `reserved_run_args` in the past AND it is publicly accessable, this is _technically_ an API break :(

</details>

<details>

<summary>Option 3</summary>

I would need to check, but I'm pretty sure we only really need this variable to support `__contains__`.  If that's the case, we could change the type to a more generic `t.Collection` and supply a list instead

```py
class Problematic:
    reserved_run_args: t.ClassVar[t.Collection[str]] = []
    def set(
        self, arg: str, value: str | None = None, condition: bool = True
    ) -> None: ...
```


Pros: Adds even MORE static type information to better express intent, allows subclasses to be a hair more generic

Cons: Because `reserved_run_args` is public and we changed the type of collection, this is still _technically_ and API break (albeit slightly less severe), allowing ANY collection could allow subclasses to accidentally choose an inefficient contain for element look ups.

</details>

I would be fine with any of the above, and given how public facing `.set` is compared to `.reserved_run_args`, if we are going to make a break here, I think I would prefer breaking the latter attr (though I could be talked out of this). I'll leave further impl details up to your discretion!

smartsim/_core/config/config.py
MattToast: Okay! I think I found a "solution" to get past the linter!!

 It seems like this cycle specifically is already a known issue within SmartSim based on existing linter suppressions, but if we update this import path to import this function directly from `smartsim._core.utils.helpers`

```py
from ..utils.helpers import expand_exe_path
# NOTE:     ^^^^^^^^  Full path now used
```

AND we let `pylint` know explicitly that we are doing a bad thing at [line 229 in `smartsim._core.utils.helpers`](https://github.com/al-rigazzi/SmartSim/blob/6945f92278142fe4b27c7188e80a45c2f1f1e3b0/smartsim/_core/utils/helpers.py#L228-L230)

```py
def redis_install_base(backends_path: t.Optional[str] = None) -> Path:
    # NOTE:                       Update here!!!  vvvvvvvvvvvvvv
    # pylint: disable-next=import-outside-toplevel,cyclic-import
    from ..._core.config import CONFIG
```

That should suppress the error without the duplication for the time being


That being said, in the future, when we have time, I think we should address some of these cycles. I'm fairly certain most of any cyclic imports are coming from `CONFIG` being defined in `smartsim._core.config` instead of `smartsim._core.config.config` and it being used pretty globally throughout the codebase

smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
al-rigazzi: Yeah, it's strange that in the GH actions I do not get that error. Maybe they use a different CMake version. OK, I'll add it!

smartsim/settings/base.py
al-rigazzi: I'm going with `frozenset` and wil let you know

.github/workflows/run_tests.yml
al-rigazzi: 👁️ YOU WILL 👁️

docker/prod-cuda12/Dockerfile
al-rigazzi: 👁️ YOU WILL 👁️

smartsim/settings/base.py
al-rigazzi: It works and I'm happier now.

smartsim/ml/data.py
MattToast: Can't find a great alternative for this suppression, but this seems like a reasonable enough suppression to make (and I would actually argue that I think the cast is necessary if this code were to be multi-threaded at some point since the types are marked `Optional` lol).

Marking this as resolved!!

smartsim/_core/_install/configs/mlpackages/DarwinX64CPU.json
MattToast: I must have somehow managed to bork my CMake config... BUT I'm glad my ineptitude can help make `smart build` more intuitive for even the least functioning computers out there, hahahahaha

smartsim/settings/base.py
```diff
@@ -156,7 +156,7 @@ def env_vars(self, value: t.Dict[str, t.Optional[str]]) -> None:
         self._env_vars = copy.deepcopy(value)
 
     # To be overwritten by subclasses. Set of reserved args a user cannot change
-    reserved_run_args = set()  # type: set[str]
+    reserved_run_args: frozenset[str] = frozenset()
```
MattToast: Without wanting to add too much additional work onto this PR, can we mark this as a `ClassVar`? Or do we actually treat it like an instance variable?

```suggestion
    reserved_run_args: t.ClassVar[frozenset[str]] = frozenset()
```

If this leads to additional typing errors, it might make sense to address them in a later PR, seeing as this is now, at best, tangential to upgrading to Python 3.12.

MattToast (approved):
Three last comments!

- [x] I think I found a way to remove the `expand_exe_path` code duplication by suppressing the `pylint` circular import warning for now ([see above](https://github.com/CrayLabs/SmartSim/pull/785#discussion_r2143695602))
- [x] I have one last minor typing nit regarding `reserved_run_args` ([see below](https://github.com/CrayLabs/SmartSim/pull/785#discussion_r2145755166))
- [x] Obligatory reminder to change the SmartRedis repo that we install from when the [3.12 PR](https://github.com/CrayLabs/SmartRedis/pull/527) is merged ([see above](https://github.com/CrayLabs/SmartSim/pull/785#discussion_r2138525650))

Otherwise, LGTM!!

smartsim/_core/config/config.py
al-rigazzi: Agree

smartsim/settings/base.py
al-rigazzi: Welcome to the co-authoring team!

.github/workflows/run_tests.yml
al-rigazzi: Done

PR #786, opened 2025-06-18
Bypass executable check when creating RunSettings

ashao:
In some instances, the strong check that the executable can be found by the driver script when creating RunSettings is too restrictive. This can occur in cases where the binary does not exist on the node where the driver script is being launched (e.g. remote execution) or for binaries which need to be found after setting environment variables.


al-rigazzi (changes requested):
Could you please add a test -- or even parameterize an existing one, so that we can check for correct behavior when `skip_exe_check==True`?
Because otherwise I agree that this is too restrictive and this can be merged pretty soon!

PR #788, opened 2025-07-02
Add functionality to turn off monitoring status of Models launched

rickybalin:
This PR adds the ability to turn off monitoring of Models launched with `experiment.start()` by modifying the Controller and JobManager classes. With this update, `experiment.start()` takes an additional boolean argument determining whether to monitor the particular model or not. The argument is set to `True` by default, so it should not require any changes for the default behavior of monitoring all Models launched.

The PR also adds a method to `PalsMpiexecSettings` called `set_launcher_args()` to be able to add any launcher arguments to the mpiexec command.


rickybalin:
@ashao, could I get some help getting this PR reviewed and merged? We need it for some functionality to stand up a "persistent" colocated DB, to which applications can connect and disconnect during a workflow. Thanks! 

Please let me know if there are any issues.

smartsim/experiment.py
```diff
@@ -205,11 +206,16 @@ def start(
         that all jobs launched by this experiment will be killed, and the
         zombie processes will need to be manually killed.
 
+        If `monitor=True`, all the jobs being started will be monitored
+        by the Controller. If `monitor=True`, the jobs will not be
```
MattToast:
```suggestion
        by the Controller. If `monitor=False`, the jobs will not be
```

rickybalin: Thanks for the catch!

ashao (approved):

PR #789, opened 2025-07-28
Remove telemetry, SmartDashboard, and indirect entrypoint functionality

al-rigazzi:
## Overview

This PR removes telemetry collection, SmartDashboard functionality, and indirect entrypoint capabilities from SmartSim. Additionally, it significantly improves the codebase's maintainability by centralizing directory path management through the CONFIG system and eliminating hardcoded `.smartsim` directory references throughout the codebase.

### 🗑️ Removed Features
- **Telemetry Collection**: Removed all telemetry data collection and transmission functionality
- **SmartDashboard**: Removed dashboard visualization and monitoring capabilities  
- **Indirect Entrypoints**: Removed support for indirect application entrypoints

### 🏗️ Configuration System Improvements
- **Centralized Directory Management**: Enhanced the CONFIG system with comprehensive directory path properties
- **Hierarchical Structure**: Implemented proper directory hierarchy for better organization
- **Eliminated Hardcoded Paths**: Replaced all hardcoded `.smartsim` directory references with CONFIG properties

### 📁 New CONFIG Properties
```python
CONFIG.smartsim_base_dir        # Returns ".smartsim"
CONFIG.dragon_default_subdir   # Returns ".smartsim/dragon"  
CONFIG.dragon_logs_subdir       # Returns ".smartsim/dragon/logs"
CONFIG.metadata_subdir          # Returns ".smartsim/metadata"
```

### 📂 Directory Structure
The new hierarchical structure provides better organization:
```
.smartsim/
├── dragon/
│   └── logs/           # Dragon launcher logs
├── metadata/           # Experiment metadata
│   └── run_{timestamp}/
│       ├── model/
│       ├── ensemble/
│       └── database/
└── keys/              # Security keys
```

### Core Configuration
- `smartsim/_core/config/config.py` - Enhanced with new directory properties and hierarchical structure

### Core Functionality  
- `smartsim/_core/utils/serialize.py` - Updated to use CONFIG.metadata_subdir
- `smartsim/_core/control/manifest.py` - Updated to use CONFIG.metadata_subdir

### Test Files (15+ files updated)
**High Priority:**
- `tests/test_symlinking.py` - Updated hardcoded paths to use CONFIG properties
- `tests/test_manifest_metadata_directories.py` - Updated to use CONFIG.metadata_subdir
- `tests/test_metadata_integration.py` - Updated to use CONFIG properties

**Medium Priority:**
- `tests/test_controller.py` - Updated to use CONFIG.metadata_subdir
- `tests/test_output_files.py` - Updated comments to reference CONFIG

**Dragon Test Files:**
- `tests/test_dragon_step.py` - Updated to use CONFIG.dragon_logs_subdir
- `tests/test_dragon_launcher.py` - Updated to use CONFIG.dragon_logs_subdir  
- `tests/test_dragon_client.py` - Updated to use CONFIG.dragon_logs_subdir
- `tests/test_dragon_run_policy.py` - Updated to use CONFIG.dragon_logs_subdir

### 🔧 Improved Maintainability
- **Single Source of Truth**: All directory paths now managed through CONFIG
- **Easy Customization**: Directory structure can be modified by changing CONFIG properties
- **Consistent Architecture**: Eliminates path duplication and inconsistencies

### 🧪 Better Testing
- **Configurable Paths**: Tests can easily modify directory paths through CONFIG
- **Consistent Test Environment**: All tests use the same path management system
- **Future-Proof**: New directory requirements can be added as CONFIG properties

### 🏗️ Enhanced Modularity
- **Hierarchical Organization**: Dragon components properly organized under parent directory
- **Logical Structure**: Related functionality grouped in appropriate subdirectories
- **Extensible Design**: Easy to add new directory categories

### CONFIG System Architecture
The new CONFIG properties follow a hierarchical dependency model:
- `smartsim_base_dir` is the foundation (`.smartsim`)
- `dragon_default_subdir` builds on base (`{base}/dragon`)
- `dragon_logs_subdir` builds on dragon (`{dragon}/logs`)
- `metadata_subdir` builds on base (`{base}/metadata`)

This ensures that changing the base directory automatically propagates to all subdirectories, and changing the dragon directory affects the dragon logs directory.

### Path Management Implementation
All directory references now use f-string formatting with CONFIG properties:
```python
# Before (hardcoded)
metadata_dir = pathlib.Path(temp_dir) / ".smartsim" / "metadata"

# After (CONFIG-based)
metadata_dir = pathlib.Path(temp_dir) / CONFIG.metadata_subdir
```

### Test Improvements
- **Reverted Parameterization**: Restored individual entity testing in `test_batch_symlink` and `test_symlink` for better test clarity
- **Consistent Imports**: Added CONFIG imports to all relevant test files
- **Path Flexibility**: Tests can now easily modify directory structures for different scenarios

## Testing

- ✅ All existing tests pass with CONFIG property updates
- ✅ Directory hierarchy correctly propagates changes (base → dragon → logs)
- ✅ Metadata integration tests validate proper directory structure
- ✅ Dragon launcher tests work with new nested log directory structure
- ✅ Configuration tests verify expected property values
- ✅ Symlinking tests work with CONFIG-based paths
- ✅ Manifest metadata tests validate hierarchical structure

## Migration Notes

**For Developers:**
- Use `CONFIG.metadata_subdir` instead of hardcoding ".smartsim/metadata"
- Use `CONFIG.dragon_logs_subdir` instead of hardcoding ".smartsim/logs"  
- Use `CONFIG.smartsim_base_dir` for base directory references
- All directory properties automatically adapt to base directory changes

**For Users:**
- No breaking changes to public APIs
- Directory structure remains the same by default
- Enhanced consistency in file organization

## Backward Compatibility

- All existing directory paths remain unchanged by default
- Tests continue to work without modification for end users
- Public APIs maintain compatibility
- File organization is improved without breaking existing workflows

## Summary

This PR significantly improves SmartSim's codebase quality by:
1. **Removing deprecated functionality**: Telemetry, SmartDashboard, and indirect entrypoints
2. **Establishing robust directory management**: Centralized CONFIG-based path system
3. **Eliminating technical debt**: Removed 20+ hardcoded path references across 15+ files
4. **Improving code organization**: Hierarchical directory structure with proper dependency relationships
5. **Enhancing maintainability**: Single source of truth for all directory paths
6. **Future-proofing the codebase**: Easy to extend with new directory requirements

The changes result in a cleaner, more maintainable codebase with improved consistency, better separation of concerns, and a solid foundation for future development.


al-rigazzi (commented):
In general, after some assisted modifications, I believe the structure now looks reasonable. Note that the dashboard has already been deprecated in its own repo.

tests/test_controller_metadata_usage.py
```diff
@@ -0,0 +1,169 @@
...
+    """Test that the Controller properly uses metadata directories"""
+
+    def setup_method(self):
+        """Set up test fixtures"""
+        self.temp_dir = tempfile.mkdtemp()
+        self.controller = Controller("local")
+
+    def teardown_method(self):
+        """Clean up test fixtures"""
+        import shutil
```
MattToast: Nit: we are importing `shutil` outside of the global scope? Is there a reason for this?

smartsim/_core/_cli/cli.py
```diff
@@ -62,7 +62,9 @@ def __init__(self, menu: t.List[MenuItemConfig]) -> None:
         )
 
         self.register_menu_items(menu)
-        self.register_menu_items([plugin() for plugin in plugins])
+        # Register plugin menu items (currently empty since all plugins were removed)
+        plugin_items: t.List[MenuItemConfig] = [plugin() for plugin in plugins]
+        self.register_menu_items(plugin_items)
```
MattToast, line 67: Nit: Now that we require Python >= 3.9, prefer `list[T]` over the (now deprecated) `typing.List[T]`.

```suggestion
        # Register plugin menu items (currently empty since all plugins were removed)
        plugin_items: list[MenuItemConfig] = [plugin() for plugin in plugins]
        self.register_menu_items(plugin_items)
```

Or, better yet, drop the type-hint entirely and let the checker figure it out!

```suggestion
        # Register plugin menu items (currently empty since all plugins were removed)
        plugin_items = [plugin() for plugin in plugins]
        self.register_menu_items(plugin_items)
```

smartsim/_core/control/job.py
```diff
@@ -35,8 +35,8 @@
 
 @dataclass(frozen=True)
 class _JobKey:
-    """A helper class for creating unique lookup keys within the telemetry
-    monitor. These keys are not guaranteed to be unique across experiments,
+    """A helper class for creating unique lookup keys within a job manager.
+    These keys are not guaranteed to be unique across experiments,
     only within an experiment (due to process ID re-use by the OS)"""
```
MattToast: If you want to do some more clean-up work (and iirc), `_JobKey` is only ever used as a return value for `JobEntity.key`, and from a fairly naive grep, I could find any references to that attribute.

You should be free to remove both this class and `JobEntity.key`!

smartsim/_core/control/job.py
```diff
@@ -147,22 +122,22 @@ def _map_standard_metadata(
...
         )
         entity.step_id = str(metadata.get("step_id") or "")
         entity.task_id = str(metadata.get("task_id") or "")
         entity.timestamp = int(entity_dict.get("timestamp", "0"))
         entity.path = str(exp_dir)
-        entity.status_dir = str(status_dir)
+        entity.metadata_dir = str(metadata_dir)
 
     @classmethod
     def from_manifest(
```
MattToast: Similar note here: `JobEntity.from_manifest` was an alternate constructor for making `JobEntity` instances from information provided in the manifest file for the Telemetry Monitor. Since the TM no longer exists, this constructor appears to be unused!

MattToast: And then by removing this constructor, you can remove both `JobEntity._map_standard_metadata` and `JobEntity._map_db_metadata`!

MattToast: And actually!! It doesn't look like we construct a `JobEntity` anywhere in the code base anymore. Naively removing `JobEntity` from the union types in `Job.__init__` and `JobManager.add_job` does not upset MyPy. Can we just remove `JobEntity` entirely??

smartsim/_core/config/config.py
```diff
@@ -272,24 +272,20 @@ def test_mpi(self) -> bool:  # pragma: no cover
...
-    def telemetry_subdir(self) -> str:
-        return ".smartsim/telemetry"
+    def dragon_logs_subdir(self) -> str:
+        return f"{self.dragon_default_subdir}/logs"
 
     @property
-    def dragon_default_subdir(self) -> str:
-        return ".smartsim/dragon"
+    def metadata_subdir(self) -> str:
+        return f"{self.smartsim_base_dir}/metadata"
```
MattToast: Nit: consider making these return `pathlib.Path`s just so that the type-checker can help catch "stupid mistakes" (e.g. passing a path to an arbitrary function that expects a string that is supposed to represent something else).

If, for whatever reason, that requires work too much work to fit in this PR and the return type needs to remain a string, consider replacing the manual path concatenation via f-str with `os.path.join` for better portability (even though windows isn't directly supported hahaha)

smartsim/_core/control/manifest.py
```diff
@@ -256,6 +249,9 @@ class LaunchedManifestBuilder(t.Generic[_T]):
     exp_path: str
     launcher_name: str
     run_id: str = field(default_factory=_helpers.create_short_id_str)
+    _launch_timestamp: str = field(
+        default_factory=lambda: str(int(time.time() * 1000)), init=False
+    )
```
MattToast: from what I can tell `run_id` and `_launch_timestamp` serve a very similar purpose (to determine which `SmartSimEntity`s were launched in which `Experiment.start` calls).

It also looks like `run_id` is only being dumped in the `manifest.json` file, but is not read anywhere. Is it strictly needed anymore if the Telemetry Monitor is no longer using that as input? Are we keeping it around just in case we need to bring something back?

smartsim/_core/control/manifest.py
```diff
@@ -197,17 +198,9 @@ class _LaunchedManifestMetadata(t.NamedTuple):
...
-        return _format_exp_telemetry_path(self.exp_path)
-
-    @property
-    def run_telemetry_subdirectory(self) -> pathlib.Path:
-        return _format_run_telemetry_path(self.exp_path, self.exp_name, self.run_id)
-
     @property
     def manifest_file_path(self) -> pathlib.Path:
-        return self.exp_telemetry_subdirectory / _serialize.MANIFEST_FILENAME
+        return pathlib.Path(self.exp_path) / _serialize.MANIFEST_FILENAME
```
MattToast: Just sanity checking that this was intended:

This change will make the `manifest.json` file "public" being at the root of the experiment directory. 

In the past we wanted to keep this file "private" by hiding it in a hidden `{exp_path}/.smartsim/telemetry` directory (presumably because we did not want users to modify it, thus invalidating the dashboard output).

Now that the dashboard is not supported, we don't particularly have a reason to hide the `manifest.json` file, outside of _maybe_ some lingering correctness errors surrounding multiple experiments in the same driver script with the same `exp_path`.

Just wanted to make sure that it was intended that it end up in `{exp_path}` and not `{exp_path}/.smartsim/metadata` or similar!

MattToast: Alternatively, if we no longer support the dashboard, there may be an argument to simply NOT generate the `manifest.json` file at all? Thoughts??

MattToast (changes requested):
Still making my way through the rest of this PR, but I have some initial thoughts I wanted to throw your way instead of stalling any longer 😅 

Feel free to lmk what you think!!

smartsim/_core/control/manifest.py
al-rigazzi: I know. I was going to ask your opinion on this. I don't think we ever advertised the existence of `manifest.json` to users, so we may just drop it?

smartsim/_core/config/config.py
al-rigazzi: Done

smartsim/_core/control/job.py
al-rigazzi: Done

smartsim/_core/control/job.py
al-rigazzi: Done

smartsim/_core/control/manifest.py
MattToast: I definitely am going to elect to at least keep it "private" in `{exp_path}/.smartsim/...` just because of those lingering correctness issues I mentioned. E.g. this experiment
```py

exp_1 = Experiment("my-exp-1", exp_path="/tmp/some-dir")
rs_1 = exp_1.create_run_settings("echo", exe_args=["Hello", "from", "Experiment", "1"])
m_1 = exp.create_model("hello-1", rs_1)

exp_2 = Experiment("my-exp-2", exp_path="/tmp/some-dir")
rs_2 = exp_1.create_run_settings("echo", exe_args=["Hello", "from", "Experiment", "2"])
m_2 = exp.create_model("hello-2", rs_2)

exp_1.start(m_1)
exp_2.start(m_2)
```
will generate a SINGLE `/tmp/some-dir/manifest.json` file and it will report that BOTH models `m_1` and `m_2` were launched by Experiment 1. This can be particularly confusing if `exp_1` and `exp_2` were actually using _different launchers_. AND this on top of known limitations with the `manifest.json` fromat we chose (MPMD jobs for e.g. cannot be properly JSON-ified)

That said, personally, I would be okay with just dropping it entirely!! As far as I can remember, it was only ever meant as a stop-gap solution for the Telemetry Monitor until we designed something more sophisticated. It seems unlikely that we will have the manpower required to address these correctness issues, so if nothing is using the file internally AND we think that we should hide the file from end users, I see very little reason to make it in the first place!

al-rigazzi: Well, let's delete them. They were a good stopgap solution, btw.

smartsim/_core/control/manifest.py
al-rigazzi: Nah, deleted `run_id`

smartsim/_core/launcher/dragon/dragonBackend.py
```diff
@@ -177,12 +176,7 @@ def __init__(self, pid: int) -> None:
         """Whether the server frontend should shut down when the backend does"""
         self._shutdown_initiation_time: t.Optional[float] = None
         """The time at which the server initiated shutdown"""
-        smartsim_config = get_config()
-        self._cooldown_period = (
-            smartsim_config.telemetry_frequency * 2 + 5
-            if smartsim_config.telemetry_enabled
-            else 5
-        )
+        self._cooldown_period = 5
```
MattToast, line 179: Nit, not to be addressed in this PR: Do we not have an environment variable or something to let users' configure the cool down period? Should be add this as a field to `Config`, or as a parameter to `__init__` or something?

smartsim/_core/launcher/step/step.py
```diff
@@ -27,20 +27,17 @@
...
 import typing as t
 from os import makedirs
 
-from smartsim._core.config import CONFIG
-from smartsim.error.errors import SmartSimError, UnproxyableStepError
+from smartsim.error.errors import SmartSimError
 
 from ....log import get_logger
 from ....settings.base import RunSettings, SettingsBase
-from ...utils.helpers import encode_cmd, get_base_36_repr
```
MattToast, line 43: Nit: If you want to clean up _EVEN MORE_, it looks like `encode_cmd` and `decode_cmd` defined in `smartsim._core.utils.helpers`, but not called anywhere outside of test files

tests/test_controller.py
```diff
@@ -69,7 +70,7 @@ def test_controller_batch_step_creation_preserves_entity_order(collection, monke
     )
     entity_names = [x.name for x in collection.entities]
     assert len(entity_names) == len(set(entity_names))
-    _, steps = controller._create_batch_job_step(
-        collection, pathlib.Path("mock/exp/path")
-    )
+    # Create a metadata directory for the test
+    metadata_dir = pathlib.Path("/tmp") / CONFIG.metadata_subdir
+    _, steps = controller._create_batch_job_step(collection, metadata_dir)
```
MattToast, line 77: Nit: should we consider making this a path based off of the `test_dir` fixture, just so that any artifacts that are created during the test are easier to be found/investigated in the case of a failure?

tests/test_output_files.py
```diff
@@ -106,10 +106,12 @@ def test_mutated_model_output(test_dir):
 def test_get_output_files_with_create_job_step(test_dir):
     """Testing output files through _create_job_step"""
     exp_dir = pathlib.Path(test_dir)
-    status_dir = exp_dir / CONFIG.telemetry_subdir / model.type
-    step = controller._create_job_step(model, status_dir)
-    expected_out_path = status_dir / model.name / (model.name + ".out")
-    expected_err_path = status_dir / model.name / (model.name + ".err")
+    model.path = test_dir
```
MattToast: Nit: Not really a fan of modifying attributes of globally available instances of `Model`s just because it means that this test CAN leak state in future. If we need to set this path, can we make this `Model` a fixture, or use `monkeypatch.setattr` or something?

tests/test_symlinking.py
```diff
@@ -65,7 +65,7 @@ def test_symlink(test_dir, entity):
     """Test symlinking historical output files"""
     entity.path = test_dir
     if entity.type == Ensemble:
-        for member in ens.models:
+        for member in entity.models:
```
MattToast, line 89: Nice catch!!

tests/test_symlinking.py
```diff
@@ -100,19 +104,51 @@ def test_batch_symlink(entity, test_dir):
...
-    for step in substeps:
-        slurm_controller.symlink_output_files(step, entity)
-        assert pathlib.Path(entity.path, f"{entity.name}.out").is_symlink()
-        assert pathlib.Path(entity.path, f"{entity.name}.err").is_symlink()
-        assert os.readlink(pathlib.Path(entity.path, f"{entity.name}.out")) == str(
-            status_dir / entity.name / step.entity_name / (step.entity_name + ".out")
+    # For entities with sub-entities (like Orchestrator), set their paths too
+    if hasattr(entity, "entities"):
+        for sub_entity in entity.entities:
+            sub_entity.path = test_dir
```
MattToast, line 133: Ditto previous complaint about this _technically_ leaking test state by writing to global variables (I know that we do this elsewhere in this module, e.g. line 106, don't worry it bothers me there too hahaha). Thoughts on making these entities fixtures and/or using `monkeypatch.setattr`??

tests/test_symlinking.py
```diff
@@ -100,19 +104,51 @@ def test_batch_symlink(entity, test_dir):
...
+        expected_err = pathlib.Path(substep.meta["metadata_dir"]) / (
+            substep.entity_name + ".err"
         )
 
+        assert os.readlink(symlink_out) == str(expected_out)
+        assert os.readlink(symlink_err) == str(expected_err)
+    else:
+        # For _AnonymousBatchJob (single model)
+        substep = substeps[0]
+        slurm_controller.symlink_output_files(substep, entity)
```
MattToast: Just sanity checking, is this block reachable? In theory an `_AnonymousBatchJob` has an `entities` attribute that is set to a list of length 1 containing the single model at `__init__` time. If my mental compiler is correct, I think this block is unreachable and free to remove!

tests/test_symlinking.py
```diff
@@ -100,19 +104,51 @@ def test_batch_symlink(entity, test_dir):
...
+            sub_entity.path = test_dir
+
+    # Create metadata_dir to simulate consistent metadata structure
+    metadata_dir = exp_dir / CONFIG.metadata_subdir
+    batch_step, substeps = slurm_controller._create_batch_job_step(entity, metadata_dir)
+
+    # For batch entities, we need to call symlink_output_files correctly
+    # Based on how the controller does it, we should pass the individual entities
+    if hasattr(entity, "entities") and len(substeps) > 0:
+        # Just test the first substep and entity pair
```
MattToast, line 142: Not wrong, just curious: Any reason we only test the first substep? Could we test them all??

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,348 @@
...
+            ), f"Model metadata directory should exist: {model_dir}"
+            assert (
+                not ensemble_dir.exists()
+            ), f"Ensemble metadata directory should not exist: {ensemble_dir}"
+            assert (
+                not database_dir.exists()
+            ), f"Database metadata directory should not exist: {database_dir}"
+
+            # Clean up
+            exp.stop(model)
```
MattToast: Nit: Not strictly wrong, but do we need to `stop` the model if we already polled for its completion earlier in the test? Is this just to prevent breaking on a possible API change?

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,348 @@
...
+
+            # Check for run-specific subdirectories (may be 1 or 2 depending on timing)
+            run_dirs = [
+                d
+                for d in metadata_dir.iterdir()
+                if d.is_dir() and d.name.startswith("run_")
+            ]
+            assert (
+                len(run_dirs) >= 1
+            ), f"Should have at least one run directory, found: {run_dirs}"
```
MattToast: I'm guessing this is because the orchestrator will may not spin up in time? It doesn't look like we do any verification of the orchestrator directory anyway in this test, can we just remove its launch entirely??

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,348 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            # First experiment run
+            exp1 = Experiment("test_metadata_run1", exp_path=temp_dir, launcher="local")
+            model1 = exp1.create_model(
+                "test_model1", run_settings=exp1.create_run_settings("echo", ["run1"])
+            )
+
+            exp1.start(model1, block=False)
+            exp1.poll(interval=1)
+            exp1.stop(model1)
```
MattToast: functionally equivalent to
```suggestion
            exp1.start(model1, block=True)
```
? Is this just anticipating a possible future API break?

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,348 @@
...
+
+            # Second experiment run
+            exp2 = Experiment("test_metadata_run2", exp_path=temp_dir, launcher="local")
+            model2 = exp2.create_model(
+                "test_model2", run_settings=exp2.create_run_settings("echo", ["run2"])
+            )
+
+            exp2.start(model2, block=False)
+            exp2.poll(interval=1)
+            exp2.stop(model2)
```
MattToast: Ditto here
```suggestion
            exp2.start(model2, block=True)
```

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,348 @@
...
+            model = exp.create_model(
+                "batch_model",
+                run_settings=exp.create_run_settings("echo", ["batch_hello"]),
+            )
+
+            ensemble = exp.create_ensemble(
+                "batch_ensemble",
+                run_settings=exp.create_run_settings("echo", ["batch_world"]),
+                replicas=2,
+            )
```
MattToast: Just verifying this is intended: It looks like this test is supposed to be launching a model/ensemble with batch settings, but the local launcher is being used and no batch settings were assign to either

MattToast (changes requested):
Couple of nits on tests and such, but otherwise looks about ready to go on my end! Thanks for all thorough clean-up effort!!

smartsim/_core/launcher/dragon/dragonBackend.py
al-rigazzi: Right, but we may be removing the whole ZMQ thing in the future, so maybe we won't need this anymore.

smartsim/_core/launcher/step/step.py
al-rigazzi: Done and done!

tests/test_symlinking.py
al-rigazzi: You know me, I love monkeys.

smartsim/_core/launcher/dragon/dragonBackend.py
MattToast: Ahhh! In that case, this might be something not worth being addressed at all lol 😂

tests/test_symlinking.py
al-rigazzi: Nope, let's test them all. Also removed the defensive if.

tests/test_metadata_integration.py
al-rigazzi: Removed orc launch

tests/test_metadata_integration.py
al-rigazzi: Removed

tests/test_output_files.py
al-rigazzi: Should be fixed now?

tests/test_controller_metadata_usage.py
```diff
@@ -0,0 +1,85 @@
...
+    out_file = metadata_dir / "sample.out"
+    err_file = metadata_dir / "sample.err"
+    out_file.write_text("stdout")
+    err_file.write_text("stderr")
+
+    step = _DummyStep(metadata_dir, "sample")
+
+    entity_path = tmp_path / "workdir"
+    entity_path.mkdir()
+    entity = type("Entity", (), {"name": "sample", "path": str(entity_path)})
```
MattToast: Quick sanity check here: This is creating a new _type_ `Entity` equivalent to
```py
class Entity:
    name = "sample"          # <- Note: this is a class level var
    path = str(entity_path)  # <-       ditto here

entity = Entity  # <- `entity` is now an alias for the _type_ `Entity`
```
so `isintance(entity, type) == True` and `isinstance(e(), object) == True != isinstance(e(), type)`

is this intended or did you mean something like this:
```py
@dataclasses.dataclass  # <- decorator to make class level vars instance vars
class Entity:
    name: str = "sample"
    path: str = str(entity_path)

entity = Entity()  # <- `entity` now an instance of `Entity`
```
? _Technically_ this doesn't _really_ matter as we only expect one instance of this class, but I'm not gonna give you a review without being overly pedantic 😅

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+from smartsim.entity import Ensemble, Model
+from smartsim.settings import RunSettings
+
+
+class TestMetadataDirectoryIntegration:
+    """Integration tests for metadata directory creation across the SmartSim workflow"""
+
+    def test_experiment_creates_correct_metadata_directory_structure_model_only(self):
+        """Test that launching only models creates the correct directory structure"""
+        with tempfile.TemporaryDirectory() as temp_dir:
```
MattToast: Minor nit: Should this be using the `test_dir` fixture just so that artifacts are left over in the event of a failure for manual inspection?

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+import pytest
+
+from smartsim import Experiment
+from smartsim._core.config import CONFIG
+from smartsim.database.orchestrator import Orchestrator
+from smartsim.entity import Ensemble, Model
+from smartsim.settings import RunSettings
+
+
+class TestMetadataDirectoryIntegration:
```
MattToast: Minor nit: If there is no shared stand-up/tear-down logic in this test class, can we make all of these tests top level functions, just to match the style of the rest of the test suite?

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+            ), f"Ensemble metadata directory should not exist: {ensemble_dir}"
+            assert (
+                not database_dir.exists()
+            ), f"Database metadata directory should not exist: {database_dir}"
+
+    def test_experiment_creates_correct_metadata_directory_structure_ensemble_only(
+        self,
+    ):
+        """Test that launching only ensembles creates the correct directory structure"""
+        with tempfile.TemporaryDirectory() as temp_dir:
```
MattToast: ditto: `test_dir` fixture ask here

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+            # Create an ensemble
+            ensemble = exp.create_ensemble(
+                "test_ensemble",
+                run_settings=exp.create_run_settings("echo", ["world"]),
+                replicas=2,
+            )
+
+            # Start and wait for completion
+            exp.start(ensemble, block=False)
+            exp.poll(interval=1)
```
MattToast: Style nit: equivalent to `exp.start(ensemble, block=True)`? Any reason for the separation?

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+            # Check for entity-specific subdirectories with entity names
+            model_dir = run_dir / "model" / "test_model"
+            ensemble_dir = run_dir / "ensemble" / "test_ensemble"
+
+            assert (
+                model_dir.exists()
+            ), f"Model metadata directory should exist: {model_dir}"
+            assert (
+                ensemble_dir.exists()
+            ), f"Ensemble metadata directory should exist: {ensemble_dir}"
```
MattToast: Just checking because it was added in other tests: do we want this check here also?
```suggestion
            assert (
                model_dir.exists()
            ), f"Model metadata directory should exist: {model_dir}"
            assert (
                ensemble_dir.exists()
            ), f"Ensemble metadata directory should exist: {ensemble_dir}"
            assert (
                not database_dir.exists()
            ), f"Database metadata directory should not exist: {database_dir}"
```

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            exp = Experiment("test_metadata_perms", exp_path=temp_dir, launcher="local")
+
+            model = exp.create_model(
+                "test_model",
+                run_settings=exp.create_run_settings("echo", ["permissions"]),
+            )
+
+            exp.start(model, block=False)
+            exp.poll(interval=1)
```
MattToast: Similar style nit here

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+                run_dir = run_dirs[0]
+                assert run_dir.exists() and run_dir.is_dir()
+
+                # Check for entity-specific model directory with entity name
+                model_dir = run_dir / "model" / "test_model"
+                if model_dir.exists():
+                    assert model_dir.is_dir()
+                    assert model_dir.stat().st_mode & 0o700
+
+            exp.stop(model)
```
MattToast: Style nit: `model` is polled for earlier in this test, in theory this line is redundant
```suggestion
```

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+            ]
+            if run_dirs:
+                run_dir = run_dirs[0]
+                assert run_dir.exists() and run_dir.is_dir()
+
+                # Check for entity-specific model directory with entity name
+                model_dir = run_dir / "model" / "test_model"
+                if model_dir.exists():
+                    assert model_dir.is_dir()
+                    assert model_dir.stat().st_mode & 0o700
```
MattToast: If this `model.exisits()` condition is not met, I would argue that something has gone wrong. Thoughts on changing it to an `assert` instead?
```suggestion
                aseert model_dir.exists()
                assert model_dir.is_dir()
                assert model_dir.stat().st_mode & 0o700
```

tests/test_metadata_integration.py
```diff
@@ -0,0 +1,269 @@
...
+            assert (
+                metadata_dir.stat().st_mode & 0o700
+            )  # Owner should have read/write/execute
+
+            run_dirs = [
+                d
+                for d in metadata_dir.iterdir()
+                if d.is_dir() and d.name.startswith("run_")
+            ]
+            if run_dirs:
```
MattToast: Similar argument here, if the `run_dirs` is empty, I argue the test failed. Thoughts on making this an `assert` as well?
```py
assert len(run_dirs) == 1, "Incorrect number of run dirs created"
# TODO: Dedent everything in this `if` block
```

tests/test_metadata_integration.py
MattToast: applies for remaining tests in this class/file as well

tests/test_symlinking.py
```diff
@@ -48,24 +48,45 @@
...
+    ids=["ensemble", "model"],
 )
-def test_symlink(test_dir, entity):
+def test_symlink(test_dir, request, entity_fixture):
     """Test symlinking historical output files"""
+    entity = request.getfixturevalue(entity_fixture)
     entity.path = test_dir
     if entity.type == Ensemble:
-        for member in ens.models:
+        for member in entity.models:
```
MattToast, line 89: Not you're change, but this must have gotten past a different review. In theory `entity.type` should always return a `str`, meaning this block is unreachable. Do you think this should read
```py
if entity.type == "Ensemble":
    ...
```
or
```py
if isinstance(entity, Ensemble):
    ...
```
instead??

tests/test_symlinking.py
```diff
@@ -48,24 +48,45 @@
...
+
+
+@pytest.fixture
+def ensemble_entity():
+    return Ensemble("ens", params={}, run_settings=rs, batch_settings=bs, replicas=3)
+
+
+@pytest.fixture
+def orchestrator_entity():
+    return Orchestrator(db_nodes=3, batch=True, launcher="slurm", run_command="srun")
```
MattToast, line 64: LOVELY!! Thank you for adding!!

MattToast (approved):
Couple of small nits in the test suite, but nothing that I'm willing to hold this PR up over (I've already held it up to long lol 😅)

al-rigazzi:
@MattToast thanks for the review. I believe I addresseed your last comments and I will merge soon.

tests/test_symlinking.py
al-rigazzi: Yes, I believe you are right, let me see if I can change it and make this reachable

al-rigazzi: Done

PR #792, opened 2025-12-17
Drop Unused `sympy` Import

MattToast:
Looks like this import was added to this test file and is not used. Dropping because it's not listed in the test requirements, does not look like its used, and was causing me some problems when attempting to run tests locally.


al-rigazzi (approved):
Nice catch, LGTM pending changelog update!

al-rigazzi:
The macos-13 actions are deprecated, you should be able to merge the current `develop` into your branch and they'll go away.

PR #794, opened 2026-02-20
Miscellaneous test fixes

al-rigazzi:
This PR adds miscellaneous fixes to tests, preventing files from spilling in the root directory, allowing better control of Redis retries, and also preventing issues when Slurm would not assign CPU 0, 1, ... to the run, resulting in issues with cgroup. This last one is the most fragile of all fixes and it could be ignored - or improved.

I won't be able to follow what will happen with this PR, I'll leave it up to the team.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -747,7 +747,9 @@ def _cleanup_db(name: str) -> None:
     global database_registry
     db = database_registry[name]
     if db and db.is_active():
-        exp = Experiment("cleanup")
+        cleanup_dir = str(pathlib.Path(test_output_root) / "cleanup")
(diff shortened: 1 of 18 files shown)


MattToast (approved):
As my final gift to you before you leave the team I have no pedantic feedback for you (for once, lol)!

If you could document that we disable the custom pinning of CPUs `0..N` in one last issue on the off-chance we decide to come up with something more sophisticated in future, that would be much appreciated, but I'm not willing to drag out a "good enough" solution longer than it has to.

LGTM!! Thanks again for all of your amazing work on this project over the years!! :')

PR #799, opened 2026-04-22
fix: treat path=None as getcwd() in Model, Ensemble, and Orchestrator

### LINKED ISSUE #556: Unexpected `path` attr set for SmartSim entities when passing an `Optional` `path` param
## Description
Currently the `path` parameter to the `__init__` methods of `Ensemble`/`Model`/`Orchestrator` are type hinted as so (or equivalent):
```py
def __init__(
    self,
    ...
    path: str | None = os.getcwd(),
    ...
) -> None:
```
However, if a  user were to exercise the `Optional` type of `path`, an invalid path of the string `"None"` is set on the object, effectively leaving the object in an invalid state that cannot be launched. 

## How to reproduce
Consider the following repl
```py
>>> from smartsim.database.orchestrator import Orchestrator
>>> from smartsim.entity.ensemble import Ensemble
>>> from smartsim.entity.model import Model
>>> from smartsim.settings import RunSettings
>>>
>>> Orchestrator(path=None).path
'None'
>>> rs = RunSettings("echo", ["hello", "world"])
>>> Ensemble("test-ens", {}, path=None, run_settings=rs, replicas=1).path
'None'
>>> Model("test-model", {}, rs, path=None).path
'None'
```

## Expected behavior
The path should not be set to the string `"None"`, but should instead be set to the current working directory. Whether that is the current working directory of at "compile time" or at "run time" remains to be decided.

Alternatively, we could have the `path` parameter not have a default value.

## System
Any/All

## Acceptance Criteria
- [ ] Decide on if/what the default path for an `Model`/`Ensemble`/`Orchestrator` should be
- [ ] Passing a `path=None` param to `Model`/`Ensemble`/`Orchestrator` results in a "valid" path, or disallowed entirely
- [ ] If a default path is agreed upon, add tests to ensure that when `path=None` is specified the correct default path is set

CrepuscularIRIS:
## What

Replace `str(path)` with `path if path is not None else getcwd()` in the `__init__` of `Model`, `Ensemble`, and `Orchestrator`, so that passing `path=None` defaults to the current working directory rather than storing the literal string `"None"`.

## Why

Fixes #556.

When a caller passes `path=None` explicitly (or relies on the default), `str(None)` evaluates to `"None"`, which is stored as `self.path`. Any subsequent file I/O (e.g. writing output files, resolving configuration paths) fails because `"None"` is not a valid filesystem path. Since `getcwd` is already imported in all three files, the fix requires no new imports.

## How

At each of the three call sites that forwarded `path` to `super().__init__()`, replace:

```python
str(path)
```

with:

```python
path if path is not None else getcwd()
```

The affected lines are:
- `smartsim/entity/model.py` — `Model.__init__` → `super().__init__`
- `smartsim/entity/ensemble.py` — `Ensemble.__init__` → `super().__init__`
- `smartsim/database/orchestrator.py` — `Orchestrator.__init__` → `super().__init__`

No changes were made to the base class (`SmartSimEntity`) or any other file.

## Scope

- **Included:** fix for `Model`, `Ensemble`, and `Orchestrator`; regression tests for `Model` and `Ensemble`
- **NOT included:** a regression test for `Orchestrator` — the constructor checks for Redis binaries at import time, which requires the full SmartSim extension stack. The fix is applied and the logic is identical to the other two classes; a test could be added once the CI environment provides those binaries.

## Verification

- [ ] `tests/test_entity_path_none.py::test_model_path_none_defaults_to_cwd` — passes after fix, fails on unpatched code
- [ ] `tests/test_entity_path_none.py::test_model_explicit_path_unchanged` — passes in both cases (regression guard)
- [ ] `tests/test_entity_path_none.py::test_ensemble_path_none_defaults_to_cwd` — passes after fix, fails on unpatched code
- [ ] `tests/test_entity_path_none.py::test_ensemble_explicit_path_unchanged` — passes in both cases (regression guard)

Manual check (no SmartSim install needed for the logic itself):
```python
from os import getcwd
path = None
result = path if path is not None else getcwd()
assert result == getcwd()   # True — never "None"
```

## AI Disclosure

AI tools were used to assist with identifying the fix sites and drafting the test file. All changes were manually reviewed for correctness before submission.

### DIFF
--- a/smartsim/database/orchestrator.py
+++ b/smartsim/database/orchestrator.py
@@ -229,7 +229,7 @@ class Orchestrator(EntityList[DBNode]):
 
         super().__init__(
             name=db_identifier,
-            path=str(path),
+            path=path if path is not None else getcwd(),
             port=port,
             interface=interface,
             db_nodes=db_nodes,
--- a/smartsim/entity/ensemble.py
+++ b/smartsim/entity/ensemble.py
@@ -96,7 +96,7 @@ class Ensemble(EntityList[Model]):
         self.run_settings = run_settings
         self.replicas: str
 
-        super().__init__(name, str(path), perm_strat=perm_strat, **kwargs)
+        super().__init__(name, path if path is not None else getcwd(), perm_strat=perm_strat, **kwargs)
 
     @property
     def models(self) -> Collection[Model]:
--- a/smartsim/entity/model.py
+++ b/smartsim/entity/model.py
@@ -91,7 +91,7 @@ class Model(SmartSimEntity):
         :param batch_settings: Launcher settings for running the individual
                                model as a batch job
         """
-        super().__init__(name, str(path), run_settings)
+        super().__init__(name, path if path is not None else getcwd(), run_settings)
         self.params = _parse_model_parameters(params)
         self.params_as_args = params_as_args
         self.incoming_entities: list[SmartSimEntity] = []
--- a/tests/test_entity_path_none.py
+++ b/tests/test_entity_path_none.py
@@ -1,0 +1,104 @@
+# BSD 2-Clause License
+#
+# Copyright (c) 2021-2025, Hewlett Packard Enterprise
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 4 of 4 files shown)


CrepuscularIRIS:
Closing this PR — no maintainer activity in 2+ weeks. Happy to reopen or resubmit if you'd like to revisit. Thanks for the project!

